PR #33, opened 2023-08-07
Support LLaMa2, bump dependencies, and update README

yizhongw:


open_instruct/finetune.py
```diff
@@ -393,7 +394,7 @@ def main():
 
     # no default pad token for llama!
     # here we add all special tokens again, because the default ones are not in the special_tokens_map
-    if isinstance(tokenizer, LlamaTokenizer):
+    if isinstance(tokenizer, LlamaTokenizer) or isinstance(tokenizer, LlamaTokenizerFast):
```
hamishivi, line 397: I had some annoyances with the way the fasttokenizer dealt with adding tokens in the past, although it's probably fixed now. Worth double checking the tokenization makes sense after adding the special characters.

requirements.txt
```diff
@@ -1,24 +1,27 @@
 torch
+scipy
+packaging
 sentencepiece
 datasets
-deepspeed==0.9.1
-# accelerate==0.18.0
-git+https://github.com/huggingface/accelerate.git@e06e7b35e7c49563ed70f31b78126bf70b0252b9  # this version fixes a bug about gradient accumulation in 0.18.0
```
hamishivi, line 6: Guessing this bug got fixed?

hamishivi (approved):
All seems good to me, I'm a little worried about minor tokenizer issues but we can patch them as they pop up. I haven't been able to test out llama 2 much due to holidays so my previous problems might just be fine now.

requirements.txt
yizhongw: Yeah I have checked the gradient accumulation and it works fine.

open_instruct/finetune.py
yizhongw: I tried using the fasttokenizer in both training and inference and it seems okay. Let's merge it for now!

PR #46, opened 2023-08-18
When doing LORA fine tuning with grouped optimizers we get an error:

prajdabre:
ValueError: max() arg is an empty sequence

To avoid this we dont use grouped optimizers.

This is a temporary fix.


hamishivi:
Hi, I wasn't able to reproduce this issue on my own - running deepspeed and LoRA worked fine. Could you share more details on what software versions and commands you are using?

PR #52, opened 2023-08-28
[Bug Fix] Fix step misalignment when resuming from a checkpoint

tongyx361:
**Note 1: `completed_steps` update as the optimizer steps, while `step` as the dataloader**, so there is a scaling factor of `gradient_accumulation_steps` between them.

**Note 2: `step` starts from 0, while `completed_steps` starts from 1**, so we need to add 1 to `step` to decide whether to add 1 to `completed_steps`.
Otherwise, e.g. "step_300" , resume_step = 4800, args.gradient_accumulation_steps = 16, 
completed_steps < 300 should be skipped, while completed_steps >= 300 should not; 
0 <= step < 4800 should be skipped, while step >= 4800 should not be skipped. 
When step = 4784 (0 = 0 * 16, ..., 4784 = 299 * 16) => completed_steps = 299; 
after `completed_steps += 1`, completed_steps = 300; 
then step = 4785, completed_steps = 300, and this step batch would not be skipped, 
which is not what we expect.

### DIFF
--- a/open_instruct/finetune.py
+++ b/open_instruct/finetune.py
@@ -550,61 +550,55 @@ def main():
     # Potentially load in the weights and states from a previous save
     if args.resume_from_checkpoint:
         if args.resume_from_checkpoint is not None or args.resume_from_checkpoint != "":
-            accelerator.print(f"Resumed from checkpoint: {args.resume_from_checkpoint}")
-            accelerator.load_state(args.resume_from_checkpoint)
+            checkpoint_path = args.resume_from_checkpoint
             path = os.path.basename(args.resume_from_checkpoint)
         else:
             # Get the most recent checkpoint
             dirs = [f.name for f in os.scandir(os.getcwd()) if f.is_dir()]
             dirs.sort(key=os.path.getctime)
-            path = dirs[-1]  # Sorts folders by date modified, most recent checkpoint is the last
+            path = dirs[
+                -1
+            ]  # Sorts folders by date modified, most recent checkpoint is the last
+            checkpoint_path = path
+            path = os.path.basename(checkpoint_path)
+
+        accelerator.print(f"Resumed from checkpoint: {checkpoint_path}")
+        accelerator.load_state(path)
         # Extract `epoch_{i}` or `step_{i}`
         training_difference = os.path.splitext(path)[0]
 
         if "epoch" in training_difference:
             starting_epoch = int(training_difference.replace("epoch_", "")) + 1
             resume_step = None
+            completed_steps = starting_epoch * num_update_steps_per_epoch
         else:
             # need to multiply `gradient_accumulation_steps` to reflect real steps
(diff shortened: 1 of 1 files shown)


yizhongw:
Hi @tongyx361 thanks for the PR! The issue you pointed and changes you suggested look correct to me.

I noticed the huggingface official example also had a few new commits (https://github.com/huggingface/transformers/commit/b191d7db442fcd4ea794768363155c091e60b868, https://github.com/huggingface/transformers/commit/f7d80cb3d25e26cfae7fa5c91bc6fe8552ed68ea, https://github.com/huggingface/transformers/commit/656e17f6f7eded9df87ad59cbd064fdf5f44f708) for fixing this. So, we need to update this code anyway. 

Would you like to update the PR by using the huggingface way to do this? That will be very much appreciated!

tongyx361:
> Hi @tongyx361 thanks for the PR! The issue you pointed and changes you suggested look correct to me.
> 
> I noticed the huggingface official example also had a few new commits ([huggingface/transformers@b191d7d](https://github.com/huggingface/transformers/commit/b191d7db442fcd4ea794768363155c091e60b868), [huggingface/transformers@f7d80cb](https://github.com/huggingface/transformers/commit/f7d80cb3d25e26cfae7fa5c91bc6fe8552ed68ea), [huggingface/transformers@656e17f](https://github.com/huggingface/transformers/commit/656e17f6f7eded9df87ad59cbd064fdf5f44f708)) for fixing this. So, we need to update this code anyway.
> 
> Would you like to update the PR by using the huggingface way to do this? That will be very much appreciated!

Thanks for your attention!
I've updated the PR by using the huggingface way to fix the resuming logics. (see 82cc80e9a97d3ee1bdad29755bfcdeec21283359)

tongyx361:
> Hi @tongyx361 thanks for the PR! The issue you pointed and changes you suggested look correct to me.
> 
> I noticed the huggingface official example also had a few new commits ([huggingface/transformers@b191d7d](https://github.com/huggingface/transformers/commit/b191d7db442fcd4ea794768363155c091e60b868), [huggingface/transformers@f7d80cb](https://github.com/huggingface/transformers/commit/f7d80cb3d25e26cfae7fa5c91bc6fe8552ed68ea), [huggingface/transformers@656e17f](https://github.com/huggingface/transformers/commit/656e17f6f7eded9df87ad59cbd064fdf5f44f708)) for fixing this. So, we need to update this code anyway.
> 
> Would you like to update the PR by using the huggingface way to do this? That will be very much appreciated!

I've refactored it in the HF Transformers example way and  with minimal revision. (see 1e9a885779c9d34d16e5c9c8da3b4dceee609b0c)

yizhongw:
Lgtm. Thanks for the contribution! I merged it into the main branch. 

Cc @hamishivi

PR #55, opened 2023-09-03
[Refactor] Update evaluation scripts to support customized prompting format for specific models (e.g., LLaMa2-chat)

yizhongw:


eval/bbh/run_eval.py
```diff
@@ -8,6 +8,7 @@
 import random
 import evaluate
 from eval.utils import load_hf_lm_and_tokenizer, generate_completions, query_openai_chat_model
+from eval.templates import *
```
hamishivi, line 11: nit: I dislike * imports, would prefer specific imports if possible.

eval/bbh/run_eval.py
```diff
@@ -21,10 +22,16 @@ def eval_hf_model(args, model, tokenizer, examples, task_prompt, save_path=None)
 
     prompts = []
     for example in examples:
+        prompt = task_prompt.strip() + "\n\nQ: " + example["input"]
         if args.use_chat_format:
-            prompt = "<|user|>\n" + task_prompt.strip() + "\n\nQ: " + example["input"] + "\n<|assistant|>\nA:"
+            messages = [{"role": "user", "content": prompt}]
+            prompt = eval(args.chat_formatting_function)(messages, add_bos=False)
```
hamishivi: eval + * import seems like not the best way to do this, imo - would it be possible to e.g. import a class/method from a user-specific file (using https://docs.python.org/3/library/importlib.html#importlib.import_module or similar)?

eval/bbh/run_eval.py
```diff
@@ -47,15 +54,14 @@ def eval_hf_model(args, model, tokenizer, examples, task_prompt, save_path=None)
...
-        output = output.strip().split("\n\n")[0].strip()
-
         # extract the first answer after `So the answer is` and before the next period.
         # if there is no such answer, we will just use the raw output.
         results = re.search(r"So the answer is (.*?)\.", output)
         if results:
             prediction = results.group(1).strip()
         else:
+            # only keep the first part of the output - this is mainly for vanilla language models.
+            output = output.strip().split("\n\n")[0].strip()
```
hamishivi, line 69: Why this change? For the behaviour with Llama-2?

hamishivi (approved):
I think the functionality is fine (assuming you've tested this doesn't actually yield prompt changes, but from a glance it seems fine). I'm not sure the * import + eval is the best way to implement this, though... 

We can always redesign that, and this code is 'run at your own risk', so I think its okay for now.

eval/bbh/run_eval.py
yizhongw: Yeah, this might be a better way. Let me change this.

eval/bbh/run_eval.py
yizhongw: Sometimes the "So the answer is" appear in the second part after "\n\n", so I think we can try to find the "So the answer is" first, and rollback to the first part of the answer if not found.

For LLaMa2-chat, it's more complicated. The model cannot follow the few-shot demonstration format. It cannot generate the "So the answer is" string, so its performance is still bad.

PR #56, opened 2023-09-03
Update merge_lora.py

TJKlein:
Use the base model for the tokenizer. LoRa models do not necessarily have tokenizer information, just the adapter files.

### DIFF
--- a/open_instruct/merge_lora.py
+++ b/open_instruct/merge_lora.py
@@ -22,7 +22,7 @@ if __name__ == "__main__":
     print("Merging the lora modules...")
     merged_model = lora_model.base_model.merge_and_unload()
     output_dir = args.output_dir if args.output_dir else args.lora_model_name_or_path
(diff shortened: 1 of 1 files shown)


yizhongw:
Hi @TJKlein thanks for creating this PR. The LoRa model's vocabulary might be different from the base model (e.g., when you add some special tokens during finetuning). I think a more proper way to do this is to allow specifying `tokenizer_name_or_path` in the arguments. By default, we should still use the lora model's vocabulary.

TJKlein:
Then I guess we should change it to something like below. So it would try loading the customized version if it exists, and as a fallback option, take the base tokenizer.
```python
try:
    tokenizer = AutoTokenizer.from_pretrained(args.lora_model_name_or_path)
except:
    tokenizer = AutoTokenizer.from_pretrained(args.base_model_name_or_path)
```

hamishivi:
I've made this change in https://github.com/allenai/open-instruct/pull/60 - thanks for the suggestion!

PR #59, opened 2023-09-15
QLora support

hamishivi:
- Adding support for training using Qlora (DDP only for now)
- Add gradient checkpointing support
- Add gradient clipping support
- Add llama2 70b support

Recommended hyperparameters for training included in the example script. I found that merging with the non-quantized model usually was slightly better, despite some chatter online saying otherwise.

Ignore the commits from jacob, they have all been reverted :)


open_instruct/finetune.py
```diff
@@ -381,12 +432,33 @@ def main():
...
+                from_tf=bool(".ckpt" in args.model_name_or_path),
+                config=config,
+                load_in_4bit=True,
+                quantization_config=bnb_config,
+                device_map=device_map,
+                torch_dtype=torch.bfloat16,
+            )
+        else:
+            device_index = accelerator.process_index
+            device_map = {"": device_index} # force data-parallel training.
```
yizhongw: If not using qlora, does this still force data-parallel training? Why do we need to?

open_instruct/llama_flash_attn_monkey_patch.py
yizhongw, line 1: Is this file still necessary? Will the llama2 one also work for llama1?

open_instruct/finetune.py
hamishivi: Ah, good point. Yes, this still forces data-parallel training. You have to do this unless you use deepspeed (since FSDP doesn't currently work with mixed requires_grad), but probably the default for non-qlora should be to use deepspeed. Let me fix this up.

open_instruct/llama_flash_attn_monkey_patch.py
hamishivi: Good point, I'll double check, but my guess is the llama2 one should be fine, since 7/13b is the same architecture anyway.

open_instruct/finetune.py
yizhongw: Yeah let's use deepspeed by default for non-qlora

PR #62, opened 2023-09-22
Fix MMLU answer token and probability calculation

OyvindTafjord:
@hamishivi @yizhongw This is a fix I added a while back, might be worth incorporating in next round of evaluations. It addresses two things:

1) The MMLU evaluation was scoring answers like "Answer:A" rather than "Answer: A", which meant the probabilities over the candidate tokens didn't add up to anything close to 1 (but probably still quite correlated with the version with prefixed space).

2) The reported probability over answer choices was done after restricting to the candidate tokens, meaning it would always add up to 1. This tweak to the code does normalization over all tokens instead, so it more easily highlights this problem when the probability mass on candidate tokens is small.

This is probably particularly troublesome if using a few-shot prompt where the space after "Answer:" is explicit in the examples


yizhongw:
@OyvindTafjord Thanks for the catch! This looks good to me. Did you observe performance change before and after adding the space prefix?

OyvindTafjord:
I didn't run many experiments, I'll forward an email exchange with Hamish about this, one result I had was:

yizhongw:
Interesting. The change in predictions do look serious to me. I will merge this, and we need to rerun other evals I guess.

PR #63, opened 2023-09-22
Use latest alpaca_eval library to evaluate performance on alpaca farm.

yizhongw:


eval/alpaca_farm/run_eval.py
```diff
@@ -0,0 +1,167 @@
...
+    alpaca_eval_data = datasets.load_dataset("tatsu-lab/alpaca_farm", "alpaca_farm_evaluation")["eval"]
+    prompts = []
+    chat_formatting_function = dynamic_import_function(args.chat_formatting_function) if args.use_chat_format else None
+    for example in alpaca_eval_data:
+        input_text = example["instruction"] + "\n\n" + example["input"] if example["input"] != "" else example["instruction"]
+        if args.use_chat_format:
+            messages = [{"role": "user", "content": input_text}]
+            prompt = chat_formatting_function(messages, add_bos=False)
+        else:
+            prompt = example["instruction"] + "\n\n" + example["input"] if example["input"] != "" else example["instruction"]
```
hamishivi: nit: you can probably just make this `prompt = input_text`

requirements.txt
```diff
@@ -27,8 +27,8 @@ einops
 flash-attn
 auto-gptq
 fire
-# alpaca-farm eval - my fork with some fixes.
-git+https://github.com/hamishivi/alpaca_farm.git
+alpaca-eval==0.3.1
```
hamishivi, line 30: The one thing (discussed elsewhere) is that alpacaeval uses a different prompt to alpacafarm. I'm happy to shift to the new package, but it makes our old numbers not really comparable.

scripts/eval/alpaca_farm.sh
```diff
@@ -1,4 +1,26 @@
-# example script for alpacafarm
-export OPENAI_API_KEY=YOUR_API_KEY
+# python -m eval.alpaca_farm.run_eval \
```
hamishivi: Don't you still need the openai key?

eval/alpaca_farm/run_eval.py
```diff
@@ -0,0 +1,167 @@
...
+
+def main(args):
+    random.seed(42)
+    os.makedirs(args.save_dir, exist_ok=True)
+
+    logging.info("loading data and model...")
+    alpaca_eval_data = datasets.load_dataset("tatsu-lab/alpaca_farm", "alpaca_farm_evaluation")["eval"]
+    prompts = []
+    chat_formatting_function = dynamic_import_function(args.chat_formatting_function) if args.use_chat_format else None
+    for example in alpaca_eval_data:
```
hamishivi, line 21: In the original script I had a check for saved generations. Do you think its worth adding this? I found it useful, e.g. in case I ran the code and forgot to have my api key set - I could just rerun the script after setting the key, and skip the (lengthy) generation process.

hamishivi (approved):

eval/alpaca_farm/run_eval.py
yizhongw: This should be fine, since the generation is pretty quick now with vllm (4 mins for running 70B model on 4 GPUs). Let's keep the code clean.

scripts/eval/alpaca_farm.sh
yizhongw: That is expected to be set in the environment. I want to delete this because if it's written in the code, we have to comment it out before committing and pushing it to github. Let me add a comment.

PR #67, opened 2023-10-01
Refine the lora merging code.

yizhongw:


hamishivi (approved):
Seems fine. I think that maybe we could just run the dequant entirely on cpu by default, but this is okay too.

PR #73, opened 2023-10-16
Switch to using the built-in flash attention 2 support in Transformers library.

yizhongw:


Dockerfile
```diff
@@ -10,9 +10,9 @@ WORKDIR /stage/
 
 COPY requirements.txt .
 RUN pip install --upgrade pip setuptools wheel
-RUN pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cu118
+RUN pip install torch==2.0.1 torchvision torchaudio --index-url https://download.pytorch.org/whl/cu118
```
hamishivi, line 13: is it possible to bump to 2.1.0 at some point?

hamishivi (approved):
seems fine assuming you tested this to make sure that PEFT training still worked fine!

yizhongw:
> seems fine assuming you tested this to make sure that PEFT training still worked fine!

I tested the scripts and the training works. I just ran a few steps though so not sure how this will affect the final performance. We can see if later training works fine.

Dockerfile
yizhongw: Some of our dependencies require 2.0.1 iirc. Currenly, 2.1.0 might be too new to incorporate unless there are some features we need. Otherwise, let's revisit later.

PR #85, opened 2023-11-27
replace `:` in eval/mmlu/run_eval.py

Forence1999:
replace `prompt += "The answer is:"` by `prompt += "The answer is "`. The colon will cause slight decrease and fluctuation of performance, since the model has not been trained in such sentences.

### DIFF
--- a/eval/mmlu/run_eval.py
+++ b/eval/mmlu/run_eval.py
@@ -58,9 +58,9 @@ def eval_hf_model(args, subject, model, tokenizer, dev_df, test_df, batch_size=1
             messages = [{"role": "user", "content": prompt}]
             prompt = chat_formatting_function(messages, add_bos=False)
             if prompt[-1] in ["\n", " "]:
-                prompt += "The answer is:"
+                prompt += "The answer is "
             else:
-                prompt += " The answer is:"
+                prompt += " The answer is "
 
         tokenized_prompt = tokenizer(prompt, truncation=False, add_special_tokens=False).input_ids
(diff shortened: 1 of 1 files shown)


yizhongw:
Hi @Forence1999 , sorry for being late on this thread. May I know how much improvement you saw from this change, and on which model? Models these days can be a bit sensitive about the prompts. The colon was used in the [original MMLU eval script](https://github.com/hendrycks/test/blob/4450500f923c49f1fb1dd3d99108a0bd9717b660/evaluate.py#L33). Our evaluation results of llama can match the llama paper, so we tend to keep this unless there is a notable difference.

Forence1999:
> Hi @Forence1999 , sorry for being late on this thread. May I know how much improvement you saw from this change, and on which model? Models these days can be a bit sensitive about the prompts. The colon was used in the [original MMLU eval script](https://github.com/hendrycks/test/blob/4450500f923c49f1fb1dd3d99108a0bd9717b660/evaluate.py#L33). Our evaluation results of llama can match the llama paper, so we tend to keep this unless there is a notable difference.

Hi, Yizhong, Sorry for this late reply! I finetune LLaMA2-7B model on SuperNI dataset, and observed slightly over 1% improvement on MMLU. Intuitively,deleting the colon should be better, but following the regular practice should also be acceptable :)

hamishivi:
Closing since it seems that the performance diff isn't massive, and to keep consistent with the original MMLU implementation.

PR #86, opened 2023-11-28
DPO scripts + docker update + readme update

hamishivi:
This PR:
- Adds dpo training script and example bash scripts. I've confirmed that the full-finetuning script replicates tulu-v2-dpo-7b. PEFT works with it too, although I haven't done any hyperparameter tuning or trained a model using it.
- Inline the beaker docker container into the dockerfile to make it portable out of Ai2.
- Update readme for Tulu 2 and associated paper :)


yizhongw:
Lgtm, thanks!!

PR #92, opened 2023-12-19
[WIP] Some various fixes and changes for olmo

hamishivi:
A bunch of changes. Mostly fixes to allow Olmo finetuning. Please let me test this a bit first before merging, as I need to make sure the Olmo tokenizer configs upstream are correct.

Additionally, I added:
- option to force bos token adding in training
- option to control distributed training timeout from the 30 minute default
- turned off safe saving due to it not working with our pinned accelerate version
- fixed the embedding check
- option for loading models with remote code
- eos token explicitly set in generation util
- ability to use HF_TOKEN env variable for model loading during evaluation

### DIFF
--- a/eval/templates.py
+++ b/eval/templates.py
@@ -16,6 +16,24 @@ def create_prompt_with_tulu_chat_format(messages, bos="<s>", eos="</s>", add_bos
     formatted_text = bos + formatted_text if add_bos else formatted_text
     return formatted_text
 
+# weirdness with olmo tokenizer means IP_ADDR is the eos token.
+def create_prompt_with_tulu_chat_format_olmo_format(messages, bos="|||IP_ADDRESS|||", eos="|||IP_ADDRESS|||", add_bos=True):
+    formatted_text = ""
+    for message in messages:
+        if message["role"] == "system":
+            formatted_text += "<|system|>\n" + message["content"] + "\n"
+        elif message["role"] == "user":
+            formatted_text += "<|user|>\n" + message["content"] + "\n"
+        elif message["role"] == "assistant":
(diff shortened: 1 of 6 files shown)


eval/templates.py
```diff
@@ -16,6 +16,24 @@ def create_prompt_with_tulu_chat_format(messages, bos="<s>", eos="</s>", add_bos
     formatted_text = bos + formatted_text if add_bos else formatted_text
     return formatted_text
 
+# weirdness with olmo tokenizer means IP_ADDR is the eos token.
+def create_prompt_with_tulu_chat_format_olmo_format(messages, bos="|||IP_ADDRESS|||", eos="|||IP_ADDRESS|||", add_bos=True):
```
yizhongw: should we call this `create_prompt_with_olmo_chat_format`?

open_instruct/finetune.py
```diff
@@ -667,8 +720,24 @@ def main():
             active_dataloader = train_dataloader
         for step, batch in enumerate(active_dataloader):
             with accelerator.accumulate(model):
-                outputs = model(**batch, use_cache=False)                
-                loss = outputs.loss
+                outputs = model(**batch, use_cache=False)
+                if args.reduce_loss == 'mean':
+                    loss = outputs.loss
+                else:
+                    # we are going to calculate the loss ourself for sum avging.
```
yizhongw: Probably it's better to give more explanation here, and refer to this discussion: https://github.com/huggingface/transformers/issues/24725?

yizhongw (commented):
Lgtm. Feel free to merge!

eval/templates.py
hamishivi: Good point, changed to fit with the other template functions.

PR #99, opened 2024-01-07
Better HumanEval instructions

Muennighoff:
Currently, the chat format is not used for CodexEval / HumanEval. However, this does not correspond to how users use the models.

This PR adds HumanEval instructions from HumanEvalPack to hopefully enable evaluating via the chat format in a realistic setting with good performance.

### DIFF
--- a/eval/codex_humaneval/run_eval.py
+++ b/eval/codex_humaneval/run_eval.py
@@ -42,7 +42,7 @@ def main(args):
             answer = "Here is the function:\n\n```python\n"
             stop_sequences.append("\n```")
         else:
-            print(f"Could not find HumanEvalPack file at {args.data_file_hep}, using default instructions. This will result in significantly worse performance. You can download it at https://hf.co/datasets/bigcode/humanevalpack/blob/main/data/python/data/humanevalpack.jsonl")
+            print(f"Could not find HumanEvalPack file at {args.data_file_hep}, which will result in significantly worse performance. You can download it at https://hf.co/datasets/bigcode/humanevalpack/blob/main/data/python/data/humanevalpack.jsonl")
             instructions_dict = None
             answer = "Here is the completed function:\n\n\n"
 
--- a/eval/codex_humaneval/run_eval.py
+++ b/eval/codex_humaneval/run_eval.py
@@ -29,7 +29,7 @@ def main(args):
         prompts = []
         chat_formatting_function = dynamic_import_function(args.chat_formatting_function)
         # If available use more realistic instructions from HumanEvalPack (https://hf.co/datasets/bigcode/humanevalpack)
-        if os.ispath(args.data_file_hep):
+        if os.path.exists(args.data_file_hep):
             with open(args.data_file_hep, "r") as f:
                 instructions = [json.loads(l) for l in f]
                 instructions_dict = {
--- a/eval/codex_humaneval/run_eval.py
+++ b/eval/codex_humaneval/run_eval.py
@@ -29,6 +29,7 @@ def main(args):
         prompts = []
         chat_formatting_function = dynamic_import_function(args.chat_formatting_function)
         for example in test_data:
+            # For better instructions you may want to use HumanEvalSynthesize here: https://github.com/bigcode-project/bigcode-evaluation-harness/tree/main/docs#humanevalpack
             messages = [{"role": "user", "content": "Complete the following python function.\n\n\n" + example["prompt"]}]
             prompt = chat_formatting_function(messages, add_bos=False)
             if prompt[-1] in ["\n", " "]:
--- a/eval/codex_humaneval/run_eval.py
+++ b/eval/codex_humaneval/run_eval.py
@@ -55,6 +55,7 @@ def main(args):
         for example in test_data:
             if instructions_dict is not None:
                 instruction = instructions_dict[example["task_id"]]
+            # return partials that we will finalise once we have the tokenizer.
             prompts.append(lambda tokenizer: apply_chat_format(example, tokenizer, instruction, answer))
     else:
         prompts = [example["prompt"] for example in test_data]
--- a/eval/codex_humaneval/run_eval.py
+++ b/eval/codex_humaneval/run_eval.py
@@ -28,14 +28,28 @@ def main(args):
     if args.use_chat_format:
         prompts = []
         chat_formatting_function = dynamic_import_function(args.chat_formatting_function)
+        # If available use more realistic instructions from HumanEvalPack (https://hf.co/datasets/bigcode/humanevalpack)
+        if os.ispath(args.data_file_hep):
+            with open(args.data_file_hep, "r") as f:
+                instructions = [json.loads(l) for l in f]
+                instructions_dict = {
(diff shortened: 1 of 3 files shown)


eval/codex_humaneval/run_eval.py
```diff
@@ -28,13 +28,28 @@ def main(args):
     if args.use_chat_format:
         prompts = []
         chat_formatting_function = dynamic_import_function(args.chat_formatting_function)
+        # If available use more realistic instructions from HumanEvalPack (https://hf.co/datasets/bigcode/humanevalpack)
+        if os.ispath(args.data_file_hep):
```
hamishivi: I think this should be `os.path.exists`?

hamishivi (commented):
I tested this locally and using the chat instructions gave large performance differences for Tulu-2-7b-dpo (.01 vs .25 with the non-chat setup), so I think there's extra prompt formatting that might be required somewhere. If you have a reference implementation (eleutherAI harness?) I can take a look at that. The gap is smaller for less chatty models (.20 vs .22 for tulu-2-7b).

cc @yizhongw

eval/codex_humaneval/run_eval.py
Muennighoff:
```suggestion
        if os.path.exists(args.data_file_hep):
```

Muennighoff:
> I tested this locally and using the chat instructions gave large performance differences for Tulu-2-7b-dpo (.01 vs .25 with the non-chat setup), so I think there's extra prompt formatting that might be required somewhere. If you have a reference implementation (eleutherAI harness?) I can take a look at that. The gap is smaller for less chatty models (.20 vs .22 for tulu-2-7b).
> 
> cc @yizhongw

Yeah the issue is with stopping I think. In https://github.com/bigcode-project/bigcode-evaluation-harness/blob/199eeec6549eecc049f338c5155e4b5fd9bdac73/bigcode_eval/tasks/humanevalpack.py#L286 we postprocess such that whenever the model no longer generates a tab, we stop.

Another solution, is to simply use ``` and add this as a stopword, which should work well for these models as parts of their IFT data should have such marks. I did this one for now, but can also switch to the postprocessing solution if you prefer. Maybe you could check how running with the changes I just made compare with your previous numbers.

eval/codex_humaneval/run_eval.py
```diff
@@ -26,20 +26,36 @@ def main(args):
...
-        # return partials that we will finalise once we have the tokenizer.
+            prefix = " " if prompt[-1] in ["\n", " "] else ""
+            return prompt + prefix + answer + example["prompt"]      
+            
+        instruction = "Complete the following python function.\n\n\n"
         for example in test_data:
-            prompts.append(lambda tokenizer: apply_chat_format(example, tokenizer))
+            if instructions_dict is not None:
+                instruction = instructions_dict[example["task_id"]]
+            prompts.append(lambda tokenizer: apply_chat_format(example, tokenizer, instruction, answer))
```
Muennighoff:
```suggestion
            # return partials that we will finalise once we have the tokenizer.
            prompts.append(lambda tokenizer: apply_chat_format(example, tokenizer, instruction, answer))
```

eval/codex_humaneval/run_eval.py
```diff
@@ -26,20 +26,39 @@ def main(args):
...
+        if os.path.exists(args.data_file_hep):
+            with open(args.data_file_hep, "r") as f:
+                instructions = [json.loads(l) for l in f]
+                instructions_dict = {
+                    x["task_id"].replace("Python", "HumanEval"): x["instruction"] for x in instructions
+                }
+            answer = "Here is the function:\n\n```python\n"
+            stop_sequences.append("\n```")
+        else:
+            print(f"Could not find HumanEvalPack file at {args.data_file_hep}, using default instructions. This will result in significantly worse performance. You can download it at https://hf.co/datasets/bigcode/humanevalpack/blob/main/data/python/data/humanevalpack.jsonl")
```
Muennighoff:
```suggestion
            print(f"Could not find HumanEvalPack file at {args.data_file_hep}, which will result in significantly worse performance. You can download it at https://hf.co/datasets/bigcode/humanevalpack/blob/main/data/python/data/humanevalpack.jsonl")
```

Muennighoff:
Updated the PR to
a) Finalize the HumanEvalPack integration
b) Fix a regression in https://github.com/allenai/open-instruct/pull/110 where for GSM8K & HumanEval, it would always evaluate the same example in chat format. This is because when collecting the partials, it will not store the example from that point in time, but only later use the last available example. Hence in the GSM8K code, it also doesn't throw an error even though the tokenizer isn't defined when the lambda function is appended i.e. `prompts.append(lambda tokenizer: apply_chat_format(example, tokenizer))`.


Below are results with this PR:

HumanEval no chat (i.e. same as previously):
`python -m eval.codex_humaneval.run_eval --data_file evaldata/eval/codex_humaneval/HumanEval.jsonl.gz --eval_pass_at_ks 1 5 10 20 --unbiased_sampling_size_n 20 --temperature 0.1 --save_dir nochat --use_vllm --model allenai/tulu-2-dpo-7b --tokenizer_name_or_path allenai/tulu-2-dpo-7b`
{'pass@1': 0.23140243902439026, 'pass@5': 0.27792513025749455, 'pass@10': 0.29802788014673537, 'pass@20': 0.3170731707317073

HumanEval chat:
`python -m eval.codex_humaneval.run_eval --data_file /data/niklas/sgpt2/evaldata/eval/codex_humaneval/HumanEval.jsonl.gz --eval_pass_at_ks 1 5 10 20 --unbiased_sampling_size_n 20 --temperature 0.1 --save_dir humanevalpackchatoldfix --use_vllm --model allenai/tulu-2-dpo-7b --tokenizer_name_or_path allenai/tulu-2-dpo-7b --use_chat_format`
`{'pass@1': 0.0, 'pass@5': 0.0, 'pass@10': 0.0, 'pass@20': 0.0}`

HumanEvalPack:
`python -m eval.codex_humaneval.run_eval --data_file /data/niklas/sgpt2/evaldata/eval/codex_humaneval/HumanEval.jsonl.gz --eval_pass_at_ks 1 5 10 20 --unbiased_sampling_size_n 20 --temperature 0.1 --save_dir humanevalpackchatnewfix --use_vllm --model allenai/tulu-2-dpo-7b --tokenizer_name_or_path allenai/tulu-2-dpo-7b --use_chat_format --data_file_hep /data/niklas/sgpt2/evaldata/eval/codex_humaneval/humanevalpack.jsonl`
`{'pass@1': 0.22560975609756098, 'pass@5': 0.2666019312089406, 'pass@10': 0.2845676090125988, 'pass@20': 0.29878048780487804}`


So it's more or less equivalent! In https://arxiv.org/pdf/2402.09906.pdf, we also found it to be ~the same see screenshot:
Screenshot 2024-02-19 at 10 50 51 PM

As it's the more realistic setting, I suggest merging this & maybe use it going forward.

hamishivi:
Thanks for coming back and updating this! The results look good.

hamishivi (approved):

hamishivi:
Could you update the example eval scripts in `scripts/eval` too?

PR #107, opened 2024-02-05
Bump alpaca-eval dependency.

dwadden:
Bump `alpaca-eval` version. Fixes a bug where alpaca-eval sometimes hung during evaluation.

I upgrade `openai` as well because the current alpaca-eval requires `openai >= 1.5.0`. From what I could tell, the only dependence on `openai` was TruthfulQA, which is broken at present anyway.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -123,6 +123,9 @@ RUN pip install packaging
 RUN pip install flash-attn==2.2.2 --no-build-isolation
 RUN pip install -r requirements.txt
 
+# Use v1 of alpaca eval.
+ENV IS_ALPACA_EVAL_2=False
+
 COPY open_instruct open_instruct
 COPY eval eval
 COPY ds_configs ds_configs
--- a/README.md
+++ b/README.md
@@ -38,6 +38,11 @@ If you just want the dependencies for the weight diff script, use:
 pip install -r weight-diff-requirements.txt
 ```
 
+If you'd like to run experiments within a Docker environment, you can create one using:
+```bash
+docker build --build-arg CUDA=11.8.0 --build-arg TARGET=cudnn8-devel --build-arg DIST=ubuntu20.04 . -t <your tag here>
+```
+
 ## Training
 
 ### Dataset preparation
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ protobuf
 # But this PR is not compatible with the latest version of Transformers library (v4.34.0).
 # To incorporate it, we forked the Transformers library and made some changes to make it compatible with the latest version.
 git+https://github.com/yizhongw/transformers.git@left_padding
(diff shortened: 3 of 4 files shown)


hamishivi:
Did you test that the version bump doesn't significantly change scores?

dwadden:
Good point, I'll do a check this evening.

hamishivi:
Also, I think it would be good to edit https://github.com/allenai/open-instruct/blob/main/scripts/eval/alpaca_farm.sh to include the alpacaeval 2 environment variable!

dwadden:
OK, I ran a few tests and the results are a bit different. 
Setup: I ran alpaca-eval on tulu-2-dpo-7b (beaker `01HE8H1MBSVN09ZZ82X6K90NTF`) and tulu-2-dpo-70b (beaker `01HEXKXP0MFM60PT7SY71XXSWD`), using beaker images with `alpaca-eval==0.3.1` (old) and `alpaca-eval==0.5.3` (new).

Results below. Seems like there's a fairly large discrepancy for the 7b model; substantially larger than the standard errors. Should we check with the alpaca folks about what's going on?

| setting        | win_rate | standard_error | n_wins | n_wins_base | n_draws | n_total | discrete_win_rate | mode      | avg_length |
|----------------|----------|----------------|--------|-------------|---------|---------|-------------------|-----------|------------|
| tulu 70b / new |    94.53 |           0.80 |    760 |          43 |       2 |     805 |             94.53 | community |       1433 |
| tulu 70b / old |    92.05 |           0.95 |    740 |          63 |       2 |     805 |                   | community |       1433 |
| tulu 7b / new  |    84.97 |           1.26 |    684 |         121 |       0 |     805 |             84.97 | community |       1446 |
| tulu 7b / old  |    76.27 |           1.50 |    614 |         191 |       0 |     805 |                   | community |       1446 |

hamishivi:
This seems fine to me, actually for the Tulu 2 paper we used the newer GPT-4 judge version, and the older gpt-4 is deprecated. So it's good to update, and the "new" results dave shared match those in the Tulu 2 paper better (~85 for 7b, ~95 for 70b).

yizhongw:
Looks good to me! Just to clarify, I think the results we ran for tulu 2 are still valid? The old results that Dave showed are actually lower than we have in our result sheet. The newer version actually matches better?

hamishivi:
To clarify: the new version matches better because I manually made sure our Tulu 2 results used the newer GPT-4 version (0613), which the pinned older version of AlpacaEval doesn't use it (it explicitly uses GPT-4-0314).

dwadden:
Glad this worked out!

dwadden:
One other thing we talked about was making sure we pin the version of GPT that's being used as an eval. Are we confident this is fixed or should I look into this?

PR #110, opened 2024-02-07
Beaker improvements & HF formatting

hamishivi:
This PR turns the beaker script into a command-line configurable thing, making it easier for people to use it without actually reading the script and understanding what's happening. @dwadden may turn this into something more tightly integrated with beaker-py.

Additionally, I've added support for just applying the HF tokenizer in the eval chat formatting. This required a little refactoring to pass the tokenizer into the chat template. This allows us just to reuse the chat template given in an HF tokenizer (saving us time in implementing and testing any given chat format, hopefully).

### DIFF
--- a/scripts/submit_eval_jobs.py
+++ b/scripts/submit_eval_jobs.py
@@ -14,8 +14,8 @@ parser.add_argument("--model_name", type=str, default="hf-opt-7B")
 parser.add_argument("--location", type=str, default=None)
 parser.add_argument("--beaker_subfolder", type=str, default=None)
 parser.add_argument("--cluster", type=str, default="ai2/allennlp-cirrascale")
-parser.add_argument("--num_gpus", type=int, default=1)
 parser.add_argument("--is_tuned", action="store_true")
+parser.add_argument("--use_hf_tokenizer_template", action="store_true")
 args = parser.parse_args()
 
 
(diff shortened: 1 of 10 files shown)


yizhongw:
This looks nice!

I guess the scripts in `scripts/eval` should still work since the refactored eval code can still accept a specified template function? Did you test them?

scripts/submit_eval_jobs.py
```diff
@@ -1,25 +1,35 @@
...
     default_yaml = f.read()
 d1 = yaml.load(default_yaml, Loader=yaml.FullLoader)
 
-# cluster = "ai2/general-cirrascale"
-cluster = "ai2/allennlp-cirrascale"
-# cluster = "ai2/general-cirrascale-a100-80g-ib"
-# cluster = "ai2/prior-elanding"
-num_gpus = 1
+cluster = args.cluster
+num_gpus = args.num_gpus
```
yizhongw, line 30: Later, the num_gpus will be dynamically changed based on the model size. So, the argument here might be confusing. Shall we move the hard-coded GPUs into some post init functions for these arguments?

scripts/submit_eval_jobs.py
```diff
@@ -39,110 +49,16 @@
     "alpaca_eval",
 ]
 
-# model to evaluate, each in the followng format: model name, their beaker id, checkpoint subfolder
+# format: model name, their beaker id, checkpoint subfolder, tuned or base.
+# or: name, path, None, tuned or base
 models = [
```
yizhongw, line 54: Can we move these models into a spreadsheet or somewhere to make them easily sharable?

scripts/submit_eval_jobs.py
hamishivi: Good point. Maybe for now I'll leave it alone.

scripts/submit_eval_jobs.py
hamishivi: Yup, this is happening: https://beaker.org/ws/ai2/adapt-models/datasets?sort=created:descending&results=false&committed=true
@dwadden is thinking about how to share this better too.

hamishivi:
Yup, I've given this some quick tests.

natolambert:
I'd say adding tests with all this chat template stuff would be great, maybe I can do this next week to get started here.

hamishivi:
Tests would be good, maybe in a future PR :)

PR #111, opened 2024-02-13
Trust HF OLMo models.

dwadden:
Previously, attempting to load OLMo models in a Beaker job raised an error because these models run custom code. This fixes it by setting `trust_remote_code=True` for OLMo models.

At some point we should just unify the command line parsers for the different eval tasks and allow the user to pass `--trust_remote_code` as a command line flag; this will take a bit of a refactor.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -35,4 +35,4 @@ alpaca-eval==0.3.1
 flask
 vllm 
 openpyxl
-hf_olmo       # Needed to work with HF OLMo models.
+ai2-olmo       # Needed to work with HF OLMo models.
--- a/requirements.txt
+++ b/requirements.txt
@@ -35,3 +35,4 @@ alpaca-eval==0.3.1
 flask
 vllm 
 openpyxl
+hf_olmo       # Needed to work with HF OLMo models.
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -35,3 +35,4 @@ alpaca-eval==0.3.1
 flask
 vllm 
 openpyxl
+ai2-olmo       # Needed to work with HF OLMo models.
```
hamishivi: I'd rather not add this to the requirements, since this repo should be able to run models beyond just olmo. If someone specfically wants to run olmo, they should install this themself. Adding an error message where if they try to load olmo without this package sems fine, but I think HF should already do that?

dwadden:
Hmm I see. Does adding this dependency actually break things for other HF models?
The reason to add this would be: if we wanna use open-instruct internally to run evals of OLMo models, then it makes sense to be able to run OLMo models by default. But if this breaks other things, agreed this doesn't work.

hamishivi:
I don't think it does, but I don't want extra packages that might mess with someone's environment that aren't actually required...

dwadden:
Got it. I removed the requirement and added a note in the `readme`, does this seem reasonable?

hamishivi:
Dockerfile, is fine, lgtm!

PR #112, opened 2024-02-13
Update TruthfulQA evaluation using retrained LLaMa judge models.

yizhongw:
Please find the details about how the llama-based truthfulqa judge models are trained [here](https://github.com/allenai/truthfulqa_reeval).

### DIFF
--- a/eval/truthfulqa/run_eval.py
+++ b/eval/truthfulqa/run_eval.py
@@ -19,7 +19,7 @@ from eval.truthfulqa.utilities import (
     format_best,
     set_columns,
 )
-from eval.truthfulqa.metrics import run_gpt3_classifier_eval, run_hf_classifier_eval, MC_calcs
+from eval.truthfulqa.metrics import run_gpt_classifier_eval, run_hf_classifier_eval, MC_calcs
 from eval.truthfulqa.configs import BEST_COL, ANSWER_COL, INCORRECT_COL
 
 
@@ -185,7 +185,7 @@ def run_hf_model(questions, model, tokenizer, tag, preset="qa", batch_size=1, ma
     prompts = [
         format_prompt(questions.loc[idx], preset, format='general') for idx in questions.index
     ]
-    if chat_formatting_function:
+    if chat_formatting_function is not None:
         for idx, prompt in enumerate(prompts):
             messages = [{"role": "user", "content": prompt}]
             prompts[idx] = chat_formatting_function(messages, tokenizer, add_bos=False)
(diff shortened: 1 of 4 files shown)


eval/truthfulqa/run_eval.py
```diff
@@ -15,12 +15,11 @@
 )
 from eval.truthfulqa.utilities import (
     format_prompt,
-    format_prompt_with_answer_strings,
     split_multi_answer,
     format_best,
     set_columns,
 )
-from eval.truthfulqa.metrics import run_gpt3_classifier_eval, MC_calcs
+from eval.truthfulqa.metrics import run_gpt3_classifier_eval, run_hf_classifier_eval, MC_calcs
```
hamishivi: nit: rename `run_gpt3_classifier_eval` to `run_gpt_classifier_eval` since gpt3 is deprecated?

eval/truthfulqa/run_eval.py
```diff
@@ -60,9 +59,6 @@ def run_chatgpt(questions, engine, tag, preset='qa', batch_size=1, cache_path=No
         temperature=0.0
     )
     assert len(responses) == len(instances)
-
-    for idx, response in zip(questions.index, responses):
-        questions.loc[idx, tag] = trim_answer(response["output"])
```
hamishivi, line 65: why is this removed?

scripts/eval/trutufulqa.sh
```diff
@@ -2,74 +2,74 @@
...
-    --gpt_judge_model_name curie:ft-allennlp:gpt-judge-2023-07-26-09-37-48 \
-    --gpt_info_model_name curie:ft-allennlp:gpt-info-2023-07-26-11-38-18 \
+    --hf_truth_model_name_or_path allenai/truthfulqa-truth-judge-llama2-7B \
+    --hf_info_model_name_or_path allenai/truthfulqa-info-judge-llama2-7B \
     --eval_batch_size 20 \
     --load_in_8bit
 
 
-# Evaluating Tulu 7B model using chat format, getting the judge and info scores and multiple choice accuracy
+# # Evaluating Tulu 7B model using chat format, getting the truth and info scores and multiple choice accuracy
```
hamishivi, line 27: maybe link to https://github.com/allenai/truthfulqa_reeval when it is public?

eval/truthfulqa/run_eval.py
```diff
@@ -189,26 +185,29 @@ def run_hf_model(questions, model, tokenizer, tag, preset="qa", batch_size=1, ma
     prompts = [
         format_prompt(questions.loc[idx], preset, format='general') for idx in questions.index
     ]
-    if use_chat_format:
-        chat_formatting_function = dynamic_import_function(args.chat_formatting_function)
+    if chat_formatting_function:
```
hamishivi: I guess all functions are truthy, but maybe an explicit `is not None` check would be nicer?

hamishivi (approved):
Could run it locally, seems good to me! I have some minor nitpicks but nothing blocking a merge. Seems like the results I got testing Tulu 2 SFT match up with the old results reasonably (within ~1pt).

eval/truthfulqa/run_eval.py
yizhongw: Good catch. So, I removed the postprocessing for all chat models. I found this could significantly improve llama2-chat, as llama2-chat will output `\n\n` that will lead to trimming in the postprocessing of answers. I remove this also for chatgpt.

PR #119, opened 2024-02-21
Fix tokenizer setup in eval.

hamishivi:
Fix tokenizer issue with vllm, have to load manually.
Also add budget to the eval job script, and send to multiple clusters.

### DIFF
--- a/eval/toxigen/run_eval.py
+++ b/eval/toxigen/run_eval.py
@@ -12,7 +12,7 @@ from transformers import AutoTokenizer, AutoModelForSequenceClassification
 
 from eval.utils import (
     generate_completions,
-    load_hf_lm_and_tokenizer,
+    load_hf_lm,
     query_openai_chat_model,
     load_hf_tokenizer,
 )
@@ -95,7 +95,7 @@ def main(args):
             del model  # free up GPU memory to load the classifier later.
         else:
             print("Loading model and tokenizer for generations...")
-            model, tokenizer = load_hf_lm_and_tokenizer(
+            model = load_hf_lm(
                 model_name_or_path=args.model_name_or_path,
                 tokenizer_name_or_path=args.tokenizer_name_or_path if args.model_name_or_path else args.model_name_or_path,
                 load_in_8bit=args.load_in_8bit,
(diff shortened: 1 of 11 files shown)


yizhongw:
Could you provide some context on why manual tokenizer loading is necessary for vllm? The changes look good to me though.

hamishivi:
Ah, it strictly isn't, but I feel it's preferable to expose it rather than rely on vllm internals? Happy to revert it back. There are some other small fixes in this PR too (bug in toxigen eval).

eval/toxigen/run_eval.py
```diff
@@ -61,9 +62,13 @@ def main(args):
     # Generate the outputs
     if args.model_name_or_path:
         prompts = []
+        tokenizer = load_hf_tokenizer(
```
yizhongw, line 65: Yeah, I see the point that a tokenizer is needed for creating the prompts now since the template function now needs the tokenizer argument.

I just feel the current logic for tokenizer/model loading and data loading/processing is not very consistent in the different evals. 

Probably it's better to separate the model loading, and tokenizer loading in the `load_hf_lm_and_tokenizer` function. For each eval, we load the tokenizer first, then the model (dependent on whether it's vllm), both of which happen before the data processing (which relies on the tokenizer).

eval/utils.py
```diff
@@ -291,6 +291,29 @@ def load_hf_lm_and_tokenizer(
         
     return model, tokenizer
 
+def load_hf_tokenizer(
```
yizhongw, line 270: Since we now have the tokenizer loading function separate, I think it's better to change the other function to only load the model.

scripts/submit_eval_jobs.py
```diff
@@ -13,9 +13,10 @@
 parser.add_argument("--model_name", type=str, default="hf-opt-7B")
 parser.add_argument("--location", type=str, default=None)
 parser.add_argument("--beaker_subfolder", type=str, default=None)
-parser.add_argument("--cluster", type=str, default="ai2/allennlp-cirrascale")
+parser.add_argument("--cluster", nargs='+', default=["ai2/allennlp-cirrascale", "ai2/general-cirrascale", "ai2/general-cirrascale-a100-80g-ib", "ai2/mosaic-cirrascale-a100", "ai2/s2-cirrascale-l40"])
```
yizhongw, line 16: Great, thanks!

eval/toxigen/run_eval.py
hamishivi: This is a good point, I'll make this change and re-request review.

yizhongw: Seems it's done. Shall I merge it?

hamishivi: Let me test it, just uploading a docker image with this code now. If it works, i'll merge?

yizhongw: Sounds good! Thanks!

PR #120, opened 2024-02-21
Add flag indicating which flan version to sample.

dwadden:
Tulu v1 and v2 used different mixes for flan; this adds a flag to request one or the other.

Having this will be useful if I need to run a beaker job that creates a big flan pretraining mix.


yizhongw:
Looks great to merge.

PR #121, opened 2024-02-23
Add model revision flag for finetuning

dwadden:
Makes it possible finetune different checkpoints of the same HF model; this is useful for experimentation with OLMo.

I think I found all the places where the revision flag would need to be added but it's probably good for @yizhongw or @hamishivi to double-check.

### DIFF
--- a/open_instruct/finetune.py
+++ b/open_instruct/finetune.py
@@ -66,6 +66,12 @@ def parse_args():
         required=False,
     )
     parser.add_argument(
+        "--model_revision",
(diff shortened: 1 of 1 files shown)


hamishivi:
This is just in the finetuning script, did we want to add this to eval scripts too?

hamishivi:
Also, will it always be the case that the tokenizer revision == the model revision?

hamishivi:
And yeah, the argument parser thing makes sense. I'm also going to split the tokenizer and model loading for evals later today, so refactoring should probably wait at least until next week.

PR #126, opened 2024-02-27
Subfolder evaluation

dwadden:
Fixes some bugs to enable evaluation of models stored in subfolders. Changes are as follows:

- Fixes inconsistency between `model_name_or_path` vs. `model` as a command line arg, and similarly for `tokenizer_name_or_path` vs. `tokenizer`; the `_name_or_path` version are used in the actual eval scripts, but in the `yaml` template and and at some places in the `submit_eval_jobs` script, `model` and `tokenizer` are used instead. I made it consistent to always use `model_name_or_path`, and `tokenizer_name_or_path`. Previously, some of the code attempted to change `model` to a subdirectory, which caused an error.
- Makes it so that passing a subdirectory changes the `model_name_or_path` but not `tokenizer_name_or_path`. We want this because the training code by default doesn't dump a separate tokenizer for each subdirectory, it only saves one to the top level. The current behavior caused an error because it couldn't find the tokenizer.
- Explicitly imports `OlmoTokenizerFast` in `utils.py`. Without this import, HF can't find the OLMo tokenizer and throws an error.

As far as I can tell things work now; I ran some evals and the code seems to properly find model subdirectories and is able to evaluate.

### DIFF
--- a/eval/utils.py
+++ b/eval/utils.py
@@ -7,6 +7,7 @@ import os
 from importlib import import_module
 from transformers import StoppingCriteria
 from eval.dispatch_openai_requests import dispatch_openai_chat_requesets, dispatch_openai_prompt_requesets
+import warnings
 
 
 class KeyWordsCriteria(StoppingCriteria):
@@ -275,6 +276,13 @@ def load_hf_tokenizer(
         token=os.getenv("HF_TOKEN", None),
     ):
         from transformers import AutoTokenizer
+
+        # Need to explicitly import the olmo tokenizer.
+        try:
+            from hf_olmo import OLMoTokenizerFast
+        except ImportError:
+            warnings.warn("OLMo not installed. Ignore if using a different model.")
+
         if not tokenizer_name_or_path:
             tokenizer_name_or_path = model_name_or_path
         try:
--- a/scripts/submit_eval_jobs.py
+++ b/scripts/submit_eval_jobs.py
@@ -107,7 +107,7 @@ for model_info, experiment_group in itertools.product(models, experiment_groups)
             --data_dir /data/bbh \
             --save_dir /output/ \
             --use_vllm \
-            --model /model \
+            --model_name_or_path /model \
             --tokenizer_name_or_path /model \
             --max_num_examples_per_task 40 \
             --no_cot \
@@ -120,7 +120,7 @@ for model_info, experiment_group in itertools.product(models, experiment_groups)
             --data_dir /data/bbh \
             --save_dir /output/ \
             --use_vllm \
-            --model /model \
+            --model_name_or_path /model \
             --tokenizer_name_or_path /model \
             --max_num_examples_per_task 40 \
             --use_chat_format \
@@ -133,7 +133,7 @@ for model_info, experiment_group in itertools.product(models, experiment_groups)
             --max_num_examples 200 \
             --save_dir /output/ \
             --use_vllm \
-            --model /model \
+            --model_name_or_path /model \
             --tokenizer_name_or_path /model \
             --n_shot 8 \
             --no_cot \
@@ -147,7 +147,7 @@ for model_info, experiment_group in itertools.product(models, experiment_groups)
             --max_num_examples 200 \
             --save_dir /output/ \
             --use_vllm \
-            --model /model \
+            --model_name_or_path /model \
             --tokenizer_name_or_path /model \
             --n_shot 8 \
             --use_chat_format \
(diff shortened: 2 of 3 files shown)


eval/utils.py
```diff
@@ -275,6 +276,13 @@ def load_hf_tokenizer(
         token=os.getenv("HF_TOKEN", None),
     ):
         from transformers import AutoTokenizer
+
+        # Need to explicitly import the olmo tokenizer.
+        try:
```
hamishivi, line 281: I don't really like having this here... is a bit annoying to have this warning if you don't care about olmo. Is there another way we can do this?

scripts/submit_eval_jobs.py
```diff
@@ -120,7 +125,7 @@
             --data_dir /data/bbh \
             --save_dir /output/ \
             --use_vllm \
-            --model /model \
+            --model_name_or_path /model \
```
hamishivi, line 128: How come using `--model` was fine before? The scripts all use `--model_name_or_path`... Just curious...

dwadden: Yeah, this was a weird one. The error shows up if you specify a `--beaker_subfolder`. If you do this, then the code performs a [re.search](https://github.com/allenai/open-instruct/blob/main/scripts/submit_eval_jobs.py#L262) for `model_name_or_path` so that it can replace this path with a path to the subfolder. If the config uses `model` rather than `model_name_or_path`, then asking for the first `group` found by the `re.search` throws an error; that's how I found it. My guess is that something must have changed from the last time you guys were doing subfolder evals for tulu v1?

But I'm still confused actually. For instance, take a look at the command for this [gsm eval job](https://beaker.org/ex/01HQNWZ8CAAYXJ0NJ2PNBDPY5S/tasks/01HQNWZ8CE24V2MH3TV7HEQ1C1/job/01HQNWZJXACM3T6EESHGX2MYBB). There's no `--model_name_or_path` specified. Looking at the script [gsm/run eval.py](https://github.com/allenai/open-instruct/blob/main/eval/gsm/run_eval.py#L256), this should throw an error, but it doesn't seem to.

So I'm actually not sure why it doesn't break as currently written. I can try to debug more but thought I'd check with you first.

eval/utils.py
dwadden: Sure. I was imitating the warning in the [https://github.com/allenai/open-instruct/blob/main/open_instruct/finetune.py#L43](trainer) if OLMo isn't installed but we can do it another way.
- Option 1: Just try to import and fail silently.
- Option 2: Only import if `"olmo" in model_name_or_path`. Not perfect because it requires the user to include `olmo` in the model name, but that's already done [here](https://github.com/allenai/open-instruct/blob/main/scripts/submit_eval_jobs.py#L331) anyhow. We should probably add a note about this somewhere; can do that later.

Any preference?

scripts/submit_eval_jobs.py
hamishivi: Yeah, I'm not sure whats going on here rn. @yizhongw might have more insight. I agree the subfolder stuff seems broken as-is, though.

eval/utils.py
hamishivi: Yeah, I guess since I did it for finetuning its okay. Hopefully at some point we can remove these checks (maybe if/when olmo becomes a first-class HF model, instead of relying on importing code).

dwadden: OK, I'll just leave it. Annoying but not terrible.

scripts/submit_eval_jobs.py
dwadden: OK. Let's wait to merge till we hear back from @yizhongw to make sure I haven't subtly introduced a bug?

yizhongw: Sorry for the delay. I just checked and it looks good to me!

yizhongw (approved):

scripts/submit_eval_jobs.py
dwadden: Just closing the loop on why the pre-pr setup wasn't causing an error: Yizhong pointed out that it's because argparse [allows abbreviations](https://docs.python.org/3/library/argparse.html#allow-abbrev) by default.
So, the parsing code was doing error recovery for us. Still it's probably safer to have the argument names be consistent.

hamishivi: Awesome, I didn't know about that feature of argparse! agreed that we should keep argument names consistent.

PR #129, opened 2024-03-04
Run all evals as jobs within a single Beaker experiment

dwadden:
I find it convenient to have all evals for a given model associated with a single [experiment](https://beaker.org/ex/01HQQ5VX550JASBSF634JV1ZXQ/tasks/01HQQ5VX5AS4M7XCDK6GMF8NCT/job/01HQQ5XZ41FWR2RDVX12N2C8TC).

This PR implements this. It also has the added side effect only creating 1 new template `yaml` instead of `n_tasks` of them. I'll comment the code below to explain what's happening.

I've tested this (for instance the beaker job above) and it seems to work. This is also purely personal preference, so if you don't like this we can just make it a separate script called `submit_eval_experiment` or something; whatever you guys prefer.

### DIFF
--- a/scripts/submit_eval_jobs.py
+++ b/scripts/submit_eval_jobs.py
@@ -56,32 +56,27 @@ experiment_groups = [
     "alpaca_eval",
 ]
 
-# format: model name, their beaker id, checkpoint subfolder, tuned or base.
-# or: name, path, None, tuned or base
-models = [
-    # llama1 models
-    # ("llama1-7B", "01HCCBK1MYKXKQC0C6CSVW1F22", None, "vanilla_lm"),
-
-    # other causal models
-    # ("hf-opt-7B", "facebook/opt-6.7b", None, "vanilla_lm"),
-    # ("finetuned_opt", "01H13EBXSADXXJCRERART90ZKJ", None, "tuned_lm"),
-    (args.model_name, args.location, args.beaker_subfolder, model_type),
-]
+model_info = (args.model_name, args.location, args.beaker_subfolder, model_type)
 
 #--------------- experiments about number of supervision tasks -------------------------
 
 # for experiment_group, model_info in itertools.product(experiment_groups, models):
-for model_info, experiment_group in itertools.product(models, experiment_groups):
-    print(f"Submitting {experiment_group} for model: {model_info[0]}")
-    d = copy.deepcopy(d1)
 
-    model_name = model_info[0] + f"_{model_info[2]}" if model_info[2] is not None else model_info[0]
+d = copy.deepcopy(d1)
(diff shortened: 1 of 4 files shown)


scripts/submit_eval_jobs.py
```diff
@@ -56,32 +56,27 @@
...
-models = [
-    # llama1 models
-    # ("llama1-7B", "01HCCBK1MYKXKQC0C6CSVW1F22", None, "vanilla_lm"),
-
-    # other causal models
-    # ("hf-opt-7B", "facebook/opt-6.7b", None, "vanilla_lm"),
-    # ("finetuned_opt", "01H13EBXSADXXJCRERART90ZKJ", None, "tuned_lm"),
-    (args.model_name, args.location, args.beaker_subfolder, model_type),
-]
+model_info = (args.model_name, args.location, args.beaker_subfolder, model_type)
```
hamishivi, line 61: Could you leave the comments for hf/beaker/etc examples? Just to make it extra clear what these arguments should be.

hamishivi: We also lose the ability to batch-submit models (via the list) here, but this is probably fine.

scripts/submit_eval_jobs.py
```diff
@@ -56,32 +56,27 @@
...
 
 #--------------- experiments about number of supervision tasks -------------------------
 
 # for experiment_group, model_info in itertools.product(experiment_groups, models):
-for model_info, experiment_group in itertools.product(models, experiment_groups):
+
+d = copy.deepcopy(d1)
```
hamishivi, line 67: Did you alter the default yaml config at all? I'll check this in running this myself, but just making sure.

hamishivi (commented):
Seems good to me! Will approve and merge once I get the chance to run it myself.

As a note, you can restart/cancel tasks in an experiment independent of the other tasks, right? e.g. if an eval task needs to be restarted or rerun.

dwadden:
You can cancel jobs independently. To do this, you click on the job you way to cancel, copy the hash after the last slash in the URL (that's the job id), and do `beaker job cancel [job-id]`.

You can only restart at the experiment level rather than job level. By default, when you call `beaker experiment resume [experiment-id]`, Beaker will check for any failed or stopped jobs in the experiment and re-run them, which seems reasonable? We could ask the beaker team to add functionality to restart individual jobs if you think this is important though.

hamishivi:
Do you mean that if you hit restart, it only restarts jobs that have failed or were cancelled? That's fine for now, just making sure that you don't rerun everything if one task fails for an ephemeral reason (e.g., RAM issues).

dwadden:
Correct, it only restarts failed or canceled jobs and leaves the completed ones untouched.

hamishivi (approved):
Tested and works.

PR #137, opened 2024-04-05
Tokenizer path fix

dwadden:
Previously, `model_name_or_path` was replaced twice, rather than replacing `model_name_or_path`.

I also added an `olmo` flag so that the code will do the right thing for OLMo models in cases where "olmo" isn't in the directory name.


hamishivi (approved):
lgtm

PR #138, opened 2024-04-05
Script to collect metrics from Beaker eval job.

dwadden:
Give it the ID of a beaker job launched with `submit_eval_jobs.py` and it will grab the results and dump in a `json`.

### DIFF
--- a/scripts/collect_eval_results.py
+++ b/scripts/collect_eval_results.py
@@ -1,9 +1,12 @@
 """
 Usage example:
 python scripts/collect_eval_results.py \
-    --experiment_id 01HTNY22Z9YJZ6HDQBF182BGE6 \
-    --job_suffix _v0-step_1.5T-warmup_true-steps_50B_04052024 \
-    --output_file metrics.json
+    --experiment_id 01HV0P4E3MW9211HX0JEKM0PXM \
+    --job_suffix _tulu2_13b_dpo_ultrainteract_04082024 \
+    --output_file scratch/metrics.json \
+    --task_order gsm_cot gsm_direct toxigen alpaca_eval \
+    --print_table \
+    --table_file scratch/metrics.tsv
 """
 
 import beaker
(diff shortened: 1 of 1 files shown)


hamishivi:
Could you add a copy-paste-friendly print in a configurable task order? Would be very useful to me!!

scripts/collect_eval_results.py
```diff
@@ -0,0 +1,103 @@
...
+    )
+    parser.add_argument(
+        "--output_file", type=Path, help="Output file for collected results."
+    )
+
+    return parser
+
+
+def get_summary_metric(metrics, eval_task):
+    # TODO we need summary metrics for XSTest and AlpacaEval
```
dwadden: @hamishivi what should I use for XSTest and AlpacaEval?

dwadden:
Sure, what were you thinking exactly? Just print out the summary metrics I wrote in line 41, or all metrics? I can save as a pandas series, that's pretty easy to copy/paste.

scripts/collect_eval_results.py
hamishivi: XSTest: gpt4_label. refusal_prf.f1 (using dot notation for nested dicts)
AlpacaEval: win_rate (although I need to test what this is called for AlpacaEval2)

hamishivi:
Just printing out the summary metrics, or some easy to configure set of metrics. Basically just something that lets you copy to a row of something like the tulu results spreadsheet easily.

dwadden:
Do you have an eval job where all tasks have successfully completed that I can use to debug? None of my recent alpaca-evals worked so I can't check this.

hamishivi:
https://beaker.org/ex/01HV0P4E3MW9211HX0JEKM0PXM/tasks/01HV0P4E3T3BPZRPK7DYFJ80V3/job/01HV0P4FF8CY8TX053ZRF97G45 here ya go!

dwadden:
Thanks! Oh also -- what summary metric should we use for `ifeval`?

hamishivi:
Either loose or strict accuracy works for ifeval.

dwadden:
OK, all set. I added the new evals (ifeval, alpaca, etc); they all seem covered now.
You can now specify `--table_file` to dump summary metrics to a `tsv`, and / or `--print_table` to print to console.
You can control which tasks are printed, and in what order, with `--task_order`.

scripts/collect_eval_results.py
```diff
@@ -0,0 +1,190 @@
...
+        summary_table = make_summary_table(metrics_all["summary"], args)
+
+        if args.print_table:
+            print(summary_table)
+        if args.table_file is not None:
+            summary_table.to_csv(
+                args.table_file, sep="\t", index=True, float_format="%0.4f"
+            )
+
+    json.dump(metrics_all, open(args.output_file, "w"), indent=2)
```
hamishivi:
```suggestion
    if args.output_file:
        json.dump(metrics_all, open(args.output_file, "w"), indent=2)
```

hamishivi (approved):
Looks good, worked from a quick test. one minor note is it might be nice to allow printing without dumping to file.

scripts/collect_eval_results.py
dwadden: Done.

PR #142, opened 2024-04-08
Resize token embeddings of the base model before merging

notoookay:
Hi, After I trained llama2-7b model with lora, I tried to merge it, it seems like resizing token embeddings of the base model should before merging, I reordered the process, and it worked. Can you check the correctness?


hamishivi:
Thanks, this seems to work fine.

hamishivi (approved):

PR #144, opened 2024-04-10
add the parameter "tokenizer_mode" when using the vllm.LLM() function

ShuheWang1998:
Some tokens have different "tokenized ids" in fast and slow modes, so we need to specify the parameter "tokenizer_mode" while loading the model with vllm.


yizhongw (approved):
Cool, this seems reasonable to be since some of the tokenizers don't support fast mode. I missed adding that argument for alpaca eval. Thanks for finding out this issue and the contribution!

PR #146, opened 2024-04-12
Add --experiments argument to submit_eval_jobs.py

OyvindTafjord:
Allow specifying which experiments to run in `submit_eval_jobs.py`.


dwadden (approved):
lgtm

PR #154, opened 2024-04-26
Add MBPP+ and HumanEval+

hamishivi:
Adds humaneval+ data and mbpp+ eval from the [evalplus project](https://github.com/evalplus/evalplus), to give us some more rigorous code evaluations.

Thanks to @Nanami18 (Muru) for writing most of this, I just tweaked it for open instruct. Currently a draft until I properly test this. 

HumanEval plus is easy - just swap out the data source to the HumanEval+ released data, and it should work! It even works nicely with HumanEvalPack for better evaluation of instruction-tuned models. MBPP+ uses data from huggingface, so it has a slightly different setup.

### DIFF
--- a/eval/codex_humaneval/run_eval.py
+++ b/eval/codex_humaneval/run_eval.py
@@ -44,7 +44,8 @@ def main(args):
         else:
             print(f"Could not find HumanEvalPack file at {args.data_file_hep}, which will result in significantly worse performance. You can download it at https://hf.co/datasets/bigcode/humanevalpack/blob/main/data/python/data/humanevalpack.jsonl")
             instructions_dict = None
-            answer = "Here is the completed function:\n\n\n"
+            answer = "Here is the completed function:\n\n\n```python\n"
+            stop_sequences.append("\n```")
 
         def apply_chat_format(tokenizer, inst, suffix):
             messages = [{"role": "user", "content": inst}]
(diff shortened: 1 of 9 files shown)


hamishivi:
@yizhongw what do you think is best here? I compared scores on llama-3-8b-instruct to the official [evalplus leaderboard](https://evalplus.github.io/leaderboard.html) and got the following results:

HumanEval+ (w/ HEP): 55.5 (ours) vs 56.7 (theirs)
MBPP: 53.6 (ours) vs 59.3 (theirs)

The lower MBPP score is likely due to differences in prompting formats... we could align them, but unsure if worth while. Another worth noting is that these scores are pass@1 w/ greedy decoding, while for original humanEval, we did temperature sampling and estimated pass@1, pass@10. We could just do this for these (and currently thats what the eval scripts say to do), which is further from the leaderboard but also better matches how MBPP/HumanEval are typically run (e.g., in the codellama paper).

hamishivi:
Okay, I mucked around and better matched the prompt setup used for the evalplus leaderboard. To keep old results consistent, I added an option: `use_evalplus_prompt`. Setting this flag will set using the prompt that matches evalplus, as opposed to our 'home-grown' solution.
Some results with llama-3-8b-instruct, pass@1 with greedy decoding:

original: 53.6%
`use_evalplus_prompt`: 58.3%
official leaderboard: 59.3%

I checked with @yizhongw , and I'm just going to merge this so we can finish up this uplift.

PR #158, opened 2024-04-30
Add trailing backslash in GSM eval command.

dwadden:
Without this, when we append `--stop_at_double_newline`, that flag goes on its own line and bash interprets it as a separate command.


OyvindTafjord (approved):
LGTM

PR #160, opened 2024-04-30
Wandb updates: Separate DPO logging from IFT + add team option

natolambert:
Will test this shortly.

### DIFF
--- a/open_instruct/dpo_tune.py
+++ b/open_instruct/dpo_tune.py
@@ -260,6 +260,12 @@ def parse_args():
         action='store_true',
         help='Trust remote code when loading pretrained models and tokenizers. Use only when you trust the remote code.',
     )
+    parser.add_argument(
+        '--wandb_entity', 
+        type=str,
+        default=None,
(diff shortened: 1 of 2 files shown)


hamishivi:
Did you end up testing this?

natolambert:
@hamishivi yup, just launched a bogus job to immediately error post wandb init, but here's my new project https://wandb.ai/natolambert/open_instruct_dpo

natolambert:
Also, with `--wandb_entity rlhf-llm-dev`
https://wandb.ai/rlhf-llm-dev/open_instruct_dpo

hamishivi (approved):

hamishivi:
LGTM! Seems like your run hit some error we might need to work out at some point, but unrelated to this change.

natolambert:
Yeah @hamishivi there was a llama 2 dpo tokenizer error that I was confused by, the assert was failing, but I commented it out to just get things running quickly to make sure this works.

Will test more soon.

PR #162, opened 2024-05-09
Refactor GPU and batch size adjustment code.

dwadden:
I refactored out the logic to adjust the batch size and number of GPU's into their own functions. It reproduces the logic we had before and also coves the case where the user explicitly passes a `--batch_size_reduction` or `--gpu_multiplier` flag. I did some spot-checks that the auto-generated configs don't change for models with the name `7B`, `13B`, `30B`, `65B`.


hamishivi (approved):
LGTM, so long as you've given it a quick test.

PR #177, opened 2024-06-24
Llama 3 evals

hamishivi:
Adds `additional_stop_sequence` to most eval scripts to allow for an additional stop token (primarily useful for llama 3 instruct). I also merged from main so there are random other commits in here.


natolambert (approved):
LGTM.
Can you put an example of the llama 3 stop sequence somewhere?

hamishivi:
Yup, it's `<|eot_id|>`. I'll add a comment.

PR #183, opened 2024-07-10
Configs for DPO; some repo organization

natolambert:
FYI @hamishivi, will request review when done + tested on my end


hamishivi (approved):
Seems good, and I tested and it worked fine.

PR #191, opened 2024-07-12
remove duplicated load_in_xbit call

hamishivi:
Newer transformers doesnt allow passing a quantization config and `load_in_4bit` together, so this is just removing the second duplicated flag.


natolambert (approved):
LGTM! Yeah this stuff is 😕

PR #193, opened 2024-07-15
Testing automatic image building

natolambert:
Note: dockerfile changed to improve debug time. Do not merge.
Will fail now without correct secrets.

Internal reference: https://github.com/allenai/docker-images


natolambert:
I ran some training on the new auto image, seemed fine. 
Need to lower the LR on the default SFT config, but that's unrelated.

PR #200, opened 2024-07-22
Remove unnecessary outdated files.

yizhongw:


yizhongw:
I removed the gradio demo files, the safe saver for old transformer versions, and merged the encoding template of SuperNI into the single data reformmating file.

Should I also remove `open_instruct/finetune_trainer.py`? It seems no one is using it, and it is hard to maintain two different training codes.

natolambert (approved):
Nice. I don't think I've touched any of these so happy to see them go.

PR #201, opened 2024-07-22
Minor qlora fix

hamishivi:
`load_in_4bit` now errors when its used with a quantisation config, so this fixes that. Also change the DPO script to use the "train" split. In the future we should make that configurable.


natolambert (approved):
seems fine, I remember seeing this elsewhere too.

PR #204, opened 2024-07-24
Add online trainers

vwxyzjn:
The online trainers are ready for review! The docs are available here if you want to try: https://github.com/allenai/open-instruct/blob/online-trainers/docs/algorithms/online_dpo.md

Check out this wandb report: https://wandb.ai/ai2-llm/open_instruct_internal/reports/PPO-vs-online-DPO--Vmlldzo5MzM3NDU0

https://github.com/user-attachments/assets/da8a7144-bff1-4a76-b632-b8acc65e0b50

Implemented auto resume as well, but only tested with small models. Larger models may take more time to save and is blocked by https://github.com/allenai/beaker/issues/5420

image


open_instruct/online_sft_trainer.py
```diff
@@ -0,0 +1,534 @@
...
+        # trainer state initialization
+        episode = 0
+        for update in range(1, args.num_total_batches + 1):
+            episode += 1 * args.batch_size
+            self.lr_scheduler.step()
+            data = next(iter_dataloader)
+            with torch.no_grad():
+                queries = data["input_ids"].to(device)
+                # repeat interleave [q1, q2, q3] -> [q1, q1, q1, q2, q2, q2, q3, q3, q3]
+                queries = queries.repeat_interleave(args.num_generation_per_prompt, 0)
```
vwxyzjn: Here is how the online SFT logic works:

we first repeat interleave the queries like `[q1, q2, q3] -> [q1, q1, q1, q2, q2, q2, q3, q3, q3]`. This means `queries[0] == queries[1]`.

open_instruct/online_sft_trainer.py
```diff
@@ -0,0 +1,534 @@
...
+                    best_idxes + torch.arange(args.local_batch_size, device=device) * args.num_generation_per_prompt
+                )
+                worst_idxes_offset = (
+                    worst_idxes + torch.arange(args.local_batch_size, device=device) * args.num_generation_per_prompt
+                )
+                best_query_responses = query_responses[best_idxes_offset]
+                # worst_query_responses = query_responses[worst_idxes_offset] TODO: maybe interesting to see the worse responses.
+                best_scores = scores[best_idxes_offset]
+                worst_scores = scores[worst_idxes_offset]
+                scores_margin = best_scores - worst_scores
```
vwxyzjn: Then we basically find out the indices corresponding to the best scores and get the best query responses for online SFT.

ValentinaPy (approved):
Code looks good to me! I'll test it out soon

nouhadziri (commented):
Looks great Costa, I have some nitpick about `dop_utils.py` for DPO data processing in  L173 which was contributed by Nathan and Jacob I guess. I usually prefer processing the data (in this case concatenating chosen and rejected ids) before passing the data to DataLoader to avoid looping over the batch. But all good since the tensors are not placed in the GPUs yet

vwxyzjn:
@nouhadziri, that's a good point. We can look into refactoring `dpo_utils.py` after the current projects :) 

Merging as is now. Thanks @nouhadziri and @ValentinaPy for review.

PR #205, opened 2024-07-26
Add rejection sampling script

vwxyzjn:
This PR adds the rejection sampling script. For demo, see 

https://drive.google.com/file/d/1vxhgtWbjXFOWBxOSOpSf1iOnJzKdtjJo/view?usp=sharing


mason.py
natolambert, line 1: Why is this called `mason.py`?

rejection_sampling/generation.py
```diff
@@ -0,0 +1,122 @@
...
+def print_rich_table(df: pd.DataFrame) -> Table:
+    console = Console()
+    table = Table(show_lines=True)
+    for column in df.columns:
+        table.add_column(column)
+    for _, row in df.iterrows():
+        table.add_row(*row.astype(str).tolist())
+    console.print(table)
+
+def main(args: Args, dataset_args: DatasetArgs, gen_args: GenerationArgs):
```
natolambert: I see why you made that minor change in the other PR.

rejection_sampling/generation.py
```diff
@@ -0,0 +1,122 @@
...
+    # prompt | completions
+    # -------|------------
+    # q1     | a1
+    # q1     | a2
+    # q1     | a3
+    # q2     | a1
+    # ...
+    table = defaultdict(list)
+    for output, messages in zip(outputs, ds[dataset_args.dataset_train_split]["messages"]):
+        # TODO: filter out duplicate messages ?
```
natolambert: Remove comment or done :)

natolambert (changes requested):
Some high level comments on organization + long term management. 
* Where should we put this directory? 
* Is it good to be separated? Should this live in `open_instruct/*`
* Should add `rejection_sampling/` dir to makefile for style + add license to the top (make it 1st part). If not, we should add a "work in progress" tag to the top of the readme.
* Some comments / TODOs to remove

rejection_sampling/generation.py
vwxyzjn: just implemented it

mason.py
vwxyzjn: Taken from https://github.com/allenai/mason/tree/main/mason. I had to make some modification to it

vwxyzjn:
Thanks for the review @natolambert, I have moved the script to `open_instruct` and also the bash script to `scripts` folder, which should make more sense.

natolambert (commented):
@vwxyzjn were the files `gpt_eval_judge.py` and `gpt_eval_judge1.py` added by mistake?

vwxyzjn:
Yeah that was added by mistake. I added a note on `mason.py`

natolambert (approved):

PR #217, opened 2024-08-05
add reward training script

vwxyzjn:
Docs are still todos, but it's already kind of usable. I am planning to add more benchmark experiments and trained models to make sure it actually works.


ValentinaPy (approved):
LGTM!

PR #218, opened 2024-08-05
Add infra for multinode jobs + enable data mixer for finetuning

jacob-morrison:


configs/train_configs/sft/tulu3_8b_preview_mix.yaml
```diff
@@ -0,0 +1,29 @@
+# Note, the exact model was trained on TPUs in a different repo
+# Example file for mixing, use exact tulu 2 mixture for replication allenai/tulu-v2-sft-mixture
+model_name_or_path: meta-llama/Meta-Llama-3-8B
+model_revision: main
+use_flash_attn: true
+tokenizer_name: meta-llama/Meta-Llama-3-8B
+use_slow_tokenizer: true
+dataset_mixer:
+  allenai/tulu-v2-sft-mixture-olmo-4096: 1.0
```
natolambert: For none olmo configs :) 
```suggestion
  allenai/tulu-v2-sft-mixture: 1.0
```

open_instruct/finetune.py
```diff
@@ -219,10 +219,10 @@ def main(args: FlatArguments):
             args.dataset_name,
             args.dataset_config_name,
         )
-    elif args.dataset_mixer is not None:
+    elif args.dataset_mixer_list is not None:
```
natolambert, line 231: This looks weird. Isn't this going to fail with the existing configs where we just pass in the dataset mixer and not the list object?

natolambert (changes requested):
Some minor questions on the mixer configs used in `finetune.py` and one config fix. Should double check.

Something we can consider is just passing args to get_datasets, and then we don't need to expose the weird naming to the user in the script and just have it in the get_datasets function (which can just handle local, or one name too).

cc @hamishivi

configs/train_configs/sft/tulu3_8b_preview_mix.yaml
```diff
@@ -0,0 +1,29 @@
...
+model_revision: main
+use_flash_attn: true
+tokenizer_name: meta-llama/Meta-Llama-3-8B
+use_slow_tokenizer: true
+dataset_mixer:
+  allenai/tulu-v2-sft-mixture: 1.0
+  HuggingFaceH4/no_robots: 1.0
+  meta-math/MetaMathQA: 0.25
+  m-a-p/CodeFeedback-Filtered-Instruction: 1.0
+  ai2-adapt-dev/daring-anteater-specialized: 1.0
```
hamishivi: did we want to commit this config? Don't have a strong opibion either way.

scripts/submit_finetune_job.py
```diff
@@ -132,9 +143,13 @@ def parse_args(args):
 
     new_arguments = override_and_reconstruct_command(d1['tasks'][0]['arguments'][0], train_config, unknown_args)
     
-    # place --num_processes with args.num_gpus
+    # place --num_processes with args.num_gpus * args.num_nodes
     # will be --num_processes {N} before
-    new_arguments = re.sub(r'--num_processes \d+', f'--num_processes {args.num_gpus}', new_arguments)
+    new_arguments = re.sub(r'--num_processes \d+', f'--num_processes {args.num_gpus * args.num_nodes}', new_arguments)
```
hamishivi, line 148: oh, for multinode num processes is the total processes across all machines? Thats a bit odd to me.

scripts/submit_finetune_job.py
```diff
@@ -59,7 +60,11 @@ def main():
         default_yaml = f.read()
     d1 = yaml.load(default_yaml, Loader=yaml.FullLoader)
 
-    d1["tasks"][0]["image"]["beaker"] = args.image
+    if args.num_nodes > 1:
+        args.num_gpus = 8 # assume we use entire nodes if doing multinode training
```
hamishivi: Can we add some sort of assert, logging, etc. to make it clear if someone tries to submit multiple nodes with < 8 gpus this is getting overriden. Probably an assert is best?

open_instruct/utils.py
```diff
@@ -152,9 +152,10 @@ def get_datasets(
     Loads and mixes datasets according to proportions specified in `dataset_mixer`.
 
     Args:
-        dataset_mixer (`dict`):
+        dataset_mixer (`list` or `dict`):
             Dictionary containing the dataset names and their training proportions.
             By default, all test proportions are 1.
+            If a list is passed in, it will be converted to a dictionary.
```
hamishivi: Can you give a little example of list format? I'm guess its `key1 value1 key2 value2...`?

hamishivi (approved):
Generally looks good. It would be good to have documentation. I'm not 100% on the dataset mixer logic, but sort of happy to just merge if you've tested it to some reasonable degree.
Left some comments that are more clarification questions than anything else.

scripts/submit_finetune_job.py
jacob-morrison: I think so -- that's how I've been doing it at least and it's worked; this also follows the old og default_finetune_multinode.yaml

configs/train_configs/sft/tulu3_8b_preview_mix.yaml
jacob-morrison: Yes and no -- this will be our current canonical preview mix config, but I'll update this before submitting the PR to follow what we talked about today in the tulu standup

open_instruct/finetune.py
jacob-morrison: I realized I can just check if either is not none (b/c we enforce that only one can be not none at a time), so now this should be able to handle both cases

scripts/submit_finetune_job.py
jacob-morrison: Good call, done (with an assert). I also changed the default value of `num_gpus` to 8, so the flag can be left out if doing multinode training

open_instruct/utils.py
jacob-morrison: done!

vwxyzjn (commented):
Overall looks good! Not necessarily for this PR but would be useful to have some docs on how to run these mixed data experiments

open_instruct/finetune.py
```diff
@@ -237,7 +237,7 @@ def main(args: FlatArguments):
             splits=["train"],
             save_data_dir=args.dataset_mix_dir,
             columns_to_keep=["messages"],
-            need_columns=["messages"],
+            # need_columns=["messages"],
```
natolambert: Why do this? Should we just remove the whole arg (including from `get_datasets`)?

open_instruct/utils.py
```diff
@@ -56,19 +56,25 @@ def is_openai_format(messages: Any) -> bool:
 
 
 # functions for handling different formats of messages
-def instruction_output_to_messages(example):
+def convert_alpaca_gpt4_to_messages(example):
```
natolambert, line 59: Why not leaving these as general, as we can use them with multiple datasets?

open_instruct/finetune.py
jacob-morrison: I'm probably going to -- I'm just testing this for now and I had to push the change to rebuild the docker image. tbd

open_instruct/utils.py
jacob-morrison: This is to mirror how they're currently being used -- none of the functions I renamed are shared right now. On the other hand I left `conversations_to_messages` as is. I also had to edit a few of the functions to add dataset-specific prompts. These will continue to evolve as we make changes to the mix, so the design will be in flux for a bit.

natolambert: I know is just sad to go from general design to specific, just for naming, and expect to need to change back to general 🙃

jacob-morrison: Candidly, I'm not expecting to change any of these back to general 🙂 Each dataset tends to have its own logic if it's not already in our format, so I expect we'll have a handful of dataset-specific functions and that's it.

Dataset-specific is also cleaner in general, because adding a new dataset requires no changes to preexisting logic. You either re-use a function, or just write a short new one.

natolambert: Many datasets have the same logic....

vwxyzjn: If we want to avoid this logic, (not in this PR) we can maintain our own dataset. We have like maybe 4 `create_dataset.py` for different styled datasets. Good for reproduction purposes, and we can distribute these logic in many files. 

E.g., https://github.com/huggingface/trl/tree/main/examples/datasets

Easy for people to add new datasets, too.

jacob-morrison: out of all the datasets we're looking at that aren't already in a messages format, only 3 have the same logic, and they're all in the `conversations` format. all of the others are unique. like I said earlier, if a new dataset shares the same logic as a preexisting dataset, they can share a function. otherwise, each having their own is less brittle.

jacob-morrison: > If we want to avoid this logic, (not in this PR) we can maintain our own dataset. We have like maybe 4 `create_dataset.py` for different styled datasets. Good for reproduction purposes, and we can distribute these logic in many files.
> 
> E.g., https://github.com/huggingface/trl/tree/main/examples/datasets
> 
> Easy for people to add new datasets, too.

Agreed, this is also a good solution (for a separate PR).

PR #223, opened 2024-08-07
Use the ai2-llm wandb team if possible

vwxyzjn:
This PR tries to use the ai2-llm wandb team if possible.

### DIFF
--- a/open_instruct/finetune.py
+++ b/open_instruct/finetune.py
@@ -548,7 +548,7 @@ def main(args: FlatArguments):
         if args.wandb_entity is None:
             args.wandb_entity = maybe_use_ai2_wandb_entity()
         accelerator.init_trackers(
-            "open_instruct_sft", experiment_config, init_kwargs={"wandb": {"entity": args.wandb_entity}}
+            "open_instruct_internal", experiment_config, init_kwargs={"wandb": {"entity": args.wandb_entity}}
         )
 
     # Train!
--- a/open_instruct/finetune.py
+++ b/open_instruct/finetune.py
@@ -50,6 +50,7 @@ from open_instruct.utils import (
     ArgumentParserPlus,
     FlatArguments,
     get_datasets,
+    get_wandb_tags,
     maybe_use_ai2_wandb_entity,
 )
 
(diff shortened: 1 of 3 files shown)


open_instruct/finetune.py
```diff
@@ -545,8 +550,12 @@ def main(args: FlatArguments):
         experiment_config = vars(args)
         # TensorBoard cannot log Enums, need the raw value
         experiment_config["lr_scheduler_type"] = experiment_config["lr_scheduler_type"]
+        if args.wandb_entity is None:
+            args.wandb_entity = maybe_use_ai2_wandb_entity()
         accelerator.init_trackers(
-            "open_instruct_sft", experiment_config, init_kwargs={"wandb": {"entity": args.wandb_entity}}
+            "open_instruct_internal",
+            experiment_config,
+            init_kwargs={"wandb": {"entity": args.wandb_entity, "tags": [os.path.basename(__file__)[: -len(".py")]]}},
```
natolambert: Should we hardcode the tags instead of doing the python pathing that may become broken in future file moving?

open_instruct/finetune.py
```diff
@@ -46,7 +46,12 @@
     get_scheduler,
 )
 
-from open_instruct.utils import ArgumentParserPlus, FlatArguments, get_datasets
+from open_instruct.utils import (
+    ArgumentParserPlus,
+    FlatArguments,
+    get_datasets,
+    maybe_use_ai2_wandb_entity,
```
natolambert, line 54: lol at naming

open_instruct/utils.py
```diff
@@ -639,7 +653,7 @@ def parse_yaml_and_args(self, yaml_arg: str, other_args: Optional[List[str]] = N
 
         return outputs
 
-    def parse(self) -> DataClassType | Tuple[DataClassType]:
+    def parse(self) -> Union[DataClassType, Tuple[DataClassType]]:
```
natolambert, line 721: Why this needed?

natolambert (approved):
Minor q's, but LGTM.

open_instruct/finetune.py
vwxyzjn: I was hardcoding initially, but decided to do it this way to reduce lines of code differences. Then, we also don't have to do this for different files (e.g., `ppo.py`, `online_dpo.py`)

image

open_instruct/utils.py
vwxyzjn: Friendlier for python 3.9. This syntax is only supported at python 3.10+

vwxyzjn:
@natolambert  I added some more quality of life tags, like the git related tags and pr-number if possible. So people have more controls over what to group and filter.

image

PR #224, opened 2024-08-07
[WIP] Rejection Sampling with multiple rewards 

nouhadziri:
Added multiple RMs to score each completion for each prompt and then take the majority vote for both chosen and rejected sample.


nouhadziri:
The code works fine, will add comments shortly.

PR #226, opened 2024-08-08
Dockerfile size fixes

hamishivi:
Just making our dockers build again. Removed cruft from our image and deleted some extra stuff in the runner. Stuff builds now.

### DIFF
--- a/.github/workflows/push-image-olmo.yml
+++ b/.github/workflows/push-image-olmo.yml
@@ -22,14 +22,14 @@ on:
       - 'Dockerfile.olmo'
       # Note, add .olmo dockerfile + requirements if adding auto build to those
     branches: [main,dockerfile-size-fixes]
-  pull_request: # note, comment this out for running on every push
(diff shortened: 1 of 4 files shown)


.github/workflows/push-image-olmo.yml
```diff
@@ -21,7 +21,7 @@ on:
       - 'requirements-olmo.txt'
       - 'Dockerfile.olmo'
       # Note, add .olmo dockerfile + requirements if adding auto build to those
-    branches: [main]
+    branches: [main,dockerfile-size-fixes]
```
natolambert: Leave this in?

natolambert (approved):
LGTM. One question on if we leave action in on the branch this is for OLMo.

.github/workflows/push-image-olmo.yml
hamishivi: whoops, should rm.

PR #229, opened 2024-08-08
Rejection Sampling with multiple RMs

nouhadziri:
The code now supports multiple Reward Models (RMs) to score each completion for every prompt. It then selects the final chosen and rejected samples based on a majority vote, ensuring more reliable and consensus-driven outcomes.

A TODO is to support LMs as judge instead of RM. I see @vwxyzjn has added `gpt_as_judge` which can be nicely integrated in this code.


nouhadziri:
I added also lots of comments to help understand faster the code, especially added dimensions for different tensors

natolambert (approved):
Mostly leave this up to @vwxyzjn. This looks similar to when I tried this for RewardBench. Eventually, a centralized inference tool for reward models would help alot and make this easier.

vwxyzjn (commented):
Hi @nouhadziri very nice change! I also really like the comment you have added. Have you tested it out with more than two 8B models? I can imagine the script going OOM if we are running, say five 5B models. 

One way to address it might be to call the original rejection sampling script five times, outputting different file names like

```
scores_model1_shard1.jsonl
scores_model1_shard2.jsonl
scores_model2_shard1.jsonl
scores_model2_shard2.jsonl
scores_model3_shard1.json1
scores_model3_shard2.json1
```

Then have a `majority_vote` script that combines these results.

nouhadziri:
Good point @vwxyzjn!! I haven't tested it for more than 3 models so far and so I presume it will go OOM eventually with higher number. 

But, why would we run into OOM with more than 8B, we are not running all the models scoring at the same time, instead we're creating a multiprocessing pool for each model and distributes the shards over the available GPUs sequentially for each model. So if we run into OOM it's because one model is too big to fit within the allocated GPUs? no? 🤔

If we call the original rejection sampling script, we might see a hit on runtime speed, the overhead won't be significant though ..

open_instruct/rejection_sampling_multiple_rewards.py
```diff
@@ -0,0 +1,321 @@
...
+from huggingface_hub import HfApi
+from collections import Counter
+from typing import Union
+
+api = HfApi()
+
+
+@dataclass
+class Args:
+    model_names_or_paths: tuple[str] = ("allenai/llama-3-tulu-2-8b-uf-mean-rm", "cleanrl/EleutherAI_pythia-1b-deduped__reward__tldr", "weqweasdas/RM-Mistral-7B")
```
vwxyzjn:
```suggestion
    model_names_or_paths: List[str] = field(default_factory=lambda: ["allenai/llama-3-tulu-2-8b-uf-mean-rm", "cleanrl/EleutherAI_pythia-1b-deduped__reward__tldr", "weqweasdas/RM-Mistral-7B"])
```

This enables us to tune `model_names_or_paths` in the command line :)

open_instruct/rejection_sampling_multiple_rewards.py
```diff
@@ -0,0 +1,321 @@
...
+    num_samples = offsets_per_model[next(iter(offsets_per_model))].size(0)
+    # Initialize tensor to store the majority votes
+    majority_votes = torch.zeros(num_samples, dtype=torch.long)
+
+    # Tally the votes and determine the majority vote for each sample
+    for i in range(num_samples):
+        # Collect votes from all models for the current sample
+        votes = [offsets_per_model[model][i].item() for model in offsets_per_model]
+        # Determine the most common vote
+        majority_vote = Counter(votes).most_common(1)[0][0]
```
vwxyzjn:
```suggestion
        counter = Counter(votes)
        # Try to get ther majority vote, but if all models disagree, we randomly choose one
        if len(offsets_per_model) != len(counter):
            majority_vote = counter.most_common(1)[0][0]
        else:
            majority_vote = votes[np.random.randint(len(votes))]
```

One special case: what if the models disagree all the time? In that case we should randomly choose one. The current implementation I think might have always choose the the first RM.

vwxyzjn (commented):
> But, why would we run into OOM with more than 8B

Ah I was worried torch might not be able to relase the allocated memory. However I just tested your script and found it works fine :) 


Given that my worries are resolved! I added some additional nits, but overall the PR looks really great. Feel free to 

* make a change to `open_instruct/rejection_sampling.py` directly and 
* update the documentation in `docs/algorithm`

open_instruct/rejection_sampling_multiple_rewards.py
nouhadziri: Wonderful, thanks:)

nouhadziri: Fair enough, I just tested it with 4 RMs and it works fine so far :) 
Thanks, will merge and add documentation

open_instruct/rejection_sampling_multiple_rewards.py
nouhadziri: Good catch, will fix this after the meeting.

PR #231, opened 2024-08-09
Remove HF tag

hamishivi:
Really small change - remove the `hf-` tag in front of the model name in the hf script for uploading to the leaderboard. This is useful to have more flexible naming in the leaderboard.


ValentinaPy (approved):
lgtm

PR #234, opened 2024-08-09
DPO Submission Script

hamishivi:
Adding a version of `submit_finetune` for DPO. Since we are more likely to be doing DPO from internal checkpoints, I added some stuff for mounting beaker datasets.

Appears to work: https://beaker.org/ex/01J4WJK6MHPNARSN14QBJV17HY/tasks/01J4WJK6MQ0DQGH0ASQF5H0TJG/job/01J4WJKB9X1Q9MF8KK012CRWBD

I haven't tested multinode stuff, though.

### DIFF
--- a/configs/beaker_configs/default_dpo.yaml
+++ b/configs/beaker_configs/default_dpo.yaml
@@ -1,0 +1,69 @@
+version: v2
+description: open-instruct-finetune
+budget: ai2/oe-adapt
+tasks:
+  - name: open-instruct-finetune
+    image:
+      beaker: nathanl/open_instruct_auto
+    command: [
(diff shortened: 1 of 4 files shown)


jacob-morrison:
LGTM! I used the DPO script in a multinode setting to run DPO on OLMoE, and it seemed to work fine? We probably just need to create a multinode config and everything else should Just Work:tm:. It seems very quick to run though, so I'll approve this and we can add it later if you want

jacob-morrison (approved):

hamishivi:
Cool! Yeah, lets handle multinode later (when we need to).

natolambert:
Nice!

PR #237, opened 2024-08-11
Add SimPO+WPO+DPO norm loss

hamishivi:
[SimPO](https://github.com/princeton-nlp/SimPO) has had some promising performance on AlpacaEval. This PR adds SimPO as an option into the DPO script (since SimPO is very similar).

Additionally adds label smoothing into DPO, following [Eric Mitchell's note](https://ericmitchell.ai/cdpo.pdf).

Also just added WPO implementation on yizhong's suggestion, although I need to test.
Also adds 'dpo_norm', which is just length normalized DPO. As noted in [the SimPO repo](https://github.com/princeton-nlp/SimPO/issues/20), this can be quite effective with setups like ultrafeedback, but requires a much larger beta value (>= 5, from my testing).

From testing, the training seems to work (models get better Aeval, don't diverge, etc), but I'm still playing around with hyperparameters etc. So I think its generally okay to merge.

### DIFF
--- a/configs/train_configs/dpo/tulu_3_preview_test_simpo.yaml
+++ b/configs/train_configs/dpo/tulu_3_preview_test_simpo.yaml
@@ -1,0 +1,25 @@
+model_name_or_path: /model
+model_revision: main
+use_flash_attn: true
+gradient_checkpointing: true
+tokenizer_name: /model
+use_slow_tokenizer: true
+dataset_name: princeton-nlp/llama3-ultrafeedback-armorm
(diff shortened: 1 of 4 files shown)


yizhongw:
Great, have you experimented with it?

yizhongw:
Also, WPO also seems promising and easy to implement (they claimed better performance than SimPO): https://arxiv.org/abs/2406.11827.

natolambert (approved):
LGTM. It is nice to be able to play with these.

PR #238, opened 2024-08-11
Add safety evaluation as a submodule

nouhadziri:
Added [safety-eval](https://github.com/allenai/safety-eval) repo as a submodule. Running the evaluation is straightforward. I've included a README under `open-instruct/docs/safety-eval` ([safety.md](https://github.com/allenai/open-instruct/blob/safety-eval/docs/safety-eval/safety.md)) that details how to run the scripts both in an interactive session and on Beaker. The submodule supports evaluation on 7 different benchmarks.

Currently, the submodule points to a fork of [safety-eval](https://github.com/allenai/safety-eval). If you’d like to make changes to the safety evaluation scripts, let me know, and I can add you to the fork.

Why a Submodule?
- Generally, I prefer not to use submodules, but this is a temporary solution to quickly integrate safety evaluation.
- Rewriting the code in the open-instruct evaluation logic would be time-consuming and may not be worthwhile, especially since Oyvind mentioned (on friday) that safety evaluation will be integrated into oe-eval within the next two weeks.

Are results added to the leaderboard?
- I'm missing info on how to add them properly, currently the code is uploading scores into a table in HF repo but doesn't look nice, you can look [here](https://github.com/nouhadziri/safety-eval-fork/blob/0291617446f4edf32297c313389dd0b0ae1e4cff/evaluation/run_all_generation_benchmarks.py#L149-L168) to make changes:

### DIFF
--- a/docs/safety-eval/safety.md
+++ b/docs/safety-eval/safety.md
@@ -48,7 +48,7 @@ In the examples below, text within {} tags should be replaced with your own valu
 As a convenience, you can use the `evaluation/gantry_run.sh` script which includes some necessary arguments. You can use it the same way as `gantry run`, but excluding these boilerplate arguments (take a look at the script to see what it includes). Example usage:
 
 ```bash
-./evaluation/gantry_run.sh --workspace {workspace} --cluster {cluster} --gpus {n_gpus} \
+PYTHONPATH=safety-eval ./safety-eval/evaluation/gantry_run.sh --workspace {workspace} --cluster {cluster} --gpus {n_gpus} \
     --priority {priority} -- python evaluation/run_all_generation_benchmarks.py \
     --model_name_or_path allenai/tulu-2-dpo-7b \
     --model_input_template_path_or_name tulu2 \
--- a/docs/safety-eval/safety.md
+++ b/docs/safety-eval/safety.md
@@ -60,7 +60,7 @@ Here is an example using the full `gantry run` command. Use the beaker image `se
 
 **Important**: Please include all the beaker arguments exactly as in the examples unless intentionally modifying some configuration. Many of them are necessary to avoid job failures, such as `--beaker-image`, `--venv`, and `--env-secret`. Note that `openai_api_key` and `hf_token` are Beaker workspace secret names, so should *not* be replaced with actual values (see One-Time Setup).
 
-Note that the `--` divides the gantry command from the evaluation command - you can edit the second part to run whatever eval suite you want from the `eval.py` script, or any other script in the OE-Safety repo. Any additional Beaker arguments such as a dataset mount to use a model from a Beaker dataset or adding a priority tag can be added before the `--`.
+Note that the `--` divides the gantry command from the evaluation command - you can edit the second part to run whatever eval suite you want from the `eval.py` script. Any additional Beaker arguments such as a dataset mount to use a model from a Beaker dataset or adding a priority tag can be added before the `--`.
 
 You can also run all generator evaluations parallelized across the GPUs allocated to your batch job, like so:
 ```bash
--- a/docs/safety-eval/safety.md
+++ b/docs/safety-eval/safety.md
@@ -1,0 +1,114 @@
+
+## Running on an interactive session
+
+Run the following command from the main project dir `open-instruct`.
+### Safety benchmarks
+
+For all benchmarks requiring safety classification unless noted otherwise, as a default, we use the [WildGuard](https://github.com/allenai/wildguard) classifier to evaluate the safety of model outputs.
+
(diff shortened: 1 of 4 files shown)


requirements-olmo.txt
```diff
@@ -36,4 +36,6 @@ immutabledict
 black
 flake8
 isort
-autoflake
\ No newline at end of file
+autoflake
+tenacity==8.4.1
```
hamishivi, line 40: could we also add to `requirements.txt` if possible?

docs/safety-eval/safety.md
```diff
@@ -0,0 +1,101 @@
...
+## Running on a Beaker Batch Job
+
+The recommended way to run evaluation on Beaker is through [beaker-gantry](https://github.com/allenai/beaker-gantry). The basic format of a gantry command is `gantry run {beaker arguments} -- python {script path from repo root} {python script arguments}`. 
+
+**Important**: Before you run any command with gantry, make sure you *commit and push*, since gantry will attempt to clone the repo with your local latest commit hash.
+
+See the "One-Time Setup" section below before running commands. To test your setup, run the following command -- if this job succeeds, then you're ready to run evaluations with gantry.
+
+```bash
+gantry run --workspace {workspace} --budget ai2/oe-adapt --beaker-image kavelr/oe-safety --venv base --cluster ai2/mosaic-cirrascale --env-secret OPENAI_API_KEY=openai_api_key --env-secret HF_TOKEN=hf_token -- python -c 'print("Hello world")'
```
hamishivi, line 41: is there a way to run without gantry? My guess is with a pre-built image, then we might not even need to clone the repo in. We just setup a beaker config pointing at the right image and gives the right commands (which we have example code for in `submit_eval_jobs`...

requirements-olmo.txt
nouhadziri: Done!

docs/safety-eval/safety.md
nouhadziri: It should definitely work with `submit_eval_jobs` with the existing image. I just find granty easier to use as it requires minimal effort to run.

hamishivi: Cool! Let me test that then... gantry is fine, but if we have a prebuilt image with the safety stuff then we can edit the dockerfile to clone the repo there, rather than have it as a submodule here.

nouhadziri: The image is `seungjuh/oe-safety-support-olmo17`

docs/safety-eval/safety.md
```diff
@@ -0,0 +1,101 @@
...
+- [WildGuardTest](https://arxiv.org/abs/2406.18495)
+- [Harmbench](https://arxiv.org/abs/2402.04249)
+- [ToxiGen](https://arxiv.org/abs/2203.09509)
+- [XSTest](https://arxiv.org/abs/2308.01263)
+- [JailbreakTrigger (in TrustLLM)](https://arxiv.org/abs/2401.05561)
+- [Do-anything-now](https://arxiv.org/abs/2308.03825)
+- [WildJailbreak](https://arxiv.org/abs/2406.18510) (both harmful and benign contrast sets)
+
+```commandline
+PYTHONPATH=safety-eval python safety-eval/evaluation/run_all_generation_benchmarks.py    \
```
hamishivi, line 18: managed to run locally! But is there a way to use the chat template provided in the HF tokenizer? Currently thats what the rest of our eval suite does, and so having to specify a different thing here is a bit annoying.

hamishivi: If this is easy, I can add a script to run the safety eval automatically with the rest of the eval suite. If not, I say we just leave this PR here for now in case people want to run safety evals, and otherwise wait for oe-eval.

hamishivi: also have some small fixes to the safety fork as well if we decide to merge, mainly just me wanting to upload one overall safety score rather than 6 new numbers.

nouhadziri: right now since most models are a finetuned version of llama3, it can be simple to add the template [here](https://github.com/allenai/OE-Safety/blob/main/src/templates/single_turn.py), but if we want to use HF directly, that would require a few changes here and there. 
Averaging scores sounds good, we just need to ensure that we only average scores that measure the same thing.

hamishivi: Oh looks like it already supports using the hf chat template! https://github.com/allenai/OE-Safety/blob/main/src/templates/single_turn.py#L260

hamishivi:
I'm going to close in favour of #286, which seems a bit cleaner from my perspective. This will do until we (hopefully) get the evals integrated into oe-eval.

PR #239, opened 2024-08-11
Create config for my dpo experiment with custom data

ValentinaPy:


configs/train_configs/dpo/valpy_dpo.yaml
```diff
@@ -0,0 +1,22 @@
+model_name_or_path: /model
+model_revision: main
+use_flash_attn: true
+gradient_checkpointing: true
+tokenizer_name: /model
+use_slow_tokenizer: true
+dataset_name: valpy/uf_anteater_wildchat
+max_seq_length: 4096
```
hamishivi, line 8: You might want to set this to 2k, otherwise you might get oom.

hamishivi (approved):

PR #241, opened 2024-08-12
Utility improvements

vwxyzjn:
This PR does a few things:

1. improve `save_with_accelerate` to use `api.upload_folder` instead of `model.push_to_hub`. Sometimes `model.push_to_hub` does not successfully upload the model shards like blow, so this PR makes sure the folder is uploaded.

image

2. when preparing for deepspeed, this PR accepts the mixed precision mode directly.


natolambert (approved):
Seems fine to me as long as tested

PR #243, opened 2024-08-12
Does prompt make sense?

vwxyzjn:
This PR includes a simple script to judge the quality of the SFT prompts in the dataset.

image

### DIFF
--- a/scripts/does_prompt_make_sense.py
+++ b/scripts/does_prompt_make_sense.py
@@ -1,0 +1,127 @@
+import asyncio
+import random
+import time
+from dataclasses import dataclass
+from typing import Optional
+
+import pandas as pd
+from datasets import load_dataset
+from openai import AsyncOpenAI
+from tqdm.asyncio import tqdm_asyncio
+from transformers import AutoTokenizer, HfArgumentParser
+
+
+@dataclass
+class LLMJudgeConfig:
+    n: int = 64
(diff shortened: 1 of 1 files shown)


hamishivi:
As a very basic first pass this makes sense, but I wonder if:
(a) we can be more specific / ask for fine-grained scores (e.g., coherence, length, etc.). I feel like recent llm-as-a-judge work is trending to more fine-grained scores (e.g. ultrafeedback finegrained vs overall).
(b) If we can make this more quantitative somehow. Can we have some analysis of false positives or similar? Or maybe some small val set that we can reason about and quality-check with?

scripts/does_prompt_make_sense.py
```diff
@@ -0,0 +1,127 @@
...
+from datasets import load_dataset
+from openai import AsyncOpenAI
+from tqdm.asyncio import tqdm_asyncio
+from transformers import AutoTokenizer, HfArgumentParser
+
+
+@dataclass
+class LLMJudgeConfig:
+    n: int = 64
+    model: str = "gpt-3.5-turbo-0125"
```
hamishivi, line 19: should we use 4o?

hamishivi (approved):
Maybe we could make some sort of experimental folder for a rough script like this? Unsure, this sort of thing feels way more rough that e.g. the submit_eval/finetune scripts.

PR #252, opened 2024-08-13
Generate and score responses (LLM-as-judge) with API-based models

nouhadziri:
This PR does two things:
- API-mased model response generation: In addition to generating completions from HF and local models, the code now supports generating completions using API-based models (e.g., GPT-4). In a future PR I will address sampling responses from an ensemble of HF and API models.

- Response scoring with API models (aka LLM-as-judge): We now use API-based models to score responses based on the given prompt and scoring criteria. Note that this approach is different from prompting the model to rank multiple responses simultaneously. We need to address this in a separate PR and compare the results.

I have added a file `prompt_templates.py` which contains two types of prompts: i) generation and ii) judgment. This file also supports different skills, which will allow us to experiment with prompts for each skill added to Tulu.


vwxyzjn (commented):
Very cool work @nouhadziri. Some high-level comments: 

* could you put `.idea` stuff in `.gitignore`?
* could you add docs in the `docs/algorithm/rejection_sampling.md` to show some hello world commands people can quickly run and get familiar with it?

nouhadziri:
My bad, those git stuff got pushed accidentally. I fixed it now and added docs.

natolambert:
Am happy with it if Costa is, otherwise can look more tomorrow.

docs/algorithms/rejection_sampling.md
```diff
@@ -7,6 +7,7 @@ different number of completions per prompt.
 
 # Debug run (use an interactive session)
 
+This code supports HF models, local models and also API-based models (e.g., `gpt-4`). For generating completions, the code now accepts one model at a time, but we're working on adding an ensemble of models. Stay tuned. 
 ```bash
 ## tulu v3 recipe
 # 1. first sample a bunch of completions given prompts
```
vwxyzjn, line 13: This needs to be updated. Now it's `python open_instruct/rejection_sampling/rejection_sampling.py`. There are also other places that need to be changed.

docs/algorithms/rejection_sampling.md
```diff
@@ -16,7 +17,10 @@ python open_instruct/generation.py \
     --n 3 \
     --save_filename output/completions.jsonl \
     --sanity_check \
-    
+   ```
+### Scoring completions
+You can use either a single RM to score responses or a list of RMs. In the latter case, we will take the majority vote to compute the final score. The RMs can be models explicitly trained as RMs, HF LMs, or API-based models.
+```
```
vwxyzjn, line 23: Could you add more docs here on the usage? Perhaps something like 


### Scoring completions

You can use either a single RM to score responses or a list of RMs. In the latter case, we will take the majority vote to compute the final score. The RMs can be models explicitly trained as RMs, HF LMs, or API-based models.
```
# 2.1 tokenize them and run a reward model to filter them
python open_instruct/rejection_sampling.py \
    --input_filename output/completions.jsonl \
    --model_names_or_paths allenai/llama-3-tulu-2-8b-uf-mean-rm \
    --save_filename output/rejection_sampled_completions.jsonl \
    --n 3 \
    --push_to_hub \
    --num_gpus 1 \
# 2.2 tokenize them and run multiple reward models
... command to be filled
# 2.3 use API-based models to do rejection sampling
... command to be filled
```

open_instruct/rejection_sampling/generation.py
```diff
@@ -33,12 +36,13 @@
 class Args:
     model_name_or_path: str = "cleanrl/EleutherAI_pythia-1b-deduped__sft__tldr"
     save_filename: str = "completions.jsonl"
+    mode: str = "generation"
+    skill: str = "chat"
```
vwxyzjn: Are these used?

open_instruct/rejection_sampling/generation.py
```diff
@@ -97,6 +89,56 @@ def main(args: Args, dataset_args: DatasetArgs, gen_args: GenerationArgs):
...
+        # Generate using vLLM.
+        outputs = generate_with_vllm(args.model_name_or_path, prompt_token_ids, gen_args)
+
+    else:
+        tokenizer = AutoTokenizer.from_pretrained("allenai/llama-3-tulu-2-8b")
+        ds = ds.map(
+            lambda x: {"prompt": tokenizer.apply_chat_template(x["messages"][:-1], tokenize=False)},
+            num_proc=multiprocessing.cpu_count(),
+        )
+        messages = ds[dataset_args.dataset_train_split]["prompt"]
```
vwxyzjn: We should perhaps use openai's native conversation format (i.e., you can send `[item[:-1] for item in ds[dataset_args.dataset_train_split]["messages"]]`?

```python
from openai import OpenAI
client = OpenAI(
    # Defaults to os.environ.get("OPENAI_API_KEY")
)

chat_completion = client.chat.completions.create(
    model="gpt-4o-mini",
    messages=[
        {"role": "user", "content": "Hi! Do you like reading? I really like reading Game of Thrones."},
        {"role": "assistant", "content": "Hi, I do like reading."},
        {"role": "user", "content": "What did I say I like reading?"},
    ]
)
>>> chat_completion
ChatCompletion(id='chatcmpl-9vsvucRqJKLPaH1UpiqAXIkifroVs', choices=[Choice(finish_reason='stop', index=0, logprobs=None, message=ChatCompletionMessage(content='You mentioned that you really like reading "Game of Thrones." It\'s a fascinating series with rich characters and intricate plots! What do you enjoy most about it?', refusal=None, role='assistant', function_call=None, tool_calls=None))], created=1723582886, model='gpt-4o-mini-2024-07-18', object='chat.completion', service_tier=None, system_fingerprint='fp_48196bc67a', usage=CompletionUsage(completion_tokens=31, prompt_tokens=45, total_tokens=76)
```

open_instruct/rejection_sampling/generation.py
```diff
@@ -107,19 +149,13 @@ def main(args: Args, dataset_args: DatasetArgs, gen_args: GenerationArgs):
     # ...
     table = defaultdict(list)
     for output, messages in zip(outputs, ds[dataset_args.dataset_train_split]["messages"]):
-        # if the model completions are exactly the same across all completions, we can skip this
-        if len(set([item.text for item in output.outputs])) == 1:
-            continue
```
vwxyzjn, line 112: Why did we remove this? In this case all the completions are the same.

open_instruct/rejection_sampling/generation.py
```diff
@@ -129,6 +165,7 @@ def main(args: Args, dataset_args: DatasetArgs, gen_args: GenerationArgs):
                     "messages": table["messages"][i],
                     "model_completion": table["model_completion"][i],
                     "reference_completion": table["reference_completion"][i],
+                    "model": args.model_name_or_path,
```
vwxyzjn, line 187: nice add!

open_instruct/rejection_sampling/rejection_sampling.py
```diff
@@ -50,6 +51,8 @@ class Args:
     hf_entity: Optional[str] = None
     hf_repo_id: str = "rejection_sampling"
     add_timestamp: bool = True
+    mode: str = "judgement"
+    skill: str = "chat"
```
vwxyzjn, line 56: Still not used

open_instruct/rejection_sampling/rejection_sampling.py
```diff
@@ -195,18 +245,30 @@ def main(args: Args):
...
+
+        results = []
+        if not use_openai:
+            with mp.Pool(args.num_gpus) as pool:
+                for i in range(args.num_gpus):
+                    results.append(pool.apply_async(process_shard, (i, model_name_or_path, args, shards[i])))
+        else:
             for i in range(args.num_gpus):
-                results.append(pool.apply_async(process_shard, (i, model_name_or_path, args, shards[i])))
+                results.append(process_shard_api(model_name_or_path, args, shards[i]))
```
vwxyzjn, line 261: We should make a note that when using LLM as a judge, `num_gpus` doesn't actually use that many GPUs, instead it's just the number of shards.

vwxyzjn (changes requested):
Very nice work! I left some more comments :)

docs/algorithms/rejection_sampling.md
nouhadziri: Good catch, I fixed it. Also fixed the path in `rejection_sampling_tulu.bash` and `rejection_sampling_hello_world.bash`

docs/algorithms/rejection_sampling.md
nouhadziri: Done

open_instruct/rejection_sampling/generation.py
nouhadziri: Yes, they're needed in `api_generate.py`. I removed the redundancy of those args in `api_generate.py` though.

open_instruct/rejection_sampling/generation.py
nouhadziri: Since I’m using a template to guide the model on how to respond or provide ratings, I prefer to format it myself rather than relying on the native chat completion. So, the entire conversation will be provided as a single "multi-turn prompt" to the model. Although I could create different templates for "user" and "assistant" I find the outcome to be the same. So, I prefer using a single template to save time and maintain conciseness. But, we can discuss this more and happy to change it if you feel we should stick with native :) 

In `api_generate.py`, text will be my template with the chat formatted between user A and user B. 
``` 
response = await self.async_client.chat.completions.create(
                        model=self.config.model,
                        messages=[
                            {"role": "system", "content": "You are a helpful assistant."},
                            {"role": "user", "content": text},
                        ],
                    )

open_instruct/rejection_sampling/generation.py
nouhadziri: oh, I removed it because in my test case output["outputs"] was equal to 1 so len(set()) will be always 1 so I ended having an empty file.

open_instruct/rejection_sampling/rejection_sampling.py
nouhadziri: It's being used in `api_generate.py` for ranking, it reads those two through `args`

open_instruct/rejection_sampling/rejection_sampling.py
nouhadziri: Yes correct, we'll add a note. We can also create another var for # of shards.

open_instruct/rejection_sampling/generation.py
vwxyzjn: Makes sense. Let's stick with the current design for now and we can change it back if we found the native to work notably better.

vwxyzjn (approved):
LGTM. Thanks @nouhadziri!

PR #253, opened 2024-08-13
Add rejection sampling analysis

vwxyzjn:
```
🚀 Dataset loaded, starting to analyze...
chosen:    mean length = 163.50656107997827
rejected:  mean length = 150.02915225944915
reference: mean length = 811.4574606202721
key='allenai/llama-3-tulu-2-8b-uf-mean-rm'
chosen:    mean score = 26.952228895910338
           std score = 3.0159979740827905
rejected:  mean score = 23.47721361929176
           std score = 3.226889784341013
reference: mean score = 30.24208389637402
           std score = 4.438492691848654
reference is better than chosen
```

So I realized the issue is that we generate only 53 tokens by default. We should change it to a much more sensible number, prob 2048 @nouhadziri 

https://github.com/allenai/open-instruct/blob/151533074483ee3f5f07aa6a9fded9e9cdf99b0f/open_instruct/generation.py#L43

Let's merge this after your PR #252 and I can add documentation


nouhadziri:
This explains the performance!! I haven't had a chance to play with different generation parameters yet but increasing the length will definitely boost numbers.

nouhadziri:
LGTM

PR #254, opened 2024-08-13
Resume from preemption / checkpoints

### LINKED ISSUE #176: Add support for resuming from checkpoint
We should add support for resuming training partway through, e.g. if a job is preempted. I think most of the functionality is there, but we'll need to 1) save more often (every epoch will probably be wasteful depending on how often we get preempted) and 2) handle `--resume_from_checkpoint` well.

We'll need to properly handle `resume_from_checkpoint` at the start of training and when resuming partway through, which will most likely involve looking for a checkpoint manually first before letting the automagic huggingface code resume training. We'll also need to not overwrite the output folder.

hamishivi:
Adds support for resuming from checkpoints. There are some considerations here:
- If `overwrite_output_dir` is set to true, then we always resume from scratch (or a user-provided checkpoint).
- If `resume_from_checkpoint` is provided, but there are checkpoints in the output_dir, we assume that the output_dir is preferable to load from. This handles cases where e.g. you  start a job starting from a checkpoint, and the job runs and saves a new checkpoint, and then is preempted.
- `keep_last_n_checkpoints` to avoid saving a tonne of checkpoints.
- Some checkpoint marking scheme to avoid loading from garbled checkpoints, if we are preempted mid-save.

Example of succesful preemption+resumption for SFT: https://beaker.org/ex/01J56Y04TBTB39F2Z195DCPWDM/tasks/01J56Y04TS4E2SXQZYJMGCAMDP/job/01J56YYBPZJJQ1HDD9HDW990T5

And for DPO: https://beaker.org/ex/01J575GQD3DD6JEBZ1TN04G2X9/tasks/01J575GQD7ZN86F8KZTHBFNJME/job/01J577B0W3K1ER5C5HH0A9AS6R

Closes #176


open_instruct/dpo_tune.py
```diff
@@ -544,22 +546,13 @@ def load_model():
...
-            path = dirs[-1]  # Sorts folders by date modified, most recent checkpoint is the last
-            checkpoint_path = path
-            path = os.path.basename(checkpoint_path)
-
-        accelerator.print(f"Resumed from checkpoint: {checkpoint_path}")
-        accelerator.load_state(path)
+    last_checkpoint_path = get_last_checkpoint_path(args)
+    if last_checkpoint_path:
+        accelerator.print(f"Resumed from checkpoint: {last_checkpoint_path}")
+        accelerator.load_state(last_checkpoint_path)
```
vwxyzjn, line 560: Interesting! I did not know `accelerator.load_state` loads all the states (model, optimizer, scheduler)

vwxyzjn (approved):
Very nice work!

hamishivi:
Yeah, ideally the end user shouldn't have to do anything - the preempted job will be requeued and load back in, and resume from the last checkpoint. If the code itself errors, though, then the user will have to manually restart.

One thing is that uploading the optimizer states etc. can be a bit slow (at least, it was slow when I tested), so jobs won't resume instantly, but might take an hour or so for the upload checkpoints -> find gpus -> download and resume process to occur.

open_instruct/dpo_tune.py
hamishivi: Yeah. One thing im surprised by is that DPO seems to just work, I want to test it a little bit more, because the checkpoint seems to only be saving one model, but in DPO we have two models lying around (and I'm not telling accelerate anything special about which model to load/save).

vwxyzjn: Sometimes there may be too much magic going on. At some point we could prob verify with a picture like 

image

vwxyzjn:
> One thing is that uploading the optimizer states etc. can be a bit slow (at least, it was slow when I tested), so jobs won't resume instantly, but might take an hour or so for the upload checkpoints -> find gpus -> download and resume process to occur.

The upload thing is a bit undesirable. What about saving the checkpoint to NFS?

open_instruct/dpo_tune.py
hamishivi: Yeah, Wandb testing is on my agenda. From digging into the DPO stuff, it appears because the reference model loading is hacked in a bit, so the accelerate state doesnt pick it up (this is nice for us, tho, since the reference model doesn't change).

hamishivi:
You could do this via setting `args.output_dir` to a NFS path! This won't work for jupiter, though.

open_instruct/dpo_tune.py
natolambert: Resume from checkpoint for DPO may work best with Log Prob cacheing @hamishivi @vwxyzjn

natolambert (approved):
This looks good, is almost simpler than I expected.
For DPO, saving the last N checkpoints also can be nice because we can evaluate them to check for potential overfitting.

Discussing data on slack. Seems like additions may be needed there.

PR #256, opened 2024-08-14
Fix rejection sampling rm

vwxyzjn:
The current script hangs with RM at the `result.get()` line.

image

The issue is that the `item = result.get()` needs to live inside of the `mp.Pool` context. This PR fixes it by moving the `result.get()` in the appropriate context.

Also this PR used a more reasonable generation length.

### DIFF
--- a/open_instruct/rejection_sampling/generation.py
+++ b/open_instruct/rejection_sampling/generation.py
@@ -44,7 +44,7 @@ class Args:
 class GenerationArgs:
     n: int = 1
     temperature: float = 0.8
-    response_length: int = 53
+    response_length: int = 2048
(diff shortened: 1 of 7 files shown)


open_instruct/rejection_sampling/generation.py
```diff
@@ -44,7 +44,7 @@ class Args:
 class GenerationArgs:
     n: int = 1
     temperature: float = 0.8
-    response_length: int = 53
+    response_length: int = 2048
```
nouhadziri, line 47: thanks for fixing length ;)

nouhadziri:
LGTM. Thanks

nouhadziri (approved):
LGTM

PR #258, opened 2024-08-14
keep ids when saving data

### LINKED ISSUE #247: Save message IDs and source datasets during data mixing
Right now, the dataset mixer only stores the bare messages when mixing datasets. We should at minimum save a "source dataset" field for each dataset, and also unique IDs if possible

natolambert:
Closes #247

### DIFF
--- a/open_instruct/mix_data.py
+++ b/open_instruct/mix_data.py
@@ -35,6 +35,7 @@ def main():
         splits=["train"],
         save_data_dir=args.dataset_mix_dir,  # location where dataset is saved as json
         columns_to_keep=["messages"],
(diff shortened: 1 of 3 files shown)


vwxyzjn (approved):
Is this for better debugging purposes?

natolambert:
@vwxyzjn the ID (and maybe even subset later) Is nice for comparing prompts between datasets. They're in the original tulu mixes, so as we save mixes we would want these when we can. 

Mostly for saving / sharing purposes.

natolambert:
Just made sure the code still runs in a simple test, merging.

PR #259, opened 2024-08-14
Use updated Tulu config in oe-eval and optionally run multiturn fewshot evals

pdasigi:
Depends on https://github.com/allenai/oe-eval-internal/pull/198.

### DIFF
--- a/scripts/eval/oe-eval.sh
+++ b/scripts/eval/oe-eval.sh
@@ -42,6 +42,7 @@ while [[ "$#" -gt 0 ]]; do
         --model-name) MODEL_NAME="$2"; shift ;;
         --model-location) MODEL_LOCATION="$2"; shift ;;
         --hf-upload) HF_UPLOAD="true" ;;
+        --multiturn-fewshot) MULTITURN_FS="true" ;;
         *) echo "Unknown parameter passed: $1"; usage ;;
     esac
     shift
@@ -58,6 +59,7 @@ MODEL_NAME_SAFE=${MODEL_NAME//\//_}
 
 # Set defaults for optional arguments
 HF_UPLOAD="${HF_UPLOAD:-false}"
+MULTITURN_FS="${MULTITURN_FS:-false}"
 
 # Set HF_UPLOAD_ARG if HF_UPLOAD is true
 if [ "$HF_UPLOAD" == "true" ]; then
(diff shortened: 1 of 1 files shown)


scripts/eval/oe-eval.sh
```diff
@@ -18,21 +18,28 @@ set -ex
...
-# gsm8k::olmo1
-# drop::llama3
-# minerva_math::llama3
-# codex_humaneval
-# codex_humanevalplus
+# gsm8k::tulu
+# drop::tulu
+# minerva_math::tulu
+# codex_humaneval::tulu
+# codex_humanevalplus::tulu
```
hamishivi, line 28: Does this change lead to big changes in reported numbers? I worry a bit about the leaderboard and comparing chat to non-chat numbers.

jacob-morrison: I do think it's worth us switching to the chat format, but yeah I'm curious what the delta is

In either case when we switch, I think we should remove the current oe-evals from the leaderboard and create new columns for these. We can just rerun them for a subset of the models (or all if it's not too expensive) to ensure they're comparable

pdasigi: Yeah, it does lead to a big change in some evals. Here are the numbers (no chat -> chat) for Llama 3 8B Instruct:

MMLU: 0.6740 -> 0.4069
DROP: 0.5060 -> 0.1517
MATH: 0.1680 -> 0.0145

Let's discuss at the next meeting which evaluations to keep.

pdasigi:
Update:
MATH, GSM8K, MMLU, HumanEval, HumanEval+, IFEval, PopQA now use chat format, with the few-shot evals MATH, GSM8K, MMLU, and PopQA with demonstrations presented in multiple turns.
DROP is back to not using the chat format.

Additionally, HumanEval and HumanEval+ evaluate pass@10.

Main considerations:
- Use the chat format as much as possible, and do not use it only if the performormance drops significantly (DROP).
- Use the few-shot as multiturn setting that generally yields better results.
- Make the settings for tasks in the Tulu 1 and Tulu 2 suites be as similar as possible to those reported in the papers.

hamishivi:
Why does DROP drop so much? do we have any guesses?

pdasigi:
> Why does DROP drop so much? do we have any guesses?

It's an issue with the evaluation metric. Llama-3-8b-Instruct's responses in chat mode are quite verbose.

Example:
Q: How many field goals did Kris Brown kick?
Target: 3
Llama's response: According to the passage, Kris Brown kicked 3 field goals: a 53-yard, a 24-yard, and a 40-yard field goal.

hamishivi:
Closing in favour of #314

PR #267, opened 2024-08-16
Add my latest best config

hamishivi:
Adding my latest config with the best run from a DAA algo so far.


ValentinaPy (changes requested):
There's no: model_name_or_path: /model

hamishivi:
incorrect :P

PR #268, opened 2024-08-16
Fix resumption when dataset mixer is present

hamishivi:
When the dataset mixer is present, the presence of the output file mucked up my logic for cleaning older checkpoints. This should fix it.

### DIFF
--- a/open_instruct/utils.py
+++ b/open_instruct/utils.py
@@ -838,6 +838,7 @@ def clean_last_n_checkpoints(output_dir: str, keep_last_n_checkpoints: int) -> N
     # remove the last checkpoint to save space
     if keep_last_n_checkpoints > 0:
         folders = [f for f in os.listdir(output_dir) if is_checkpoint_folder(output_dir, f)]
+        # find the checkpoint with the largest step
         checkpoints = sorted(folders, key=lambda x: int(x.split("_")[-1]))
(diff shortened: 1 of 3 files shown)


open_instruct/utils.py
```diff
@@ -830,9 +830,14 @@ def get_last_checkpoint_path(args: FlatArguments, incomplete: bool = False) -> s
...
+def is_checkpoint_folder(dir: str, folder: str) -> bool:
+    return (folder.startswith("step_") or folder.startswith("epoch_")) and os.path.isdir(os.path.join(dir, folder))
+
+
 def clean_last_n_checkpoints(output_dir: str, keep_last_n_checkpoints: int) -> None:
     # remove the last checkpoint to save space
     if keep_last_n_checkpoints > 0:
-        checkpoints = sorted(os.listdir(output_dir), key=lambda x: int(x.split("_")[-1]))
+        folders = [f for f in os.listdir(output_dir) if is_checkpoint_folder(output_dir, f)]
+        checkpoints = sorted(folders, key=lambda x: int(x.split("_")[-1]))
```
vwxyzjn, line 842:
```suggestion
        # find the checkpoint with the largest step
        checkpoints = sorted(folders, key=lambda x: int(x.split("_")[-1]))
```

vwxyzjn (approved):

PR #274, opened 2024-08-19
Log beaker experiment URL

vwxyzjn:
This PR attempts to add the beaker experiment URL in wandb when possible, so you can directly go from wandb experiment to the beaker experiment. 

* https://wandb.ai/ai2-llm/open_instruct_internal/runs/ze11zq9p

image

image

https://github.com/user-attachments/assets/83092032-34b8-430b-a52d-3cd973be4ad0

### DIFF
--- a/open_instruct/dpo_tune.py
+++ b/open_instruct/dpo_tune.py
@@ -62,6 +62,7 @@ from open_instruct.utils import (
     clean_last_n_checkpoints,
     get_last_checkpoint_path,
     get_wandb_tags,
+    maybe_get_beaker_config,
     maybe_use_ai2_wandb_entity,
 )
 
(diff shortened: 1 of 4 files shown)


open_instruct/dpo_tune.py
```diff
@@ -523,7 +524,7 @@ def load_model():
     args.num_train_epochs = math.ceil(args.max_train_steps / num_update_steps_per_epoch)
 
     # Figure out how many steps we should save the Accelerator states
-    checkpointing_steps = args.checkpointing_steps
+    checkpointing_steps = str(args.checkpointing_steps)
```
vwxyzjn, line 527: I have to convert this to `str` because if the yaml file gives a 1000, `type(args.checkpointing_steps)` was actually int, so `checkpointing_steps.lower()` would error out.

open_instruct/utils.py
```diff
@@ -661,6 +665,24 @@ def __post_init__(self):
...
+@dataclass
+class BeakerRuntimeConfig:
+    beaker_workload_id: str
+    beaker_node_hostname: str
+    beaker_experiment_url: str
+
+
+def maybe_get_beaker_config():
+    beaker_runtime_config = None
+    if "BEAKER_JOB_ID" in os.environ:
```
hamishivi, line 677: nice!

PR #277, opened 2024-08-19
Auto evaluation via `--launch_beaker_eval_jobs`

vwxyzjn:
This PR empowers automatic evaluation of trained models, so we don't need to launch eval jobs manually anymore.

* Example beaker job: https://beaker.org/ex/01J5R69W07CVB9MKVBMP9A4091/, which at the end of training automatically submit the eval job (our poor man's dependency job)
* auto evaled beaker job: https://beaker.org/ex/01J5R6BQC4ZQ6YQ5GQMJNN1VNK/
image
image
image

* Requirement: the beaker workspace needs to have `BEAKER_TOKEN`. E.g., Need to run `beaker secret write BEAKER_TOKEN xxxxx`

You can reproduce the same thing with 

```

python scripts/submit_finetune_job.py --num_gpus 1 --priority normal  --image costah/open_instruct_test --config=configs/train_configs/sft/mini.yaml  --learning_rate 4e-5 --cluster ai2/s2-cirrascale  --output_dir /output/
python scripts/submit_dpo_job.py --num_gpus 1 --priority normal  --image costah/open_instruct_test --config=configs/train_configs/dpo/mini.yaml  --learning_rate 4e-5 --cluster ai2/s2-cirrascale  --output_dir /output/

```

- [x] switch the image back to `nathanl/open_instruct_olmo_auto`


scripts/submit_finetune_job.py
```diff
@@ -184,6 +184,9 @@ def parse_args(args):
     d['tasks'][0]['envVars'].append({
         'name': 'WANDB_RUN_GROUP', 'value': experiment_group
     })
+    d['tasks'][0]['envVars'].append({
```
vwxyzjn, line 186: Note that we have to add a `BEAKER_TOKEN` secret in the workspace.

hamishivi: Is is the beaker user token?

Dockerfile
```diff
@@ -22,6 +22,18 @@ RUN apt-get update && apt-get install -y \
     iputils-ping \
     tmux
 
+ARG BEAKER_VERSION
```
hamishivi, line 25: Can you also add this to `Dockerfile.olmo`? I'd like to keep the two dockerfiles synced up as much as possible for now.

open_instruct/dpo_tune.py
```diff
@@ -708,14 +725,21 @@ def load_model():
...
+            model,
+            tokenizer,
+            args.output_dir,
+            args.use_lora,
+            args.push_to_hub,
+            args.hf_repo_id,
+            args.hf_repo_revision,
+        )
+    if accelerator.is_main_process and args.launch_beaker_eval_jobs:
+        submit_beaker_eval_jobs(args.hf_repo_id, args.hf_repo_revision)
```
hamishivi: does this mean that submitting the beaker eval jobs requires setting push_to_hub? It might be good to assert that upfront to avoid this erroring after a long training run.

open_instruct/utils.py
```diff
@@ -639,6 +644,18 @@ class FlatArguments:
...
+    hf_entity: Optional[str] = None
+    """The user or org name of the model repository from the Hugging Face Hub"""
+    hf_repo_id: Optional[str] = None
+    """The id of the saved model in the Hugging Face Hub (can be autoset if not given)"""
+    hf_repo_revision: Optional[str] = None
+    """The revision of the saved model in the Hugging Face Hub (can be autoset if not given)"""
+    hf_repo_url: Optional[str] = None
+    """The url of the saved model in the Hugging Face Hub (will be autoset)"""
+    launch_beaker_eval_jobs: bool = False
+    """Whether to launch beaker evaluation jobs after training"""
```
hamishivi: as noted above, it looks like `launch_beaker_eval_jobs` requires `push_to_hub` to be set to true, so we should assert that in the post init.

open_instruct/utils.py
```diff
@@ -879,3 +868,74 @@ def clean_last_n_checkpoints(output_dir: str, keep_last_n_checkpoints: int) -> N
...
+    """Ai2 internal logic: try use the allenai entity if possible. Should not affect external users."""
+    orgs = HfApi().whoami()
+    orgs = [item["name"] for item in orgs["orgs"]]
+    if "allenai" in orgs:
+        return "allenai"
+    else:
+        return None
+
+
+def submit_beaker_eval_jobs(hf_repo_id: str, hf_repo_revision: str) -> None:
```
hamishivi: nice! I wonder if we should make this a bit more configurable  - being able to set workspace, beaker_image, upload to hf?

open_instruct/utils.py
```diff
@@ -879,3 +868,74 @@ def clean_last_n_checkpoints(output_dir: str, keep_last_n_checkpoints: int) -> N
...
+    command = f"""
+    python scripts/submit_eval_jobs.py \
+        --model_name hf-{hf_repo_revision} \
+        --hf_revision {hf_repo_revision} \
+        --location {hf_repo_id} \
+        --is_tuned \
+        --workspace tulu-3-results \
+        --preemptible \
+        --use_hf_tokenizer_template \
+        --beaker_image costah/open_instruct_test \
```
hamishivi: probably should swap to `nathanl/open_instruct_olmo_auto`, the autobuilt beaker image

open_instruct/dpo_tune.py
```diff
@@ -209,6 +213,21 @@ def main(args: FlatArguments):
...
+    exp_name = os.path.basename(__file__)[: -len(".py")]
+    if args.push_to_hub:
+        if args.hf_repo_id is None:  # auto-generate one
+            args.hf_repo_id = "open_instruct_dev"
+        if args.hf_entity is None:  # first try to use AI2 entity
+            args.hf_entity = maybe_use_ai2_hf_entity()
+        if args.hf_entity is None:  # then try to use the user's entity
+            args.hf_entity = HfApi().whoami()["name"]
+        args.hf_repo_id = f"{args.hf_entity}/{args.hf_repo_id}"
+        if args.hf_repo_revision is None:  # auto-generate one
```
hamishivi, line 511: so is the intended use to upload multiple runs to the same 'model' location but under different revisions? or each finetune run as a separate model?

hamishivi (commented):
Looks good, had some minor comments and small things to fix up.

scripts/submit_finetune_job.py
jacob-morrison: We've also been running the tulu 3 evals in a shared workspace, can we make the name of the secret customizable? So I can set (somewhere, cli flag or in the config) that my personal beaker token is named (e.g.) JACOBM_BEAKER_TOKEN?

open_instruct/utils.py
jacob-morrison: Why the olmo version over `nathanl/open_instruct_auto`?

open_instruct/finetune.py
```diff
@@ -150,40 +155,25 @@ def _concat_messages(messages):
...
-            state_dict=state_dict,
-            safe_serialization=False,
-        )
-
-
 def main(args: FlatArguments):
     # Initialize the accelerator. We will let the accelerator handle device placement for us in this example.
     # If we're using tracking, we also need to initialize it here and it will by default pick up all supported trackers
     # in the environment
+    exp_name = os.path.basename(__file__)[: -len(".py")]
```
yizhongw: Might be good to set this in `FlatArguments`?

scripts/submit_finetune_job.py
vwxyzjn: > Is is the beaker user token?
Yes, this is the beaker user token. We need to do `beaker secret write BEAKER_TOKEN xxxxx` in the eval workspace

> We've also been running the tulu 3 evals in a shared workspace, can we make the name of the secret customizable? So I can set (somewhere, cli flag or in the config) that my personal beaker token is named (e.g.) JACOBM_BEAKER_TOKEN?

We can only write a `BEAKER_TOKEN` in the eval workspace. It is possible to support multiple users. E.g., we could have 

```
beaker secret write JACOB_BEAKER_TOKEN xxxxx
beaker secret write COSTA_BEAKER_TOKEN xxxxx
beaker secret write HAMISH_BEAKER_TOKEN xxxxx
...
```


and then in runtime we can do something like `'name': 'BEAKER_TOKEN', 'secret': "$(whoami)$_BEAKER_TOKEN"` Would you prefer this solution? It can also generalize to HF_TOKEN, WANDB_API_KEY.

jacob-morrison: That would be perfect! (for all 3 of those tokens)

open_instruct/utils.py
vwxyzjn: Yep we should do that once the PR is merged. I need to use the custom image for testing out this workflow.

open_instruct/dpo_tune.py
vwxyzjn: Yes, something like this. 

image

We should mainly use the wandb to find models. The HF model stuff is just a dump (I suppose redundancy).

image

open_instruct/finetune.py
vwxyzjn: It's actually good to put here. Because then it returns `finetune` here and `dpo_tune` in the `dpo_tune.py`. If we put it in `FlatArguments`, `exp_name = "utils"`.

open_instruct/utils.py
vwxyzjn: Yeah let me work on that

vwxyzjn:
Btw truthfulqa is failing for some reason (https://beaker.org/ex/01J5R6BQC4ZQ6YQ5GQMJNN1VNK/tasks/01J5R6BQE5JHQFB9B8WKT5RAM9/job/01J5R6BZ10655Z7TJYH2A53R91). I need some help in looking at it

image

jacob-morrison:
> Btw truthfulqa is failing for some reason (https://beaker.org/ex/01J5R6BQC4ZQ6YQ5GQMJNN1VNK/tasks/01J5R6BQE5JHQFB9B8WKT5RAM9/job/01J5R6BZ10655Z7TJYH2A53R91). I need some help in looking at it
> 
> image

Ah I know exactly why this is failing! TruthfulQa also loads judge models, which go through the `load_hf_lm` code pathway. I see you're passing a model revision, and I bet that same revision is getting passed for the judge models as well (I accidentally did the same thing when I was evaluating OLMoE). This causes the judge model load to fail, which causes the given metric to not be calculated, which causes the key error you see there. If you make sure to only pass the revision for the model that has it, it should just work.

scripts/submit_finetune_job.py
vwxyzjn: Alright I will prepare some commands for people to run :D

open_instruct/finetune.py
```diff
@@ -710,11 +699,33 @@ def main(args: FlatArguments):
...
+            model,
+            tokenizer,
+            args.output_dir,
+            args.use_lora,
+        )
+
+    if accelerator.is_main_process:
+        if is_beaker_job() and args.try_launch_beaker_eval_jobs and args.output_dir in ["/output/", "/output"]:
+            submit_beaker_eval_jobs(model_name=run_name, location=beaker_config.beaker_dataset_ids[0])
+            args.try_launch_beaker_eval_jobs = False # don't launch the eval jobs for HF model then
```
vwxyzjn: FYI this could be the code to support launching eval using beaker dataset. However, need to make sure to sync the beaker dataset (see [source 1](https://beaker-py.readthedocs.io/en/latest/api/client.html#beaker.services.DatasetClient.sync), [source 2](https://allenai.slack.com/archives/C6MN19S05/p1724183018709369?thread_ts=1724182496.315879&cid=C6MN19S05)). I am not sure if it's worth it to implement the sync and stuff.

Maybe it's better to just use HF at the moment and we can revisit if it causes issues?

open_instruct/dpo_tune.py
vwxyzjn: This is a really good catch! I have put the `launch_beaker_eval_jobs` only under the push to hub in the script and added an assert statement.

open_instruct/utils.py
```diff
@@ -639,6 +645,18 @@ class FlatArguments:
...
+    """Whether to upload the saved model to huggingface"""
+    hf_entity: Optional[str] = None
+    """The user or org name of the model repository from the Hugging Face Hub"""
+    hf_repo_id: Optional[str] = None
+    """The id of the saved model in the Hugging Face Hub (can be autoset if not given)"""
+    hf_repo_revision: Optional[str] = None
+    """The revision of the saved model in the Hugging Face Hub (can be autoset if not given)"""
+    hf_repo_url: Optional[str] = None
+    """The url of the saved model in the Hugging Face Hub (will be autoset)"""
+    try_launch_beaker_eval_jobs: bool = False
```
vwxyzjn: Should I make this default to True?

vwxyzjn (commented):
Hey all thanks for the comments. I should have already addressed all of the comments except reverting the use of default image to the auto build one :) 

People should be able to use the autoeval provided they executed:

```
beaker secret write -w ai2/tulu-2-improvements $(whoami)_BEAKER_TOKEN xxxx
beaker secret write -w ai2/tulu-2-improvements $(whoami)_WANDB_API_KEY xxxx
beaker secret write -w ai2/tulu-2-improvements $(whoami)_HF_TOKEN xxxx
```

open_instruct/dpo_tune.py
```diff
@@ -56,19 +60,328 @@
...
 logger = get_logger(__name__)
 
 
+@dataclass
+class FlatArguments:
+    """
+    Full arguments class for all fine-tuning jobs.
+    """
+
+    exp_name: str = os.path.basename(__file__)[: -len(".py")]
```
vwxyzjn, line 88: I moved the `FlatArguments` here because then the `exp_name` would work out nicely to be the filename. 

Furthermore, it allows us to be flexible with the args. E.g., finetune.py doesn't need the DPO configs, and we can do a file diff later to figure out the differences.


image

open_instruct/dpo_tune.py
```diff
@@ -56,19 +60,328 @@
...
+        metadata={"help": "The rank of lora."},
+    )
+    lora_alpha: float = field(
+        default=16,
+        metadata={"help": "The alpha parameter of lora."},
+    )
+    lora_dropout: float = field(
+        default=0.1,
+        metadata={"help": "The dropout rate of lora modules."},
+    )
```
vwxyzjn, line 240: At some point, I should merge this with the `open_instruct.model_utils.ModelConfig`. Let keep things the way it is for now :)

open_instruct/dpo_tune.py
```diff
@@ -56,19 +60,328 @@
...
+    """Whether to upload the saved model to huggingface"""
+    hf_entity: Optional[str] = None
+    """The user or org name of the model repository from the Hugging Face Hub"""
+    hf_repo_id: Optional[str] = None
+    """The id of the saved model in the Hugging Face Hub (can be autoset if not given)"""
+    hf_repo_revision: Optional[str] = None
+    """The revision of the saved model in the Hugging Face Hub (can be autoset if not given)"""
+    hf_repo_url: Optional[str] = None
+    """The url of the saved model in the Hugging Face Hub (will be autoset)"""
+    try_launch_beaker_eval_jobs: bool = True
```
vwxyzjn, line 354: Let's do eval automatically and see how it goes.

vwxyzjn (commented):
I tested things again and all went smoothly. Could people review the PR again?

vwxyzjn:
I updated the docs here, too

image

open_instruct/utils.py
vwxyzjn: Have just revert back to the auto build image

hamishivi: olmo version because that uses the newest vllm, so its more useful at eval time. The benefit of the non-olmo image is that it has deepspeed installed, but that isn't useful at eval time.

open_instruct/dpo_tune.py
hamishivi: Would it be possible to just inherit from a shared flatarguments class instead, and then add the script-specific bits? Less code duplication that way, although I don't have strong feelings here.

scripts/submit_dpo_job.py
```diff
@@ -183,15 +183,21 @@ def parse_args(args):
...
-    #     'name': 'WANDB_API_KEY', 'value': wandb_api_key
-    # })
     d['tasks'][0]['envVars'].append({
         'name': 'WANDB_NAME', 'value': exp_name
     })
     d['tasks'][0]['envVars'].append({
         'name': 'WANDB_RUN_GROUP', 'value': experiment_group
     })
+    d['tasks'][0]['envVars'].append({
+        'name': 'BEAKER_TOKEN', 'secret': f"{getpass.getuser()}_BEAKER_TOKEN"
```
hamishivi: where does getpass get the username from? I ask since my username is different depending on what machine im using....

scripts/submit_finetune_job.py
hamishivi: I'm personally not a big fan of `"$(whoami)$_BEAKER_TOKEN"` because my username varies on different machines (laptop, ai2 servers, uw machines) and this adds some management. I feel this is ideally something that beaker should be responsible for, or otherwise can we have some non-user specific default somehow? (e.g. service account credential or something)

hamishivi (approved):
Looks good to me. Personally not a big fan of the `"$(whoami)$_BEAKER_TOKEN"` thing but I accept its a limitation of beaker for now.

scripts/submit_dpo_job.py
vwxyzjn: Now it's doing `get_beaker_whoami()` so hopefully it should work with whatever machine, as long as it has beaker on it.

vwxyzjn:
> Would it be possible to just inherit from a shared flatarguments class instead, and then add the script-specific bits? Less code duplication that way, although I don't have strong feelings here.

That's a good point. We should, for example, extract standalone configs like `open_instruct.utils.ModelConfig`. Let's keep things as is, and I will come back for another refactor about this.

vwxyzjn:
Merging as is.

PR #278, opened 2024-08-20
Propagate failure/preemptation and set timeout for multinode training

yizhongw:


configs/beaker_configs/default_finetune_lora_multinode.yaml
```diff
@@ -5,6 +5,9 @@ tasks:
     replicas: 4
     leaderSelection: true
     hostNetworking: true
+    propagateFailure: true
+    propagatePreemption: true
+    synchronizedStartTimeout: 30m
```
vwxyzjn: This seems like a very generous timeout. What about setting it to like 5 mins?

vwxyzjn (commented):
Ah cool change! What does `propagateFailure` do?

yizhongw:
> Ah cool change! What does `propagateFailure` do?

It will terminate all replicas if one of them fails. My understanding is this might be quicker than the distributed training system (e.g., deepspeed) finds no communication and terminates its processes, but I am not sure. There is some documentation for these parameters [here](https://beaker-docs.apps.allenai.org/experiments/distributed-training.html#batch-jobs) in beaker doc.

configs/beaker_configs/default_finetune_lora_multinode.yaml
yizhongw: 5 mins might be too short as one replica might be waiting the resources for get started. Let me set this to 15 mins (the recommended value in beaker doc).

vwxyzjn (approved):

PR #282, opened 2024-08-23
Add new eval Benchmarks for evaluating long context

nouhadziri:
in progress


hamishivi:
@nouhadziri do we still need these? I think oe-eval has the right support?

nouhadziri:
no you can close it. Yes, eval shifted to oe-eval.

PR #285, opened 2024-08-24
Patch checkpointing step

hamishivi:
https://github.com/allenai/open-instruct/pull/274#discussion_r1721912975 added a conversion to string that meant `checkpointing_steps` was `"None"` not `None`, so when checkpointing steps was not specified we would get errors.
This fixes that.


ValentinaPy (approved):
that should help!

PR #288, opened 2024-08-26
Remove unused checkpoints

vwxyzjn:
This PR does two things: 

1) speed up HF model upload / download through `HF_HUB_ENABLE_HF_TRANSFER=1`

2) at the end of the training, remove the checkpoints and save only the final model checkpoints. The current model checkpoint could save up to 300GB, which seems a bit excessive. https://beaker.org/ds/01J6351Y7VZ9DAQ5ZS18MCJKHR/details

### DIFF
--- a/open_instruct/utils.py
+++ b/open_instruct/utils.py
@@ -570,7 +570,8 @@ def clean_last_n_checkpoints(output_dir: str, keep_last_n_checkpoints: int) -> N
         # find the checkpoint with the largest step
         checkpoints = sorted(folders, key=lambda x: int(x.split("_")[-1]))
         if len(checkpoints) > keep_last_n_checkpoints:
-            shutil.rmtree(os.path.join(output_dir, checkpoints[0]))
+            for checkpoint in checkpoints[:len(checkpoints) - keep_last_n_checkpoints]:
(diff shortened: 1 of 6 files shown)


open_instruct/utils.py
```diff
@@ -570,7 +570,8 @@ def clean_last_n_checkpoints(output_dir: str, keep_last_n_checkpoints: int) -> N
         # find the checkpoint with the largest step
         checkpoints = sorted(folders, key=lambda x: int(x.split("_")[-1]))
         if len(checkpoints) > keep_last_n_checkpoints:
-            shutil.rmtree(os.path.join(output_dir, checkpoints[0]))
+            for checkpoint in checkpoints[:len(checkpoints) - keep_last_n_checkpoints]:
+                shutil.rmtree(os.path.join(output_dir, checkpoint))
```
natolambert, line 574: Not 100% sure what is going on here, but overall looks right.

natolambert (approved):
LGTM. One comment where I wasn't 100% on code functionality.

open_instruct/utils.py
vwxyzjn: It’s to only keep N checkpoints, if N=0 then remove all checkpoints

PR #294, opened 2024-08-26
Upload training metadata

hamishivi:
Gives the ability to upload training metadata when `hf_metadata_dataset` is set. This uploads a file called `metadata.json` to the location provided with the following info:
```bash
{
            "model_name": args.exp_name,
            "model_type": "sft",
            "datasets": dataset_list,
            "base_model": args.model_name_or_path,
            "wandb_path": wandb_tracker.run.get_url(),
            "beaker_experiment": beaker_config.beaker_experiment_url,
            "beaker_datasets": beaker_config.beaker_dataset_id_urls
        }
```

For now, I'll leave this not with any defaults set, but once we setup leaderboard 2.0 we can use this information to make it nicer. It's setup to try and make it to match up with the auto-evals naming scheme, with timestamps and stuff.

Additionally, fixed a small bug with beaker dataset urls (`maybe_get_beaker_config` wrapped the return of `get_beaker_dataset_ids` in a list, but `get_beaker_dataset_ids` already returns a list), and fixed a dockerfile issue with .git.

### DIFF
--- a/open_instruct/dpo_tune.py
+++ b/open_instruct/dpo_tune.py
@@ -851,6 +851,8 @@ def main(args: FlatArguments):
             experiment_config,
             init_kwargs={"wandb": {"entity": args.wandb_entity, "tags": [args.exp_name] + get_wandb_tags()}},
         )
+        wandb_tracker = accelerator.get_tracker("wandb")
+
 
     # Train!
     total_batch_size = args.per_device_train_batch_size * accelerator.num_processes * args.gradient_accumulation_steps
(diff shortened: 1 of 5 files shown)


open_instruct/dpo_tune.py
```diff
@@ -1034,6 +1038,27 @@ def load_model():
...
+            "wandb_path": wandb_tracker.run.get_url(),
+            "beaker_experiment": beaker_config.beaker_experiment_url,
+            "beaker_datasets": beaker_config.beaker_dataset_id_urls
+        }
+        upload_metadata_to_hf(
+            metadata_blob,
+            "metadata.json",
+            args.hf_metadata_dataset,
+            'results/' + args.hf_repo_revision,  # to match what the auto-evals name as.
+        )
```
vwxyzjn:
```suggestion
    if args.hf_metadata_dataset and accelerator.is_main_process and is_beaker_job():
        # dpo script only supports these two options right now for datasets
        dataset_name = args.dataset_name if args.dataset_name else args.train_file
        # mainly just focussing here on what would be useful for the leaderboard.
        # wandb will have even more useful information.
        metadata_blob = {
            "model_name": args.exp_name,
            "model_type": "sft",
            "datasets": [dataset_name],
            "base_model": args.model_name_or_path,
            "wandb_path": wandb_tracker.run.get_url(),
            "beaker_experiment": beaker_config.beaker_experiment_url,
            "beaker_datasets": beaker_config.beaker_dataset_id_urls
        }
        upload_metadata_to_hf(
            metadata_blob,
            "metadata.json",
            args.hf_metadata_dataset,
            'results/' + args.hf_repo_revision,  # to match what the auto-evals name as.
        )
```

`beaker_config` is already available above
https://github.com/allenai/open-instruct/blob/e77830ed5409eb4862ce5b4ad61ba560ae8d8a10/open_instruct/dpo_tune.py#L843

vwxyzjn (approved):
Nice!!!

PR #302, opened 2024-08-28
Remove oe-eval in docker image

vwxyzjn:
The oe-eval was blocked by https://github.com/allenai/oe-eval-internal/issues/214. Unfortunately it seems to be related to upstream issues in https://github.com/EleutherAI/lm-evaluation-harness. So the fix might be more difficult than expected.

To avoid blocking the current effort, let's revert to including the oe-eval in the image.

### DIFF
--- a/.github/workflows/push-image-olmo.yml
+++ b/.github/workflows/push-image-olmo.yml
@@ -45,12 +45,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - uses: actions/checkout@v3
-        with:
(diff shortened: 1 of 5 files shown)


yizhongw (commented):
Would you like to remove the `--run_oe_eval_experiments` in `submit_beaker_eval_jobs`?

vwxyzjn:
Nice catch! I also included a fix to `clean_last_n_checkpoints`, so you wouldn't be uploading 300GB of stuff. You would be only uploading the final checkpoint if the training is successful :)

yizhongw (approved):
Nice, looks good!

PR #306, opened 2024-08-28
Push rejection sampling dataset to beaker dataset as well.

vwxyzjn:
This PR puts the generated rejection sampling jsonls to beaker dataset as well. Maybe helpful to @fabrahman.

Example beaker job:

https://beaker.org/ex/01J6D0428H68FT8GN1ENWQ2QPH

image


fabrahman (approved):
approved!

PR #310, opened 2024-08-29
Auto eval actually works

vwxyzjn:
So auto eval that actually works (also with oe-eval). The idea is to do something like the following

10:30 - now testing with docker containers.

```python
    start_time = time.time()
    while time.time() - start_time < args.max_wait_time_for_beaker_dataset_upload_seconds:
        if beaker_experiment_succeeded(beaker_runtime_config.beaker_workload_id):
            print("Experiment succeeded")
            # NOTE: we are assuming the first beaker dataset has the model
            # I have checked a couple of beaker jobs and found the first dataset is the model
            # but we should check this assumption
            submit_beaker_eval_jobs(
                model_name=args.model_name,
                location=beaker_dataset_ids[0],
                run_oe_eval_experiments=True,
            )
            return
        time.sleep(args.check_interval_seconds)
    # If we reach here, the experiment failed
    print("Experiment failed")
    sys.exit(1)  # submit eval failed
```

image

# docker submission

* actual job https://beaker.org/ex/01J6FHK8NCWYMCNBCN8QJN80Y3
* wait for the actual job to upload models in beaker dataset, and then launch evals:  https://beaker.org/ex/01J6FHWKKFB6VD39R63YSGEQJS
* oe-eval: https://beaker.org/ex/01J6FJ2DET4MNYRWYEW75MABZ5
* oi-eval: https://beaker.org/ex/01J6FJ2E9S1E5JS7P9CZCH2V2T

# mason submission

* actual job https://beaker.org/ex/01J6DV30NX8XZ6NQMKAFGDC2TJ
* wait for the actual job to upload models in beaker dataset, and then launch evals: https://beaker.org/ex/01J6DV4FTTF03SFSF3ME60VS9A
* oe-eval: https://beaker.org/ex/01J6DV6PKWT2N3SK2CXYD1HK3Z
* oi-eval: https://beaker.org/ex/01J6DV6PSRX4Q02NTSKWTDD1RE


open_instruct/utils.py
```diff
@@ -681,26 +737,29 @@ def submit_beaker_eval_jobs(
...
 
-    logger.info(f"Beaker evaluation jobs: Stdout:\n{stdout.decode()}")
-    logger.error(f"Beaker evaluation jobs: Stderr:\n{stderr.decode()}")
-    logger.info(f"Beaker evaluation jobs: process return code: {process.returncode}")
+    print(f"Beaker evaluation jobs: Stdout:\n{stdout.decode()}")
+    print(f"Beaker evaluation jobs: Stderr:\n{stderr.decode()}")
+    print(f"Beaker evaluation jobs: process return code: {process.returncode}")
 
 
+@retry_on_exception()
```
vwxyzjn, line 755: Adding some retry on exception stuff. Helps with random network errors like the ones @yizhongw saw.

scripts/wait_beaker_dataset_model_upload_then_evaluate_model.py
```diff
@@ -0,0 +1,52 @@
...
+    beaker_dataset_ids = get_beaker_dataset_ids(beaker_runtime_config.beaker_workload_id)
+    print(beaker_experiment_succeeded(beaker_runtime_config.beaker_workload_id))
+
+    start_time = time.time()
+    while time.time() - start_time < args.max_wait_time_for_beaker_dataset_upload_seconds:
+        if beaker_experiment_succeeded(beaker_runtime_config.beaker_workload_id):
+            print("Experiment succeeded")
+            # NOTE: we are assuming the first beaker dataset has the model
+            # I have checked a couple of beaker jobs and found the first dataset is the model
+            # but we should check this assumption
```
vwxyzjn, line 37: @yizhongw @jacob-morrison for multi-node job I assumed the first beaker dataset will have the model. Is this always the case?

open_instruct/utils.py
```diff
@@ -704,7 +763,7 @@ def upload_metadata_to_hf(
     # about a model for leaderboard displays.
     with open("tmp.json", "w") as f:
         json.dump(metadata_dict, f)
-    api = HfApi(token=os.getenv("HF_TOKEN", None))
+    api = HfApi()
```
vwxyzjn, line 766: image

HF_TOKEN already overrides, so there is no need to pass explicitly

open_instruct/utils.py
yizhongw: great!

scripts/wait_beaker_dataset_model_upload_then_evaluate_model.py
yizhongw: I checked my experiment list. It seems so. I didn't see an exception.

open_instruct/finetune.py
```diff
@@ -1001,49 +1001,65 @@ def main(args: FlatArguments):
         )
 
     # remove all checkpoints to save space
-    if accelerator.is_main_process:
+    if accelerator.is_local_main_process:
```
vwxyzjn, line 1004: This should remove the intermedaite checkpoints on all the nodes.

yizhongw: Looks good!

natolambert (approved):
seems good

scripts/wait_beaker_dataset_model_upload_then_evaluate_model.py
hamishivi: my one concern is if this is the same when preemption happens, but thats sort of an edge case so happy to merge and then patch later.

hamishivi (approved):

PR #311, opened 2024-08-29
Faster build using docker cache

vwxyzjn:
The docker build is too slow. Trying to add cache for better build.


vwxyzjn:
Looks like a dead end. Docker cache may be too heavy 🥲

PR #314, opened 2024-08-29
Update eval suite

hamishivi:
Updating the eval suite. For tulu 3 evals, we now entirely rely on the oe-eval suite and nouhas safety eval fork (which hopefully will be added to the oe-eval tool soon). This means that we need an option to skip running the safey evals.

Also adding a script to upload metadata in case someone wants to update or add it post-hoc.


scripts/add_metadata.py
natolambert, line 1: Nice!

natolambert (approved):
Looks like what I would expect and solid.

vwxyzjn (approved):
Nice change! Could you add `--skip_oi_evals` for the auto eval as well?


https://github.com/allenai/open-instruct/blob/903d526499cfb37eb6dad385e19efe947d72c9d0/open_instruct/utils.py#L721-L739

open_instruct/utils.py
```diff
@@ -743,6 +745,10 @@ def submit_beaker_eval_jobs(
         command += f" --upload_to_hf {upload_to_hf}"
     if run_oe_eval_experiments:
         command += " --run_oe_eval_experiments"
+    if run_safety_evaluations:
+        command += " --run_safety_evaluations"
+    if skip_oi_evals:
```
hamishivi, line 750: @vwxyzjn `skip_oi_evals` is added here...

scripts/wait_beaker_dataset_model_upload_then_evaluate_model.py
```diff
@@ -39,6 +39,8 @@ def main(args: Args, beaker_runtime_config: BeakerRuntimeConfig):
                 model_name=args.model_name,
                 location=beaker_dataset_ids[0],
                 run_oe_eval_experiments=True,
+                run_safety_evaluations=True,
+                skip_oi_evals=True,
```
hamishivi, line 43: @vwxyzjn ...and then set here.

Issue #190, opened 2024-07-11
size mismatch error when running merge_lora.py


jl3676:
Hi! I'm trying to use the open-instruct code to finetune on a custom dataset. I was able to run the finetune code and save the model, but I encountered a size mismatch error at the last step `merge_lora.py`. Any advice on how I can work around this? Thanks in advance! 

The command I ran: 
```
python open_instruct/merge_lora.py --base_model_name_or_path meta-llama/Meta-Llama-3-8B --lora_model_name_or_path ../models/Meta-Llama-3-8B_qlora/ --output_dir ../models/Meta-Llama-3-8B_qlora_merged/ --qlora --save_tokenizer
```

Full terminal output: 
```
Loading the base model...
/home/jingjingl/.conda/envs/harm_project/lib/python3.10/site-packages/huggingface_hub/file_download.py:1132: FutureWarning: `resume_download` is deprecated and will be removed in version 1.0.0. Downloads always resume when possible. If you want to force a new download, use `force_download=True`.
  warnings.warn(
Loading checkpoint shards: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4/4 [00:04<00:00,  1.19s/it]
Dequantizing `model.layers.0.self_attn.q_proj`...
Dequantizing `model.layers.0.self_attn.k_proj`...
Dequantizing `model.layers.0.self_attn.v_proj`...
Dequantizing `model.layers.0.self_attn.o_proj`...
Dequantizing `model.layers.0.mlp.gate_proj`...
Dequantizing `model.layers.0.mlp.up_proj`...
Dequantizing `model.layers.0.mlp.down_proj`...
Dequantizing `model.layers.1.self_attn.q_proj`...
Dequantizing `model.layers.1.self_attn.k_proj`...
Dequantizing `model.layers.1.self_attn.v_proj`...
Dequantizing `model.layers.1.self_attn.o_proj`...
Dequantizing `model.layers.1.mlp.gate_proj`...
Dequantizing `model.layers.1.mlp.up_proj`...
Dequantizing `model.layers.1.mlp.down_proj`...
Dequantizing `model.layers.2.self_attn.q_proj`...
Dequantizing `model.layers.2.self_attn.k_proj`...
Dequantizing `model.layers.2.self_attn.v_proj`...
Dequantizing `model.layers.2.self_attn.o_proj`...
Dequantizing `model.layers.2.mlp.gate_proj`...
[...]

PR #315, opened 2024-08-29
Small fix to merge lora script

hamishivi:
As pointed out in #190, thanks to the new finetune script, which pads to a multiple of 8 (for cuda reasons), the lora merging breaks without also padding correctly. This adds a new argument `pad_to_multiple_of` to that script to allow this.


natolambert (approved):
lgtm

PR #316, opened 2024-08-29
Merge Dockerfiles

hamishivi:
Deepspeed supports pydantic 2 so now the vllm <-> deepspeed clash no longer exists. This PR removes the olmo/inference-specific dockerfile in favour of just having one. Since newer vllm requires it, I also updated the torch and flash-attn versions.

Seems to work for a small model training: https://beaker.org/ex/01J6G9DXY8NZF1TPJD7TKQ07RA/tasks/01J6G9DXYDMR77MWM5X22E5SPE/job/01J6GA38HHF0F0NPFPFCJHP7PZ


hamishivi:
Yup, we should replace. I need to test the image first though...

vwxyzjn (approved):
LGTM! Nice change!

PR #320, opened 2024-08-30
Upload metadata along with model weights

hamishivi:
I previously made an upload thing that was mainly targeted to auto-eval setup. This PR now adds:
- the metadata blob is included with the model weights, making it easier to persist through evals
- if metadata is present in the model directory (local or HF) when running an open-instruct eval, the code uploads the metadata alongside the results. This would be nice if we weren't sort of moving on from the open-instruct evals.

Tested and appears to work:
- https://huggingface.co/datasets/allenai/tulu-3-evals/tree/main/results/finetune__EleutherAI_pythia-14m__42__1725401265
- https://beaker.org/ex/01J6WZDT868FAPYSSM5GEP4HJE/tasks/01J6WZDV152RNECVX5V9GCZHG1/job/01J6WZDV7C0KY5K3SN57YY6F0S

Somewhat annoyingly, this doesnt affect the oe-evals and safety evals we are going to swap to use soon, but I figure its a useful update anyway.

### DIFF
--- a/eval/utils.py
+++ b/eval/utils.py
@@ -510,6 +510,7 @@ def upload_results_to_hf(
     os.remove("results.json")
 
 
+@retry_on_exception
 def check_and_upload_model_metadata(model_name_or_path, hf_dataset_name, hf_dataset_save_dir, hf_revision=None):
     # if metadata.json exists in the model directory, upload it to the dataset
     api = HfApi()
--- a/eval/utils.py
+++ b/eval/utils.py
@@ -500,7 +500,7 @@ def upload_results_to_hf(
     # actual save and upload
     with open("results.json", "w") as f:
         json.dump(results_dict, f)
-    api = HfApi(token=os.getenv("HF_TOKEN", None))
+    api = HfApi()
     api.upload_file(
         path_or_fileobj="results.json",
         path_in_repo=hf_dataset_save_path,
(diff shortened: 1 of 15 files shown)


eval/utils.py
```diff
@@ -500,11 +500,44 @@ def upload_results_to_hf(
...
     api.upload_file(
         path_or_fileobj="results.json",
         path_in_repo=hf_dataset_save_path,
         repo_id=hf_dataset_name,
         repo_type="dataset",
     )
     os.remove("results.json")
+
+
+def check_and_upload_model_metadata(model_name_or_path, hf_dataset_name, hf_dataset_save_dir, hf_revision=None):
```
vwxyzjn, line 514:
```suggestion
@retry_on_exception
def check_and_upload_model_metadata(model_name_or_path, hf_dataset_name, hf_dataset_save_dir, hf_revision=None):
```

Some error handling

eval/utils.py
```diff
@@ -500,11 +500,44 @@ def upload_results_to_hf(
...
+            hf_hub_download(
+                model_name_or_path,
+                filename="metadata.json",
+                local_dir=".",
+                revision=hf_revision,
+            )
+        except Exception as e:
+            print(f"Failed to download metadata.json from {model_name_or_path}")
+            print(e)
+            return
```
vwxyzjn, line 537: I didn't quite understand this part. If the `metadata.json` exists, are we trying to cloning it to the `{hf_dataset_save_dir}/metadata.json`?

vwxyzjn (commented):
Nice change!

eval/utils.py
hamishivi: If we loaded the model from HF, I don't think we have an easy way to get the metadata.json from the hf repo apart from this? So this is explicitly downloading it from there, to then upload to the eval results repo.

PR #321, opened 2024-08-31
Fix hf revision pass through

hamishivi:
OE-eval tool and safety eval tool didn't respect hf_revision when it was passed. This fixes that, should be useful for oe-eval.


natolambert (approved):
lgtm, thanks!

PR #326, opened 2024-09-03
Support mixing dataset for reward modeling

vwxyzjn:
Many improvements at once.

1. Now we support the data mixer:

```
python open_instruct/reward_modeling.py \
    --dataset_mixer '{"trl-internal-testing/sentiment-trl-style": 1.0, "ai2-adapt-dev/summarize_from_feedback_small": 1.0}' \
    --dataset_train_splits train train \
    --dataset_eval_mixer '{"trl-internal-testing/sentiment-trl-style": 1.0}' \
    --dataset_eval_splits test \
    --model_name_or_path EleutherAI/pythia-1b-deduped \
    --chat_template simple_concat_with_space \
    --learning_rate 3e-6 \
    --per_device_train_batch_size 64 \
    --per_device_eval_batch_size 64 \
    --gradient_accumulation_steps 1 \
    --max_token_length 1024 \
    --max_prompt_token_lenth 1024 \
    --num_train_epochs 1 \
    --output_dir models/rm/rm_sentiment_1b
```

2. printing is improved
image

3. we now try to cleverly figure out how many `num_proc` to use, based on allocated beaker CPU and how fast the apply chat template works. For example, in the screenshot below, we use different `num_proc` for filtering and `apply_chat_template` operations.

image

### DIFF
--- a/open_instruct/reward_modeling.py
+++ b/open_instruct/reward_modeling.py
@@ -3,7 +3,7 @@ import os
 import random
 import time
 from dataclasses import asdict, dataclass
-from typing import List, Literal, Optional
+from typing import List, Literal, Optional, Tuple
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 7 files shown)


open_instruct/reward_modeling.py
```diff
@@ -300,21 +331,43 @@ def main(args: Args, dataset_config: DatasetConfig, model_config: ModelConfig):
 
             if training_step % args.gradient_accumulation_steps == 0:
                 scheduler.step()
+                local_metrics[0] = accuracies.mean()
+                local_metrics[1] = losses.mean()
+                local_metrics[2] = chosen_rewards.mean()
+                local_metrics[3] = rejected_rewards.mean()
+                local_metrics[4] = reward_margin.mean()
+                global_metrics = accelerator.reduce(local_metrics, reduction="mean").tolist()
```
vwxyzjn, line 357: Doing nccl reduce just once.

open_instruct/utils.py
```diff
@@ -385,6 +385,99 @@ def get_datasets(
     return raw_datasets
 
 
+def combine_dataset(
```
vwxyzjn, line 388: @natolambert this is essentially the same as your `get_dataset` but focused on just mixing the splits of datasets without doing specialized logic.

I was going to use the `get_dataset` but it does not support other splits like `dev` or `validation`. We should refactor those logic and the handle different dataset logic out in the future.

natolambert: Yes we can refactor at future time. I had removed those features to reduce complexity :)

natolambert (approved):
This is awesome. Thanks for adding docs. We'll refactor when need. Great stuff!

vwxyzjn:
Thanks for the review! Merging as is so people can use it as an experimental feature. We will also prob do more iterations.

PR #329, opened 2024-09-04
Add retry method to eval utils

hamishivi:
I added a suggestion from costa without importing the function correctly (#320). Since the retry function lives in open_instruct.utils, but we want to use it in eval.utils, I just copied it over.


jacob-morrison (approved):

vwxyzjn (approved):
LGTM!

PR #335, opened 2024-09-06
Fix and improvements rejection sampling generation

vwxyzjn:


open_instruct/rejection_sampling/generation.py
```diff
@@ -146,22 +146,20 @@ def format_conversation(messages: list) -> str:
...
     if dataset_args.sanity_check:
-        for key in ds:
-            ds[key] = ds[key].select(range(min(dataset_args.sanity_check_size, len(ds[key]))))
+        ds = ds.select(range(min(dataset_args.sanity_check_size, len(ds))))
     if dataset_args.dataset_end_idx is None:
-        dataset_args.dataset_end_idx = len(ds[dataset_args.dataset_train_split])
-    for key in ds:
-        ds[key] = ds[key].select(range(dataset_args.dataset_start_idx, dataset_args.dataset_end_idx))
+        dataset_args.dataset_end_idx = len(ds)
+    ds = ds.select(range(dataset_args.dataset_start_idx, dataset_args.dataset_end_idx))
```
vwxyzjn, line 157: We are only using `ds[dataset_args.dataset_train_split]["prompt"]` anyway. Might as well just select the proper split

open_instruct/rejection_sampling/generation.py
```diff
@@ -95,17 +94,18 @@ async def generate_with_openai(model_name: str, data_list: list, args: Args, gen
 
 
 def generate_with_vllm(model_name_or_path: str, revision: str, prompt_token_ids: List[int], gen_args: GenerationArgs):
+    max_context_length = gen_args.response_length - 1
     llm = LLM(
         model=model_name_or_path,
         revision=revision,
         tokenizer_revision=revision,
         tensor_parallel_size=gen_args.tensor_parallel_size,
+        max_model_len=max_context_length,
```
vwxyzjn, line 103: Note here 

```
--max-model-len
Model context length. If unspecified, will be automatically derived from the model config.
```

PR #348, opened 2024-09-12
Suppor finetune / dpo with hf revision

vwxyzjn:
Some cleaning up.


configs/beaker_configs/default_finetune.yaml
```diff
@@ -17,7 +17,6 @@ tasks:
       open_instruct/finetune.py
       --model_name_or_path /hf_llama_models
       --use_flash_attn
-      --tokenizer_name /hf_llama_models
```
vwxyzjn, line 20: needed to remove this otherwise it's taking priority over the `model_revision`

natolambert (approved):
lgtm

PR #349, opened 2024-09-12
Allowing the RM to resize embedding to better utilize tensorcore

vwxyzjn:


vwxyzjn:
Given the minor changes, I am merging it as is.

PR #350, opened 2024-09-13
Clean the encoding function for messages and support using different chat templates

yizhongw:


yizhongw:
I did a test run [here](https://beaker.org/ex/01J7NCCAE1MB6Q9Q2TTYM0C1QJ/tasks/01J7NCCAE61MTNCNZXDHDBSGX3/job/01J7NCCAKMPV0S73EV5S2JHTDQ) and succeeded. 

I will double check the encoded messages are correctly masked for the training labels.

open_instruct/finetune.py
```diff
@@ -95,6 +95,13 @@ class FlatArguments:
         default="main",
         metadata={"help": "The specific model version to use (can be a branch name, tag name or commit id)."},
     )
+    chat_template_file: Optional[str] = field(
+        default=None,
+        metadata={"help": (
+            "A file with its content as a jinja2 template for encoding chat messages. "
+            "If not provided, we will use the default Tulu template."
+        )}
+    )
```
vwxyzjn: I love the ability to specify custom chat template. What do you think about doing something like below? We should be able to unify them in the future, but I didn't want to do a massive breaking change given the current sprint.

https://github.com/allenai/open-instruct/blob/de62ff4e8fc183e9158c2e57914b60b16c84c0fc/open_instruct/dataset_processor.py#L88-L136

https://github.com/allenai/open-instruct/blob/de62ff4e8fc183e9158c2e57914b60b16c84c0fc/open_instruct/dataset_processor.py#L147


https://github.com/allenai/open-instruct/blob/de62ff4e8fc183e9158c2e57914b60b16c84c0fc/open_instruct/reward_modeling.py#L238

yizhongw: Yeah, this should work well. 

Now I realize there are multiple places where we rely on the chat template (sft, dpo, rm training, rs), and they all have their different functions/collators to do the encoding. Shall we unify them? Probably need your help to do this then @vwxyzjn

vwxyzjn: That's what the [dataset_processor.py](https://github.com/allenai/open-instruct/blob/de62ff4e8fc183e9158c2e57914b60b16c84c0fc/open_instruct/dataset_processor.py#L88-L136) is for :) we can unify them post tulu 3.

Maybe for now what do you think about a simple fix like importing the `CHAT_TEMPLATE` from `open_instruct/dataset_processor.py`? 


If we do the unification now, I am worried it might introduce subtle bugs / regresssions, hindering the current project.

yizhongw: Fair, I will keep this change within the SFT part. Let me modify a bit to use these templates.

open_instruct/finetune.py
```diff
@@ -367,88 +374,71 @@ def __post_init__(self):
...
+        padding=False,
+        truncation=True,
+        max_length=max_seq_length,
+        add_generation_prompt=False,
+    )
     labels = input_ids.clone()
-
     # mask the non-assistant part for avoiding loss
     for message_idx, message in enumerate(messages):
         if message["role"] != "assistant":
```
hamishivi, line 401: Wont this also add loss for system messages too?

open_instruct/finetune.py
```diff
@@ -1065,7 +1043,7 @@ def main(args: FlatArguments):
             json.dump(metadata_blob, f)
 
         # upload metadata to the dataset if set
-        if args.hf_metadata_dataset:
+        if args.push_to_hub and args.hf_metadata_dataset:
```
hamishivi: Odd change? Only upload metadata if we push to HF?

hamishivi (commented):
I like the change! I agree with costa that it is probably best to leave this to after release stuff, but it makes lots of sense to lean more heavily on unified sources of truth. Left two comments.

open_instruct/finetune.py
yizhongw: This one should detect all messages that are *not* from the "assistant" and set the labels for them as -100. The non-assistant messages should include system message. So, I think it's fine?

open_instruct/finetune.py
yizhongw: Oh good catch. After checking what metadata is uploaded, I agree this shouldn't be added here.

I added there as this if branch (see below) uses the `args.hf_repo_revision`, which only exists when `push_to_hub` is set to be true. If we upload the meta data when there is no hf_repo_revisino, what would be a good argument to use to replace the `"results/" + args.hf_repo_revision` here (`hf_dataset_save_dir` argument of the function)?
```
        # upload metadata to the dataset if set
        if args.hf_metadata_dataset:
            upload_metadata_to_hf(
                metadata_blob,
                "metadata.json",
                args.hf_metadata_dataset,
                "results/" + args.hf_repo_revision,  # to match what the auto-evals name as.
            )
```

open_instruct/dataset_processor.py
```diff
@@ -108,29 +108,31 @@
...
+        "{% if message['role'] == 'user' %}"
+        "{{ '<|user|>\n' + message['content'] + eos_token + '\n' }}"
+        "{% elif message['role'] == 'system' %}"
+        "{{ '<|system|>\n' + message['content'] + eos_token + '\n' }}"
+        "{% elif message['role'] == 'assistant' %}"
+        "{{ '<|assistant|>\n'  + message['content'] + eos_token + '\n' }}"
+        "{% endif %}"
+        "{% if loop.last and add_generation_prompt %}"
+        "{{ '<|assistant|>\n' }}"
+        "{% endif %}"
```
vwxyzjn, line 128: So the main changes here is "not including a `\n` at `{% endif %}`"?

open_instruct/finetune.py
```diff
@@ -92,9 +93,18 @@ class FlatArguments:
         default=None, metadata={"help": "Pretrained tokenizer name or path if not the same as model_name"}
     )
     tokenizer_revision: Optional[str] = field(
-        default="main",
+        default=None,
         metadata={"help": "The specific model version to use (can be a branch name, tag name or commit id)."},
     )
+    chat_template_name: Optional[str] = field(
```
vwxyzjn:
```suggestion
    chat_template_name: str = field(
```

open_instruct/finetune.py
```diff
@@ -363,92 +373,78 @@ def __post_init__(self):
         ):
             raise ValueError("Cannot provide two dataset selection mechanisms.")
 
+        if self.chat_template_name is None or self.chat_template_name not in CHAT_TEMPLATES:
```
vwxyzjn:
```suggestion
        if self.chat_template_name not in CHAT_TEMPLATES:
```

If we want to use the chat template from the model itself, we should add another option like `chat_template_name = "default"`

open_instruct/finetune.py
```diff
@@ -363,92 +373,78 @@ def __post_init__(self):
...
+            raise ValueError(f"Invalid chat template name {self.chat_template_name}. Please choose from {CHAT_TEMPLATES.keys()}")
```
vwxyzjn: Nice!

vwxyzjn (commented):
Great PR @yizhongw! I put some comments down. The files changed seems a bit weird as it's showing changes from the main. Could you try merging main?

image


Also a quick question: are you planning to add the changes to `dpo_tune.py` as well?

open_instruct/finetune.py
hamishivi: oh wait, yeah, Im dumb.

open_instruct/finetune.py
hamishivi: Good question.... it should match however downstream chooses the name, which I'm not sure about on the top of my head.

open_instruct/dataset_processor.py
yizhongw: I moved the necessary `\n` to inside the double brackets. This change is mainly for reducing the confusion in the templates, as the `\n` outside blocks should be trimmed, according to this [codebase](https://github.com/huggingface/transformers/blob/98adf24883b007c2a7fb17bab1c01b1614673433/src/transformers/utils/chat_template_utils.py#L413). But this confuses me with the real `\n` that we want to keep.

For example, I changed `\n{{ '<|assistant|>' }}\n` to `{{ '<|assistant|>\n' }}`, as we want to keep the `\n` in the end as part of the generation prompt (the old template also did this, but it's just not very clear whether they are trimmed or not).

open_instruct/finetune.py
yizhongw: Probably we can allow users to provide a tokenizer name here as the chat_template_name? For example, if the user provides `meta-llama/Meta-Llama-3-8B-Instruct`, we will copy its template to be ours.

open_instruct/dataset_processor.py
vwxyzjn: Makes sense! Nice change!

yizhongw:
@vwxyzjn, feel free to take another look. I made it 1) support loading chat template by specifying a tokenizer in the `chat_template_name`, 2) use the the same logic in the DPO code.

open_instruct/finetune.py
```diff
@@ -682,22 +675,17 @@ def main(args: FlatArguments):
...
+    # set the tokenizer chat template to the training format
+    # this will be used for encoding the training examples
+    # and saved together with the tokenizer to be used later.
+    if args.chat_template_name in CHAT_TEMPLATES:
+        tokenizer.chat_template = CHAT_TEMPLATES[args.chat_template_name]
+    else:
+        try:
+            tokenizer.chat_template = AutoTokenizer.from_pretrained(args.chat_template_name).chat_template
+        except Exception as e:
+            raise ValueError(f"Could not find chat template for {args.chat_template_name}.")
```
vwxyzjn, line 687: nice

open_instruct/dpo_tune.py
```diff
@@ -744,6 +702,21 @@ def load_model():
...
+        tokenizer.chat_template = CHAT_TEMPLATES[args.chat_template_name]
+    else:
+        try:
+            tokenizer.chat_template = AutoTokenizer.from_pretrained(args.chat_template_name).chat_template
+        except Exception as e:
+            raise ValueError(f"Could not find chat template for {args.chat_template_name}.")
+
+    if args.add_bos:
+        # also add bos in the chat template
+        tokenizer.chat_template = "{{ bos_token }}" + tokenizer.chat_template
```
vwxyzjn, line 721: Should we assert `bos_token` not in `tokenizer.chat_template` if we decide to do this logic separately?

vwxyzjn (approved):
Thanks @yizhongw! This looks great to me. Could you run a sanity check experiment like the following, showing the curves don't look significantly different? (also doesn't have to be a finished experiment; maybe just run for 30 mins and check against an existing run)

 https://wandb.ai/ai2-llm/open_instruct_internal/reports/offline-DPO-regression-check--Vmlldzo5Mzg1NzAz 

image

yizhongw:
For later reference, while trying to compare the training performance using the new encoding function vs [the original tulu encoding function](https://github.com/allenai/open-instruct/blob/3add2ea758f4976d14b95eb94de18a46f7f08a66/open_instruct/finetune.py#L407-L420), I found there are some discrepancies:

1. The original [tulu encoding function](https://github.com/allenai/open-instruct/blob/3add2ea758f4976d14b95eb94de18a46f7f08a66/open_instruct/finetune.py#L407-L420) strips the while spaces after encoding the entire converstation. But for the new one, since we use `tokenizer.apply_chat_template` to get the non-assistant message masks, where intermediate messages will be passed to this function,  if we add stripping, it will also strip the `\n` when getting the masks, which leads to incorrect masking. So, after chatting with @vwxyzjn , we decided to use the following template:
```
        "{% for message in messages %}"
        "{% if message['role'] == 'system' %}"
        "{{ '<|system|>\n' + message['content'] + '\n' }}"
        "{% elif message['role'] == 'user' %}"
        "{{ '<|user|>\n' + message['content'] + '\n' }}"
        "{% elif message['role'] == 'assistant' %}"
        "{{ '<|assistant|>\n'  + message['content'] + eos_token + '\n' }}"
        "{% endif %}"
        "{% if loop.last and add_generation_prompt %}"
        "{{ '<|assistant|>\n' }}"
        "{% endif %}"
        "{% endfor %}"
```
When applying to a conversation, it will create a `\n` in the end, which only adds one token for SFT training, and would not affect the inference since eos_token will stops the generation. According to my experiments on our SFT preview mixture 3.4, this doesn't change the performance notably as well.

2. I found [this tokenization function](https://github.com/allenai/open-instruct/blob/3add2ea758f4976d14b95eb94de18a46f7f08a66/open_instruct/finetune.py#L423) in the original tulu template adds bos by default, even when we don't set add_bos=False outside. Our new encoding function will fix this issue.

open_instruct/dataset_processor.py
```diff
@@ -88,49 +88,58 @@
...
         "{% if loop.last and not add_generation_prompt %}{{ eos_token }}{% endif %}"
         "{% endfor %}"
     ),
+    "assistant_message_only": (
+        "{% for message in messages %}"
+        "{% if message['role'] == 'assistant' %}"
+        "{{ message['content'] }}"
+        "{% endif %}"
+        "{% endfor %}"
+    ),
```
yizhongw, line 116: @vwxyzjn @fabrahman This template could be used for continual pretraining when the data is formatted as messages in the following format:
`[{"role": "user", "content": ""}, {"role": "assistant", "content": "<your training document>"}]`

vwxyzjn:
Thanks @yizhongw! The changes all make sense to me.

PR #351, opened 2024-09-13
Only add IB env vars for multinode jobs

jacob-morrison:


jacob-morrison:
adding these seems to cause issues for single node jobs, so now we'll only add them for multinode

vwxyzjn (approved):
Makes sense! Feel free to merge!

PR #353, opened 2024-09-16
Fix auto eval with preemptible jobs

vwxyzjn:
image

To make sure auto eval works with pre-emptied and resumed jobs, we should try to find the dataset id which is 1) the latest and 2) non empty. This PR implements this.

This job was preemptied and resumed https://beaker.org/ex/01J7MQ76VRVK1AS36KX7XC997K/tasks/01J7MQ76VX5K6E63H972YMTB9X/job/01J7N0T5768XCXBGFVQ2RJE9M6.


vwxyzjn:
Merging as is.

PR #354, opened 2024-09-16
Log more stuff with DPO

vwxyzjn:
This PR just logs a lot more things with DPO:

https://wandb.ai/ai2-llm/open_instruct_internal/runs/zrahapo2

image

### DIFF
--- a/configs/train_configs/dpo/mini.yaml
+++ b/configs/train_configs/dpo/mini.yaml
@@ -3,7 +3,7 @@ model_revision: main
 use_flash_attn: true
 gradient_checkpointing: true
 tokenizer_name: EleutherAI/pythia-14m
-use_slow_tokenizer: true
+use_slow_tokenizer: null
 dataset_name: trl-internal-testing/tldr-preference-trl-style
 max_seq_length: 2048
 max_train_samples: 100
@@ -22,3 +22,5 @@ report_to:
 logging_steps: 1
 checkpointing_steps: 1000
 keep_last_n_checkpoints: 3
+try_launch_beaker_eval_jobs: false
+hf_metadata_dataset: ''
--- a/open_instruct/dpo_tune.py
+++ b/open_instruct/dpo_tune.py
@@ -947,11 +947,11 @@ def main(args: FlatArguments):
     # update the progress_bar if load from checkpoint
     progress_bar.update(completed_steps)
 
+    local_metrics = torch.zeros((20), device=accelerator.device)
+    episode = 0
     for epoch in range(starting_epoch, args.num_train_epochs):
         model.train()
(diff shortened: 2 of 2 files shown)


open_instruct/dpo_tune.py
```diff
@@ -1022,37 +1020,60 @@ def load_model():
...
             # Checks if the accelerator has performed an optimization step behind the scenes
             if accelerator.sync_gradients:
                 progress_bar.update(1)
                 completed_steps += 1
-                if args.logging_steps and completed_steps % args.logging_steps == 0:
-                    avg_loss = (
-                        accelerator.gather(total_loss).mean().item()
-                        / args.gradient_accumulation_steps
-                        / args.logging_steps
-                    )
```
vwxyzjn: ~~I am not sure if the existing logic is correct. When `logging_steps = 1`, and when `completed_steps = 1`, we are dividing the `args.gradient_accumulation_steps` incorrectly because there has not been any accumulated gradient steps, right?~~. Nvm, there was a flag `accelerator.sync_gradients` which is only triggered when the gradient accumulation steps are being fullfilled.

jacob-morrison (approved):
Nice!!

configs/train_configs/dpo/mini.yaml
```diff
@@ -3,7 +3,7 @@ model_revision: main
 use_flash_attn: true
 gradient_checkpointing: true
 tokenizer_name: EleutherAI/pythia-14m
-use_slow_tokenizer: true
+use_slow_tokenizer: null
```
hamishivi, line 6: ?

hamishivi (approved):
nice!

configs/train_configs/dpo/mini.yaml
vwxyzjn: I get



```
[rank0]: ValueError: Tokenizer class GPTNeoXTokenizer does not exist or is not currently imported.
[rank0]:[W916 10:35:25.965015944 ProcessGroupNCCL.cpp:1168] Warning: WARNING: process group has NOT been destroyed before we destruct ProcessGroupNCCL. On normal program exit, the application should call destroy_process_group to ensure that any pending NCCL operations have finished in this process. In rare cases this process can exit before this point and block the progress of another member of the process group. This constraint has always been present,  but this warning has only been added since PyTorch 2.4 (function operator())
```

vwxyzjn:
image


Everything worked as expected except KL. 
https://wandb.ai/ai2-llm/open_instruct_internal/reports/offline-DPO-regression-check--Vmlldzo5Mzg1NzAz


Since we are not generating on policy, we could not use off policy log probs to calculate KL, so I removed it from the PR.
image

PR #355, opened 2024-09-16
Resize embedding for reference model in DPO

yizhongw:


hamishivi (approved):

open_instruct/dpo_tune.py
```diff
@@ -738,6 +738,11 @@ def load_model():
     with deepspeed.zero.GatheredParameters(embeddings.weight, modifier_rank=None):
         if len(tokenizer) > embeddings.weight.shape[0]:
             model.resize_token_embeddings(len(tokenizer))
+    if reference_model is not None:
+        reference_embeddings = reference_model.get_input_embeddings()
+        with deepspeed.zero.GatheredParameters(reference_embeddings.weight, modifier_rank=None):
+            if len(tokenizer) > reference_embeddings.weight.shape[0]:
+                reference_model.resize_token_embeddings(len(tokenizer))
```
vwxyzjn, line 745: Minor nit: Is it possible the token embedding are different? If so, should we try to set reference model’s token embedding to be the same as the main model’s?

yizhongw: It shouldn't, with the current codebase. Both models are loaded by calling the same `load_model` function. It's just the same checkpoint being loaded twice.

PR #358, opened 2024-09-18
Fix auto eval parsing

vwxyzjn:


open_instruct/utils.py
```diff
@@ -734,14 +734,18 @@ def get_beaker_dataset_ids(experiment_id: str, sort=False) -> Optional[List[str]
...
+                        False if dataset["storage"]["totalSize"] is None else dataset["storage"]["totalSize"] > 0
+                    ),
                 )
                 for dataset in datasets
             ]
         )
     if sort:
         # sort based on empty, then commited
-        dataset_infos.sort(key=lambda x: (x.non_empty, datetime.strptime(x.committed, "%Y-%m-%dT%H:%M:%S.%fZ")))
+        dataset_infos.sort(key=lambda x: (x.non_empty, parser.parse(x.committed)))
```
vwxyzjn, line 748: This is a more robust parsing. @natolambert had `2024-09-18T18:37:26.634802914Z ValueError: time data '0001-01-01T00:00:00Z' does not match format '%Y-%m-%dT%H:%M:%S.%fZ'` error, which will be fixed by this PR.

https://beaker.org/ex/01J836MZ2DYJV26JT43X9TP7Y2/tasks/01J836MZ2H3ARGAVP2FQ1M4JXY/job/01J836MZ7JK799N2RZMQJG55EA

scripts/wait_beaker_dataset_model_upload_then_evaluate_model.py
```diff
@@ -25,8 +25,6 @@ class Args:
 
 def main(args: Args, beaker_runtime_config: BeakerRuntimeConfig):
     print(args)
-    beaker_dataset_ids = get_beaker_dataset_ids(beaker_runtime_config.beaker_workload_id, sort=True)
-    print(beaker_experiment_succeeded(beaker_runtime_config.beaker_workload_id))
```
vwxyzjn, line 29: We also don't need these runs. They are prob the culprit for `'0001-01-01T00:00:00Z'` because the job was still running.

natolambert (approved):
LGTM, merging

PR #360, opened 2024-09-19
Let reward modeling use ai2 entity

vwxyzjn:
Also the PR modify the docs to use the latest auto build image `nathanl/open_instruct_auto`.


ValentinaPy (approved):
great!

PR #363, opened 2024-09-23
Fix memory leakr in reward modeling

vwxyzjn:
This PR fixes a memory leak: see the chart below. No evidence of regression but this PR uses much less memory through the course of training.

image


vwxyzjn:
Tested it extensively. Merging as is.

PR #364, opened 2024-09-23
Online DPO optimization

vwxyzjn:
This PR attempts to optimize online DPO's end-to-end runtime (~10% faster). Specifically, I tried to re-use the active policy model's logprob for calculating the KL, saving a forward pass calculation time, unlike PPO. 

**KL calculation difference**: As demonstrated, there is no evidence of regression. However, the KL metrics will be calculated differently and will subject to some numerical precision issues, leading to the `objective/kl` being different.

Report is here: https://wandb.ai/ai2-llm/open_instruct_internal/reports/Online-DPO-optimization---Vmlldzo5NDM2NTgz

image


vwxyzjn:
Tested and confirmed there is no evidence of regression. Merging as is.

PR #365, opened 2024-09-24
Add gsm8k sft / generation dataset

vwxyzjn:
Dataset looks like this https://huggingface.co/datasets/ai2-adapt-dev/gsm8k

It's mainly for generation used in online training. E.g., with online trainers, the prompt tokens look like

image


natolambert (approved):
LGTM

PR #367, opened 2024-09-24
Create synthetic MMLU via GPT-4

nouhadziri:
Creation process:
1. Select random few-shot examples from each category in MMLU.
2. Prompt GPT4 with the few-shot examples to generate similar (but not identical) questions, options and responses.
3. Parse GPT4 outputs to extract responses and filter malformed outputs.
4. Upload data to HF.

### DIFF
--- a/open_instruct/mmlu_synthetic/generate_mmlu.py
+++ b/open_instruct/mmlu_synthetic/generate_mmlu.py
@@ -1,5 +1,4 @@
 import asyncio
-import random
 import json
 import re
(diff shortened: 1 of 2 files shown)


natolambert (changes requested):
@nouhadziri this looks good. We need to move the location though to match other data scripts
`scripts/data/...`

hamishivi:
Gonna close this since I think we've moved on -- feel free to reopen + prep for merging if you think otherwise!

PR #372, opened 2024-10-02
Minor fix to DPO

vwxyzjn:
The `if args.add_bos and if tokenizer.chat_template.startswith("{{ bos_token }}")` syntax causes minor issues.  This PR fixes it.


vwxyzjn:
Merging as is.

PR #374, opened 2024-10-03
Fix simpo metrics

vwxyzjn:
The issue is that `simpo` doesn't have reference log probs, and hence could not calculate things like accuracy :)


nouhadziri:
Looks great to me, thanks Costa for looking into this. I'm adding a readme on how to run dpo and then you can merge.

nouhadziri (approved):
LGTM!

PR #378, opened 2024-10-07
config files for MATH and IF added

fabrahman:
This commit added 3 config files for:
1) sft mix v3.6 (including all v3.4 + persona-augmented sft IF data) 
2) math specific sft mix (achieved 34.5 strict MATH)
3) preference data mix for IF eval


natolambert (approved):
Looks good. Quick things:
1. Can we try and add comments (brief) for what the datasets do in the pref configs.
2. Can we make the naming clearer. Let's make it something like tulu_pref_v3.X (and you can choose two numbers). No need for Faeze/math/etc it can be in the comments in the configs.

PR #379, opened 2024-10-07
Add configs for DPO

natolambert:


open_instruct/mix_data.py
```diff
@@ -14,12 +14,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from open_instruct.finetune import FlatArguments
+
 # script for mixing and saving data
-from .utils import ArgumentParserPlus, FlatArguments, get_datasets
+from open_instruct.utils import ArgumentParserPlus, get_datasets
```
hamishivi, line 20: Something that was missed in a prior update?

hamishivi (approved):

open_instruct/mix_data.py
natolambert: It's in the OLMo Again PR, but yes it should've been updated when @vwxyzjn refactored args

PR #384, opened 2024-10-10
[qol] making leaderboard exps easier to search

vwxyzjn:
This PR makes leaderboard experiments easier to search.

## Current limitation

We have 

image

directly searching yields nothing
image

We have to search by filtering "hf_repo_revision"

image

## This PR

This PR follows the same logic as online dpo / ppo which makes hf_repo_revision as the wandb run name, which make things much easier to search.


nouhadziri:
LGTM, this will make our life easier. Thanks Costa for adding this.

PR #385, opened 2024-10-10
Make the oi safety eval beaker name shorter

vwxyzjn:
This is a bit of ugly hack for now, needs better resolutions in the future.

image


nouhadziri (approved):
LGTM, thanks Costa!!

PR #386, opened 2024-10-12
GPT model synthetic data generation

VictoriaGraf:
Outputs of GPT model completions reformatted so that they can be added to the output table correctly.


nouhadziri (approved):
Sorry for the delay. LGTM, thanks Victoria!

PR #388, opened 2024-10-14
Add ds configs, better analysis

natolambert:


configs/train_configs/sft/tulu3_8b_preview_mix_v3.7.yaml
jacob-morrison, line 1: this is an old version, so I don't think we need to keep this. I'm open to changing my mind, but the latest version is essentially this w/ Yizhong's filtered datasets and a few other tweaks

scripts/data/get_statistics_tulu_v3.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+# Run with scripts/data/get_statistics_tulu_v3.sh
+
+# List of datasets
+datasets=(
```
jacob-morrison, line 5: is there any way we can connect this directly to the training configs, rather than hardcoding?

another problem I foresee for now is that this calculates statistics over the entire dataset, rather than the subsets we used (for subsampled datasets), which will be a little misleading. I'm planning to make static subsets after we finalize (for real) sft so we can use the other prompts for preference learning, but good to be aware of either way (e.g. we use 50k out of 2.5m from openmath-2-gsm8k)

jacob-morrison (changes requested):

jacob-morrison (approved):

natolambert:
@jacob-morrison I agree with these. Will improve when we make more final versions.

PR #392, opened 2024-10-18
Decontamination scripts

pdasigi:
Scripts for indexing training datasets and querying them with evaluation sets to quantify contamination, using exact, ngram, or embedding-based matching.


natolambert (approved):
Looks great. Can you also add a note to the top level readme on what this directory is for?

I wasn't sure if it should be its own top level folder, but I think it should be.

PR #395, opened 2024-10-23
Faster, less memory by caching dpo logprobs

vwxyzjn:
This PR caches the DPO logprobs to save memory (see [report](https://wandb.ai/ai2-llm/open_instruct_internal/reports/DPO-cache-logprobs--Vmlldzo5ODM5Mjkw))

image

I put this as a separate `dpo_tune_cach.py`. We can prob the  `dpo_tune.py` with this implementation, but maybe there are some edge cases like dataloader resuming with dpo cache logprobs or lora stuff to test, so I am also ok with just a `dpo_tune_cache.py` at the moment until we are done with the release. WDYT?

### DIFF
--- a/open_instruct/dpo_tune_cache.py
+++ b/open_instruct/dpo_tune_cache.py
@@ -1,0 +1,1162 @@
+# !/usr/bin/env python
+# coding=utf-8
+# Copyright 2024 AllenAI. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+DPO tuning script. Adapted from our finetuning script.
+"""
+
+import json
+import logging
+import math
+import os
+import random
+import subprocess
+import time
+from copy import deepcopy
+from dataclasses import dataclass, field
+from datetime import timedelta
+from functools import partial
+from typing import List, Optional, Union
+
+import datasets
+import deepspeed
+import torch
+import torch.utils
+import torch.utils.data
+import transformers
+from accelerate import Accelerator
+from accelerate.logging import get_logger
+from accelerate.utils import InitProcessGroupKwargs, set_seed
+from datasets import load_dataset
+from huggingface_hub import HfApi
+from peft import LoraConfig, TaskType, get_peft_model, prepare_model_for_kbit_training
+from torch.utils.data import DataLoader
+from tqdm.auto import tqdm
+from transformers import (
+    AutoConfig,
+    AutoModelForCausalLM,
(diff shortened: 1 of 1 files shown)


open_instruct/dpo_tune_cache.py
```diff
@@ -0,0 +1,1169 @@
...
+    progress_bar.update(completed_steps)
+
+    local_metrics = torch.zeros((20), device=accelerator.device)
+    episode = 0
+    for epoch in range(starting_epoch, args.num_train_epochs):
+        model.train()
+        train_dataloader.set_epoch(epoch)
+        if last_checkpoint_path and resume_step is not None:
+            # We skip the first `n` batches in the dataloader when resuming from a checkpoint
+            active_dataloader = accelerator.skip_first_batches(train_dataloader, resume_step)
```
hamishivi, line 957: If we skip initial steps, and then resume, I don't think `step` gets updated, but we need that to reference the cached logprobs correctly?

open_instruct/dpo_tune_cache.py
```diff
@@ -0,0 +1,1169 @@
...
+            resume_step -= starting_epoch * len(train_dataloader)
+
+    print(f"Starting from epoch {starting_epoch} and step {completed_steps}.")
+
+    # Cache the logprobs
+    average_log_prob_loss_types = ["simpo", "dpo_norm"]
+    average_log_prob = args.dpo_loss_type in average_log_prob_loss_types
+    if args.dpo_loss_type == "dpo" or args.dpo_loss_type == "dpo_norm":
+        cached_reference_chosen_logps, cached_reference_rejected_logps = get_cache_ref_logprobs(
+            model, train_dataloader, accelerator, average_log_prob
```
hamishivi: Did  you double check that the order of the samples here matches the order of the samples later? Doesn't the dataloader reshuffle every epoch?

hamishivi (commented):
Left some comments... the curves look good but i want to be really certain that this is not changing things.

open_instruct/dpo_tune_cache.py
vwxyzjn: image

Yeah, the order of the samples changes. I added logic to handle the epochs stuff.

image

open_instruct/dpo_tune_cache.py
vwxyzjn: Added the corresponding logic to the PR.

vwxyzjn:
I have added the logic to handle those two scenarios. @hamishivi WDYT about merging this version as is (not replacing the existing `dpo_tune.py`) and let people test it out more extensively before merging?

open_instruct/dpo_tune_cache.py
hamishivi: Right, but in a multi-epoch setting don't we need to change how we index into the cached logprobs to match the new ordering when we swap into a new epoch? Our configs now use 1 epoch, but it'd be good to support multi-epoch.

hamishivi:
@vwxyzjn Yeah, happy to keep this separate from dpo_tune for now in the interest of time. I think its fine to let people test out, it might be good to test this on a few datasets and hyperparameter settings against the original dpo script to be safe that the caching logic is fine. I've been bit on adding caching logic and then it actually being wrong before, hence my caution 😅

hamishivi (approved):

vwxyzjn:
Of course @hamishivi! Thanks for checking closely. Also tested it with 3 epochs :) 

image

PR #399, opened 2024-10-23
Update submit_eval_jobs.py

jacob-morrison:
Include new clusters in the default arg for evaluations


vwxyzjn (approved):
Perfect! Thanks Jacob!

PR #403, opened 2024-10-24
Decontaminate training datasets

pdasigi:
The search script only produced dataset level contamination scores earlier. Now there is an option to also output decontaminated training sets.


natolambert (approved):
great addition!

PR #390, opened 2024-10-16
Prototype ppo + ray

vwxyzjn:
TL;DR works
image

Chat models works.
image

70B works
image

PR #405, opened 2024-10-28
Typo fix

vwxyzjn:
Does the `datasets_next` impact performance? If I understand correctly, `datsets_next` is created only once.


vwxyzjn:
Closed in favor of #390

PR #406, opened 2024-10-28
Finalize v3.8 sft mix

jacob-morrison:
The 70B hparams aren't finalized yet, and I may be able to improve the 8B model a bit more with some more tuning, but the 8B is relatively final. The model I trained is listed as `L3.1-8B-v3.8-nc-final__meta-llama_Llama-3.1-8B__42__1729991287` on the leaderboard


configs/beaker_configs/default_finetune_multinode.yaml
```diff
@@ -62,6 +62,10 @@ tasks:
         value: true
       - name: HF_TOKEN
         secret: HF_TOKEN
+    datasets:
+      - mountPath: /oe-adapt-default
+        source:
+          weka: oe-adapt-default
```
natolambert, line 68: Is this in the single node config? Why just being added now?

natolambert (approved):
LGTM -- one minor question

configs/beaker_configs/default_finetune_multinode.yaml
jacob-morrison: Ah good catch, just added it (also to the dpo file). It wasn't there before because we weren't commonly using files on weka, and I added it here to access the static training file. We'll also need to use our weka bucket for 70b training (downloading the 70b model as a mounted beaker dataset is way too slow), so this'll be helpful going forward as well.

PR #407, opened 2024-10-28
Support multiple gpus for safety eval

hamishivi:
Adds the ability to use tensor parallelism (via vllm) with safety eval.


natolambert (approved):
LGTM - thanks

PR #412, opened 2024-10-31
A good DPO mix

ValentinaPy:
A good DPO mix with LJ's regenerations of Ultrafeedback and Helpsteer2!


natolambert (approved):
LGTM. Also can just make this a numbered version and increment numbers.

PR #413, opened 2024-11-01
added persona ifdata to dpo mix

fabrahman:
added persona IF data to the final dpo mix


ValentinaPy (approved):
Great! Thank you! LGTM

PR #416, opened 2024-11-05
Adding final nc configs for v3.9

jacob-morrison:
Each config contains reasonable SFT settings for 8B and 70B (OLMo(E) coming later), the full data mixer list (commented out), and the final static mix saved on HF.


natolambert (approved):
LGTM. Merging so we can do some updates to paper. Thanks for making duplicates in ai2-adapt-dev.

PR #421, opened 2024-11-06
Minor additions to the decontamination script

pdasigi:
- Adding safety evals
- Write parquet files when decontaminating datasets (instead of jsonl)


natolambert (approved):
LGTM. Merging so we don't forget it.

PR #425, opened 2024-11-10
Unseen evals n the oe-eval.sh script

pdasigi:


natolambert (approved):
We should add an arg that works with `submit_eval_jobs.py` too. Can do that in this PR or after.

pdasigi:
@natolambert I added a flag to `submit_eval_jobs.py`. The current setting is to either run dev evals or unseen evals, not both. I think this should be fine. Do you see any issues?

natolambert:
Nope, seems good. I think we'll prolly manually run unseen to keep things clean.

PR #433, opened 2024-11-13
Latest OLMOE config

natolambert:


configs/beaker_configs/default_finetune_multinode_transformers_main.yaml
```diff
@@ -0,0 +1,75 @@
+version: v2
```
vwxyzjn, line 1: How is this file different than the other multinode file?

natolambert: @vwxyzjn installs from main rather than branch?

PR #434, opened 2024-11-13
Olmo1124converter

vwxyzjn:


natolambert (approved):
Converter looks good. Seems like your beaker config is a repeat of https://github.com/allenai/open-instruct/blob/main/configs/beaker_configs/default_finetune_multinode_olmo1124.yaml?

PR #439, opened 2024-11-17
Update mmlu and deepmind_math configs in oe-eval.sh

pdasigi:
`deemind_math:0shot_cot-v3` additionally uses example outputs in the prompt.
`mmlu:cot::summarize` asks the model to "summarize" its reasoning instead of making it "step-by-step".

Both changes generally result in improvements in a range of models.


natolambert (approved):
LGTM. These are changes I expected :)

PR #441, opened 2024-11-20
faster mmlu oe-eval

hamishivi:


natolambert (approved):
LGTM. Thanks for including it @hamishivi

PR #468, opened 2024-12-01
Fix bug. The parameter --keep_last_n_checkpoints -1 doesn't work. 

shizhengLi:
Modify logic of clean_last_n_checkpoints method in utils.py file.


shizhengLi:
Hi, in the open-instruct/open_instruct/finetune.py file, there is a parameter:
` keep_last_n_checkpoints: int = field(
        default=3,
        metadata={"help": "How many checkpoints to keep in the output directory. -1 for all."},
    )`

`--keep_last_n_checkpoints -1` means keeping all the checkpoints. When I set it, it didn't work, so I just fixed it.

hamishivi:
Closing as this was fixed by #541

Issue #280, opened 2024-08-21
Make faster docker build


vwxyzjn:
This stuff never changes. We should just use a prebuilt one or build one ourselves to save time building. Many times I am just waiting for the miniconda install.

https://github.com/allenai/open-instruct/blob/58342771988f57a7b4365cf4c0e7d655e5cdf571/Dockerfile#L1-L66

PR #492, opened 2024-12-20
uv2

vwxyzjn:
This PR does a few things:
* Bump the dependencies (torch, etc)
* (dev only) introduce uv as the package manager to *only* lock dependencies (requirements.txt); people should still use the existing `pip install` workflow until further notice.
* Optimize the docker file further: we will now use a base `ghcr.io/allenai/oi-cuda-no-conda:12.1-cudnn8-dev-ubuntu20.04`, so build and load time should be substantially lower.


hamishivi:
Is there a particular process to update the uv lockfile?
Also, how does building the docker image work for non-ai2 folks? is the beaker image accessible, or do they have to go through more steps? I glanced at the documentation but didn't see anything describing that (sorry if I missed it).

vwxyzjn:
> Is there a particular process to update the uv lockfile?

Yes. Three ways:
* `uv add packageA`
* `uv lock --upgrade-package packageA`
* directly editing the main files, and run `uv lock` or `uv sync`

> Also, how does building the docker image work for non-ai2 folks?

I made the image and it's public, so anyone should be able to download it. They can also build their own image, too, via

```
 docker build --build-arg CUDA=12.1.0 --build-arg TARGET=cudnn8-devel --build-arg DIST=ubuntu20.04 -f  Dockerfile.base . -t cuda-no-conda:12.1-cudnn8-dev-ubuntu20.04
```

image

natolambert (approved):
Slick stuff! Thanks! Inspires me to try and add tests :D

hamishivi (approved):
thanks for the explanation!

hamishivi:
This probably closes #280

fabrahman (approved):
This looks great! thanks

PR #499, opened 2025-01-06
Add Enforce Eager Flag

hamishivi:
Enforce eager stops crashes ocurring when using llama 3.1 with TP=16 (see https://github.com/vllm-project/vllm/issues/6783), this adds the ability to turn it on via flag.

It would be better to use pipeline parallelism, but this requires changing the class of the vllm worker to use AsyncLLMEngine instead of LLMEngine, so its a bigger change.


vwxyzjn (approved):

vwxyzjn:
LGTM. Though it shouldn't need to. I remember if I set TP=8 I ran into similar issues. Maybe there is something with the world size = 17 or 9? Need further diagnosis, but looks good for now.

hamishivi:
Yeah, this is an issue on the vLLM side, but ideally in the future we can just not use this flag once it's resolved :)

PR #504, opened 2025-01-08
Add documentation on caching models.

vwxyzjn:


mason.py
```diff
@@ -478,10 +478,12 @@ def main():
 
     beaker_secrets = [secret.name for secret in beaker_client.workspace.secrets()]
     whoami = beaker_client.account.whoami().name
+    max_retries = 3
```
hamishivi: so now 3 retries is default for all mason jobs? Can we make this configurable?

hamishivi (approved):
LGTM! I think if we want to make 3 retries default we should also make it configurable -- i have the odd case where i dont want a default retry (e.g. 405B jobs, which often crash at the end anyway due to uploading being too slow, but I don't want to retry because the weights are in the beaker job/weka anyway).

vwxyzjn:
Good point! I just made the max_retries to be 1 and configurable.

PR #505, opened 2025-01-08
Add `--try_auto_save_to_beaker` arg

vwxyzjn:
Modified the logic the respect whatever `output_dir` given, but will try to save to `/output` if possible. In 405B tuning we should save to weak and specifically specify `--try_auto_save_to_beaker False` so we are not spending 24 hours just to upload the checkpoints.


hamishivi (approved):
Looks good to me!

PR #507, opened 2025-01-10
Apply accelerate change to dpo cache script

hamishivi:
#492 updated some stuff and accelerate call needed to be updated. This should fix.
Also fix the bug where `accelerator.is_main_process == 0` should be `accelerator.is_main_process`


vwxyzjn (approved):
Thanks for the catch!!

Issue #506, opened 2025-01-10
Unable to generate Synth_Pref dataset


ranarag:
Hi,
Thanks for sharing the code for synthectic preference dataset generation!
I have been trying to generate synthetic data using synth_pref, but I encountered some challenges while annotating preferences using scripts.synth_pref.annotate_preferences.
 I have been able to convert the annotation mix to the format required by the Batch API by using:
```
python3 -m scripts.synth_pref.annotate_preferences convert \
    --model gpt-4o-2024-08-06 \
    --input_path scripts/synth_pref/example/create_annotation_mix_out/full/helpfulness-full.jsonl \
    --output_dir scripts/synth_pref/example/create_annotation_mix_out/batch_openai \
    --rows_per_shard 10000
```
This creates instances like:

```
[...]

PR #511, opened 2025-01-13
Fix synth_pref functions

ljvmiranda921:
Reference #506

- Add missing functions 😓


ljvmiranda921:
Nice! I moved it to the main `.gitignore`. Will merge once checks are good!

PR #513, opened 2025-01-13
Allow using tokenizer chat template

hamishivi:
In the current RLVR code, we always override the tokenizer chat template. We should allow using a custom chat template defined the tokenizer (this makes using models off of HF easier).


vwxyzjn (approved):
LGTM. Could you also apply the change to `ppo_vllm_thread_ray.py`?

PR #514, opened 2025-01-13
Push evaluation results into the datalake

vwxyzjn:
Should be something like this:

image


saumyamalik:
Looks good to me!

saumyamalik (approved):
Looks good to me!

PR #522, opened 2025-01-16
Add dataset cache / mixing support

vwxyzjn:


vwxyzjn:
Confirmed that the tokenized datasets are exactly the same for SFT / DPO

image
image

docs/ai2_internal.md
```diff
@@ -1,30 +1,255 @@
...
+## Job submission
+
 `mason.py` is our job submission script. It takes in the command after `--` and runs it in the specified clusters. During the job submission, it automatically tries to setup a shared Hugging Face cache with environment variables. For example, it sets
 * `HF_HOME=/weka/oe-adapt-default/allennlp/.cache/huggingface`. 
 * `HF_DATASETS_CACHE=/weka/oe-adapt-default/allennlp/.cache/huggingface`
 * `HF_HUB_CACHE=/weka/oe-adapt-default/allennlp/.cache/hub`
 
-As a result, the `allenai/Llama-3.1-Tulu-3-8B-DPO` and `allenai/RLVR-GSM-MATH-IF-Mixed-Constraints` will be cached in the shared file system.
+
+You can run things like below for a quick spin.
```
natolambert:
```suggestion
You can run things like below for a quick spin. This example just starts a beaker job to print the Python version in your beaker image.
```

docs/ai2_internal.md
```diff
@@ -1,30 +1,255 @@
...
+    --cluster ai2/jupiter-cirrascale-2 ai2/saturn-cirrascale ai2/neptune-cirrascale \
+    --workspace ai2/tulu-3-dev \
+    --image nathanl/open_instruct_auto --pure_docker_mode \
+    --priority normal \
+    --budget ai2/oe-adapt \
+    --gpus 1 -- which python
+```
+
+### Supervised Fine-tuning (SFT):
+
```
natolambert, line 45:
```suggestion

Otherwise, the `mason.py` command can be used to launch all of our other training jobs.
```

docs/ai2_internal.md
```diff
@@ -1,30 +1,255 @@
...
+    --logging_steps 1 \
+    --reduce_loss sum \
+    --model_revision main \
+    --dataset_mixer_list allenai/tulu-3-sft-mixture 100 \
+    --dataset_mix_dir output/sft_8b \
+    --exp_name tulu-3-8b-sft \
+    --seed 123
+```
+
+
```
natolambert, line 59: Whats different about the second one?

docs/ai2_internal.md
```diff
@@ -1,30 +1,255 @@
...
+✅ Found cached dataset at https://huggingface.co/datasets/vwxyzjn/dataset-mix-cached/tree/992c2b51ba
+
+# ...
+
+Kicked off Beaker job. https://beaker.org/ex/01JJ50D88M757GZD14W9CNN7NT
+```
+
+It would be most helpful if you run the `mason.py` command on a vscode session with access to weka, that way, the dataset is also automatically downloaded to `HF_HOME`, etc.
+
+When you inspect the job, it's going to have the following outputs, meaning the cached dataset is found and used:
```
natolambert, line 134: Do we want to be able to upload cached datasets to HF or is just weka fine? seems like a future problem.

open_instruct/dpo_tune_cache1.py
natolambert, line 1: Do we need these duplicate files? This and finetune1.py. Maybe give them better names if needed?

natolambert (approved):
LGTM some minor nits and double check maybe redundant files

docs/ai2_internal.md
vwxyzjn: This one just replaces the `--cluster` with `ai2/augusta-google-1`. Let me remove it for now.

docs/ai2_internal.md
vwxyzjn: It's both to HF and weka. When you run it in the weka vscode session, it caches the dataset into the weka HF cache, making loading < 1 seconds, versus the current one which tokenized for 5 minutes.

vwxyzjn (commented):
I removed the duplicate files.

open_instruct/dpo_tune_cache1.py
vwxyzjn: I removed them. Previously `finetune1.py` does not support `--dataset_mixer` (the yaml) input, but I just added the support, so I think it's fine to completely replace `finetune.py`.

PR #523, opened 2025-01-17
Optionally save value model + GRPO

hamishivi:
I want to be able to look at the value models we learn and see what happens, so this is to add that :)
Also adds GRPO ray code.

Need to test all this out but is exciting.


vwxyzjn (approved):
LGTM. Feel free to merge whenever.

open_instruct/grpo_vllm_thread_ray_gtrl.py
```diff
@@ -1142,16 +1141,6 @@ def vllm_generate(
                     kl = kl2
                 elif args.kl_estimator == "kl3":
                     kl = kl3
-                # if self.rank==0:
-                #     print(f"{logprobs[0][:40]=}, {ref_logprobs[0][:40]=}, {kl.sum(1)=}")
-                non_score_reward = -args.beta * kl
-                non_score_reward_sum = non_score_reward.sum(1)
-                rlhf_reward = scores + non_score_reward_sum
```
vwxyzjn: Let's still log the `rlhf_reward`

PR #525, opened 2025-01-23
Olmo2 and olmoe support

vwxyzjn:


configs/beaker_configs/ray_node_setup.sh
```diff
@@ -1,6 +1,7 @@
 export CURRENT_DATETIME=$(python -c "import datetime; import pytz; print(datetime.datetime.now(pytz.timezone('America/Los_Angeles')).strftime('%m%d%y_%H%M%S'))")
 export PYTHONPATH=$REPO_PATH
 export PATH="/root/.local/bin:$PATH"
+export NCCL_CUMEM_ENABLE=0
```
vwxyzjn, line 4: This apparently helps with the multi-node stuff, so I am putting it here @hamishivi

open_instruct/reward_modeling.py
```diff
@@ -241,7 +244,8 @@ def main(args: Args, dataset_config: DatasetConfig, model_config: ModelConfig):
         tokenizer.pad_token_id = 128002  # <|reserved_special_token_0|>
     else:
         tokenizer.add_special_tokens({"pad_token": "[PAD]"})  # NOTE: we do not resize the embedding
-    tokenizer.chat_template = CHAT_TEMPLATES[dataset_config.chat_template]
+    if dataset_config.chat_template is not None:
```
natolambert, line 247: A little surprised to see this here.

open_instruct/olmo_adapter/__init__.py
```diff
@@ -0,0 +1,226 @@
...
+
+import torch
+from torch import nn
+from torch.nn import MSELoss, CrossEntropyLoss, BCEWithLogitsLoss
+from transformers.modeling_outputs import SequenceClassifierOutputWithPast
+from transformers.models.olmo2.modeling_olmo2 import Olmo2Config, Olmo2Model, Olmo2PreTrainedModel
+from transformers.models.olmoe.modeling_olmoe import OlmoeConfig, OlmoeModel, OlmoePreTrainedModel
+
+
+class Olmo2ForSequenceClassification(Olmo2PreTrainedModel):
```
natolambert, line 11: Can we link where we got this? I'm guessing its from the OLMo repo?

natolambert (approved):
Looks fine, a few small comments

open_instruct/ppo_vllm_thread_ray_gtrl.py
```diff
@@ -709,7 +712,7 @@ def from_pretrained(
             disable_dropout_in_model(self.reward_model)
             ds_config = get_eval_ds_config(
                 offload=False,
-                stage=args.deepspeed_stage,
+                stage=args.deepspeed_stage if args.deepspeed_stage == 3 else 0,
```
natolambert, line 719: This is also a little surprising. A little funny but must just be how it works. What does it fix?

open_instruct/reward_modeling.py
vwxyzjn: Yeah so the logic is if DPO already modifies the chat template, we don't try to modify it again. This is especially important for OLMo with the add_bos baked in to the chat template. 

We will refactor this anyway down the line in #522

open_instruct/olmo_adapter/__init__.py
vwxyzjn: No this is from the transformer subclass. I just implemented it based on how other models are done in transformers.  

In theory this would be better to live in the transformers upstream, but it's a bit of effort going through the community contributions stuff, so I am just doing it in our repo so life is easier.

open_instruct/ppo_vllm_thread_ray_gtrl.py
vwxyzjn: Yeah reference model / reward model does not have a stage 2.

Stage 2 is optimizer sharding. If we pass stage 2 like done for olmoe, then it errors out with stage 2 requiring an optimizer.

vwxyzjn: I am gonna add a note there.

Issue #530, opened 2025-01-29
KL loss should be differentiable in GRPO


gauravpandeyamu:
The terms involved in computing the KL loss [here](https://github.com/allenai/open-instruct/blob/ff52f32db680e8fb9d00f5c0eec871b3d6616614/open_instruct/grpo_vllm_thread_ray_gtrl.py#L1135) are computed under torch.no_grad().

Hence, adding the KL loss to the GRPO loss [here](https://github.com/allenai/open-instruct/blob/ff52f32db680e8fb9d00f5c0eec871b3d6616614/open_instruct/grpo_vllm_thread_ray_gtrl.py#L1183)  will not result in minimization of KL divergence.

The code needs to be modified so that the logprobs term in KL can be differentiated against.

PR #531, opened 2025-01-29
Kl loss should be differentiable in GRPO

gauravpandeyamu:
Fixes https://github.com/allenai/open-instruct/issues/530


vwxyzjn (approved):
Thanks for this great catch!! This is very helpful. Merging as is.

gauravpandeyamu:
Thanks :)

Issue #534, opened 2025-01-29
GRPO implementation update


vwxyzjn:
Let's use this issue to share the latest GRPO development updates. CC @gauravpandeyamu thanks for your fix.

The command below (without @gauravpandeyamu's fix) yields the charts below. Overall the training score and sequence length goes up, but the downstream eval in MATH seems to suffer. I am gonna try out @gauravpandeyamu's fix on KL regularization to see if it helps.

```
for beta in 0.03 0.01 0.0; do
for nspp in 16 32; do
for m in half-m ; do
local_rollout_batch_size=8
if [ $m == "half-m" ]; then
    local_mini_batch_size=$(($local_rollout_batch_size * $nspp / 2))
else
    local_mini_batch_size=$(($local_rollout_batch_size * $nspp))
fi
exp_name="0128_grpo_math_zs_${beta}_${nspp}_${m}_${RANDOM}"
full_bsz=$(($local_rollout_batch_size * nspp * (8 + 8 + 8 + 7) * 2))
echo $exp_name:
echo --- local_mini_batch_size=$local_mini_batch_size
echo --- full_bsz=$full_bsz
echo --- num_gradient_updates=$(($local_rollout_batch_size * $nspp / $local_mini_batch_size))
exp_name="0128_grpo_math_zs_${beta}_${nspp}_${RANDOM}"
python mason.py \
    --cluster ai2/jupiter-cirrascale-2 \
    --workspace ai2/tulu-3-dev \
    --priority high \
    --preemptible \
    --num_nodes 1 \
    --max_retries 1 \
    --budget ai2/oe-adapt \
    --gpus 8 -- source configs/beaker_configs/ray_node_setup.sh \&\& uv run python open_instruct/grpo_vllm_thread_ray_gtrl.py \
    --exp_name $exp_name \
    --beta $beta \
    --local_mini_batch_size $local_mini_batch_size \
    --number_samples_per_prompt $nspp \
    --output_dir /weka/oe-adapt-default/costah/models/$exp_name \
    --local_rollout_batch_size $local_rollout_batch_size \
    --dataset_mixer "{\"ai2-adapt-dev/math_ground_truth_zs\": 1.0}" \
    --dataset_train_splits train \
    --dataset_eval_mixer "{\"ai2-adapt-dev/math_ground_truth_zs\": 32}" \
    --dataset_eval_splits train \
    --max_token_length 2048 \
    --max_prompt_token_length 2048 \
    --response_length 2048 \
[...]

PR #535, opened 2025-01-30
GRPO loss fix

vwxyzjn:
CC @gauravpandeyamu @hamishivi for a second look. 

I think the `pg_loss = pg_loss + (args.beta * kl).mean()` is incorrect because the mean operation happens after the scores are added with the corresponding KL

image

The correct impl should be 

```
pg_loss = masked_mean(pg_loss_max + (args.beta * kl), ~padding_mask[micro_batch_inds])
```

Let me know what you think.

### DIFF
--- a/open_instruct/grpo_vllm_thread_ray_gtrl.py
+++ b/open_instruct/grpo_vllm_thread_ray_gtrl.py
@@ -490,19 +490,20 @@ def remove_padding(sequences, pad_token_id):
 class MetricsTracker:
     """A simple class to prellocate all metrics in an array
     so we can do only one allreduce operation to get the metrics mean"""
+
     def __init__(self, max_metrics: int = 32, device: torch.device = torch.device("cuda")):
         self.metrics = torch.zeros(max_metrics, device=device)
         self.names2idx = {}
         self.current_idx = 0
         self.max_metrics = max_metrics
-    
+
     def add(self, name: str, value: torch.tensor):
         if name not in self.names2idx:
             if self.current_idx >= self.max_metrics:
                 raise ValueError(f"Exceeded maximum number of metrics ({self.max_metrics})")
             self.names2idx[name] = self.current_idx
             self.current_idx += 1
-        
+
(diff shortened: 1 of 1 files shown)


gauravpandeyamu:
Yes, masked_mean is indeed correct to use here.

Also check my comment at https://github.com/allenai/open-instruct/issues/534#issuecomment-2624863523

The current kl1 estimator will completely ignore the ref_logprobs during trainining. kl2 and kl3 are fine.

open_instruct/grpo_vllm_thread_ray_gtrl.py
```diff
@@ -487,6 +487,34 @@ def remove_padding(sequences, pad_token_id):
     return [[inneritem for inneritem in item if inneritem != pad_token_id] for item in sequences]
 
 
+class MetricsTracker:
```
hamishivi, line 490: nice!

open_instruct/grpo_vllm_thread_ray_gtrl.py
```diff
@@ -1170,25 +1206,32 @@ def vllm_generate(
...
-                        # recalculate kl: difference from PPO because we want the KL loss
-                        # to backpropagate through the model
-                        # TODO: we potentially need to clip this kl loss in case some logits
-                        # exploded, but more on it later.
-                        kl1 = new_logprobs - ref_logprobs[micro_batch_inds]
-                        kl2 = (kl1) ** 2 / 2
-                        kl3 = (-kl1).exp() - 1 + kl1
+                        # Here we recalculate kl: we want the KL loss to backpropagate through the model
+                        # We also clamp the KL loss to avoid numerical instability
+                        # https://chatgpt.com/c/679c4aa5-a558-8011-a1de-142381105e56
```
hamishivi: I can't load this conversation. Maybe needs to be made public or rehost somewhere?

hamishivi (approved):
Looks good to me. Nice on the metrics improvement!

PR #538, opened 2025-01-31
Clean up rlvr a lil, add base support

hamishivi:
- Remove some of the hardcoded stuff in the RLVR code that assumed tulu chat template. Now we pass response-only + prompt+response to the verifiers. The latter isn't used right now, but I imagine it might be useful in the future.
- Add deepseek R1 zero template
- Add ability to set a stop string, needed for RL on base models.

### DIFF
--- a/open_instruct/dataset_processor.py
+++ b/open_instruct/dataset_processor.py
@@ -156,7 +156,15 @@ CHAT_TEMPLATES = {
         "The reasoning process and answer are enclosed within <think> </think>"
         "and <answer> </answer> tags, respectively,"
         "i.e., <think> reasoning process here </think>"
-        "<answer> answer here </answer>. User: prompt. Assistant:"
+        "<answer> answer here </answer>."
+        "\n\n"
+        "{% for message in messages %}"
+        "{{ '\n\n' if not loop.first else '' }}"
+        "{{ message['role'].capitalize() + ': ' + message['content'] + '\n' }}"
+        "{% if loop.last and add_generation_prompt %}"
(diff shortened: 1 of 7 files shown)


open_instruct/dataset_processor.py
```diff
@@ -148,6 +148,16 @@
...
     ),
+    "deepseek_r1_zero": (
+        "A conversation between User and Assistant."
+        "The user asks a question, and the Assistant solves it."
+        "The assistant first thinks about the reasoning process in"
+        "the mind and then provides the user with the answer."
+        "The reasoning process and answer are enclosed within <think> </think>"
+        "and <answer> </answer> tags, respectively,"
+        "i.e., <think> reasoning process here </think>"
+        "<answer> answer here </answer>. User: prompt. Assistant:"
```
vwxyzjn: Shouldn't it be something like this?

```bash
    "r1-simple-chat": (
        "A conversation between User and Assistant."
        "The user asks a question, and the Assistant solves it."
        "The assistant first thinks about the reasoning process in"
        "the mind and then provides the user with the answer."
        "The reasoning process and answer are enclosed within <think> </think>"
        "and <answer> </answer> tags, respectively,"
        "i.e., <think> reasoning process here </think>"
        "<answer> answer here </answer>."
        "{% for message in messages %}"
        "{{ '\n\n' if not loop.first else '' }}"
        "{{ message['role'].capitalize() + ': ' + message['content'] }}"
        "{% if loop.last and not add_generation_prompt %}{{ eos_token }}{% endif %}"
        "{% endfor %}"
    ),
```

PR #548, opened 2025-02-04
Properly set max length for eval

hamishivi:
This sets max length for eval generation and allows it to be passed down in the GRPO/PPO scripts.

### DIFF
--- a/open_instruct/ppo_vllm_thread_ray_gtrl.py
+++ b/open_instruct/ppo_vllm_thread_ray_gtrl.py
@@ -220,6 +220,7 @@ class Args:
     stop_strings: List[str] = None
     """List of strings that stop the generation when they are generated.
     The returned output will not contain the stop strings."""
+    eval_max_length: int = 4096  # max generation length for evaluation
 
     # online PPO specific args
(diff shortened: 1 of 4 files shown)


open_instruct/ppo_vllm_thread_ray_gtrl.py
```diff
@@ -1510,7 +1511,10 @@ def launch_ai2_evals_on_weka(self, step_dir: str, training_step: Optional[int] =
     --run_safety_evaluations \
     --run_id {wandb_url} \
     --step {training_step} \
+    --oe_eval_max_length {args.eval_max_length} \
     --skip_oi_evals"""
+            if training_step is not None:
+                command += f" --step {training_step}"
```
vwxyzjn: This is duplicated? 

image

vwxyzjn (approved):
LGTM except a very minor comment

open_instruct/ppo_vllm_thread_ray_gtrl.py
hamishivi: Ah!

PR #551, opened 2025-02-05
Add optional r1-style thinking reward

vwxyzjn:
https://beaker.allen.ai/orgs/ai2/workspaces/tulu-3-dev/work/01JKA2SEKQ1E2QJN9SVSKGRKFQ?taskId=01JKA2SEKXC7BP2XB5KJC1C9TA&jobId=01JKA2SERKZ0JWH0HGFRETSYEX

### DIFF
--- a/open_instruct/ground_truth_utils.py
+++ b/open_instruct/ground_truth_utils.py
@@ -1,12 +1,21 @@
-'''
+"""
 Collection of 'ground truth rewards' for different datasets/tasks.
 Used to give feedback to the model based on the ground truth answer.
-'''
-import re
+"""
+
 import json
+import re
 import string
-from open_instruct.math_utils import last_boxed_only_string, remove_boxed, get_unnormalized_answer, normalize_final_answer, is_equiv, hendrycks_is_equiv
+
 from open_instruct.if_functions import IF_FUNCTIONS_MAP
+from open_instruct.math_utils import (
+    get_unnormalized_answer,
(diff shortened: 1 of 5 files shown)


open_instruct/ground_truth_utils.py
```diff
@@ -137,9 +137,128 @@ def verify_flan_sample(model_output, ground_truth_answer):
...
+"""
+    invalid = "Invalid format"
+    
+    result = strict_format_reward_func([valid, invalid])
+    assert result == [1.0, 0.0], f"Multiple responses failed, got {result}"
+    print("✓ Multiple responses passed")
+
 
 # debug code
 if __name__ == "__main__":
```
natolambert, line 158: @vwxyzjn lets add this to the tests check we have?
Also, let's make the scale of the reward set by a hyperparam / config? Could get tricky reward shaping issues.

natolambert (approved):
Approved but see two small requests.

open_instruct/ground_truth_utils.py
```diff
@@ -139,7 +139,7 @@ def verify_flan_sample(model_output, ground_truth_answer):
 
 def soft_format_reward_func(responses: list[str]) -> list[float]:
```
hamishivi: Would agree with nathan that it feels like reward weight should be a param (and maybe even what pattern you are looking for?), to help with tuning reward stuff in the future.

hamishivi (approved):

PR #555, opened 2025-02-07
removing incorrect date cutoffs

natolambert:
~900 of our SFT examples have things like "As of my last update in 2023" which makes the model learn an incorrect date cutoff. Adding a script for removing this.

I filtered the following datasets:
* Origin tulu 3 sft --> https://huggingface.co/datasets/allenai/tulu-3-sft-mixture-filter-datecutoff
* OLMo 2 SFT --> https://huggingface.co/datasets/allenai/tulu-3-sft-olmo-2-mixture-filter-datecutoff
* Tulu 405B pref data (has ~400 entries) --> https://huggingface.co/datasets/allenai/llama-3.1-tulu-3-405b-preference-mixture-filter-datecutoff

We should run this before every final-ish training run.

### DIFF
--- a/scripts/data/sft/filter_cutoff_date.py
+++ b/scripts/data/sft/filter_cutoff_date.py
@@ -1,0 +1,118 @@
+import re
+import argparse
+from datasets import load_dataset
+from typing import List, Dict
+from datasets import Features, Value, Sequence
+
+"""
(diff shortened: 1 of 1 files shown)


vwxyzjn (approved):
LGTM!

Issue #537, opened 2025-01-31
Scheduler Issue in PPO/GRPO implementation


ashish230897:
Hi,

This piece of code in both ppo/grpo codebase seems incorrect (for args.num_samples_per_prompt > 1):

```
args.num_training_steps = args.total_episodes // (args.rollout_batch_size * args.number_samples_per_prompt)
num_training_steps = args.num_training_steps * args.num_train_epochs * args.num_epochs
warm_up_steps = args.warm_up_steps
if args.warmup_ratio >= 0.0:
    warm_up_steps = int(num_training_steps * args.warmup_ratio)
scheduler = get_scheduler(
        args.lr_scheduler_type,
        optimizer=self.optimizer,
        num_warmup_steps=warm_up_steps,
        num_training_steps=num_training_steps,
    )
```

When args.number_samples_per_prompt > 1, the learning rate will reduce to 0 faster if we divide by args.number_samples_per_prompt.
For args.number_samples_per_prompt > 1, there are multiple updates taking place in a single training step; so for every model.step() happening inside a training step, the learning rate decreases and eventually becomes 0 earlier than it should be. 
The above num_training_steps will work fine when num_samples_per_prompt is 1.

Also, the args.num_train_epochs seems redundant in the code.

To fix this, we can edit the code as:
```
num_scheduler_steps = args.num_training_steps * args.num_epochs * args.number_samples_per_prompt
warm_up_steps = args.warm_up_steps
if args.warmup_ratio >= 0.0:
    warm_up_steps = int(num_scheduler_steps * args.warmup_ratio)
scheduler = get_scheduler(
        args.lr_scheduler_type,
        optimizer=self.optimizer,
        num_warmup_steps=warm_up_steps,
        num_training_steps=num_scheduler_steps,
    )
```

PR #560, opened 2025-02-11
Fix scheduler and dataset shuffling

vwxyzjn:
Closes https://github.com/allenai/open-instruct/issues/537 and fix the dataset shuffling logic.

E.g., https://wandb.ai/ai2-llm/open_instruct_internal/runs/52afsubq/workspace?nw=nwusercostah

image

seems reasonable: we train for 10M episodes, by 1M, the lr decreased from 5e-7 to 4.5e-7.

### DIFF
--- a/open_instruct/dpo_tune_cache.py
+++ b/open_instruct/dpo_tune_cache.py
@@ -542,7 +542,7 @@ def main(args: FlatArguments):
             args.max_seq_length,
             args.hf_entity,
         )
-        train_dataset.shuffle(seed=args.seed)
+        train_dataset = train_dataset.shuffle(seed=args.seed)
         train_dataset.set_format(type="pt")
     if args.cache_dataset_only:
         return
--- a/open_instruct/grpo_vllm_thread_ray_gtrl.py
+++ b/open_instruct/grpo_vllm_thread_ray_gtrl.py
@@ -151,8 +151,6 @@ class Args:
     """Ratio of warmup steps to total steps (takes precedence over `warm_up_steps`)"""
 
     # various batch sizes
-    num_train_epochs: int = 1
-    """Number of epochs to train"""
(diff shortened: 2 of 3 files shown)


open_instruct/dpo_tune_cache.py
```diff
@@ -542,7 +542,7 @@ def main(args: FlatArguments):
             args.max_seq_length,
             args.hf_entity,
         )
-        train_dataset.shuffle(seed=args.seed)
+        train_dataset = train_dataset.shuffle(seed=args.seed)
```
hamishivi, line 545: ah, nice catch

open_instruct/grpo_vllm_thread_ray_gtrl.py
```diff
@@ -642,15 +641,15 @@ def from_pretrained(
         # optim_params = get_optimizer_grouped_parameters(self.policy, weight_decay)
         # self.optimizer = AdamOptimizer(optim_params, lr=args.learning_rate)
         self.optimizer = torch.optim.AdamW(self.policy.parameters(), lr=args.learning_rate)
-        num_training_steps = args.num_training_steps * args.num_train_epochs * args.num_epochs
+        num_scheduler_steps = args.num_training_steps * args.num_epochs * args.num_mini_batches
```
hamishivi, line 644: I guess the scheduler steps every ministep, not ministep*grad acc.

hamishivi (approved):
Seems good! I didn't run the math on your example but seems valid.

Relatedly, I wonder if we should just set constant LR as the default? Why use the cooldown?

open_instruct/grpo_vllm_thread_ray_gtrl.py
vwxyzjn: This is because we do a gradient step at each mini batches (i.g., local_mini_batch_size)

PR #566, opened 2025-02-14
Ppo cleanup

vwxyzjn:
Cleanup for PPO: using the same interface that GRPO now uses


hamishivi (approved):
Thanks for this... is there a way to run on 1 GPU now tho?

vwxyzjn:
It should absolutely doable. Let me merge this PR and think about how to support 1 GPU debugging.

image

PR #567, opened 2025-02-14
Support single GPU RLVR

vwxyzjn:
image

image

tested GRPO as well

image


vwxyzjn:
Note that for single GPU we have to use the gloo backend -- otherwise nccl errors out with "duplicate GPU"

hamishivi (approved):
LGTM!

PR #569, opened 2025-02-15
Using 'train' split when loading cached dataset

coding-famer:
I recommend using the ‘train’ split when loading the cached dataset temporarily. This is because any dataset uploaded to the Hugging Face Hub without a specified split defaults to 'train'. Currently, the code is unable to load a test split.


vwxyzjn:
Hi @coding-famer thanks so much for the PR. Tbh I didn't quite understand the PR, so left it idle at the time. Sorry! @hamishivi and I recently just merged #595, which should close this PR.

Thanks again!

PR #570, opened 2025-02-18
Small fix to tokenizer logic

hamishivi:
1. Qwen 1.5B tokenizer actually reuses llama tokenizer, but with its own bos/eos/pad, which errors when using open-instruct. I'm not sure why we added non-padding tokens and then asserted they weren't added, so removed.
2. Generally, our tokenizer setup is just about adding padding tokens (since base models often don't have them), so I added a check that just doesn't run the token adding if pad token is set.

### DIFF
--- a/open_instruct/dataset_transformation.py
+++ b/open_instruct/dataset_transformation.py
@@ -352,7 +352,7 @@ def get_tokenizer_tulu_v2(tc: "TokenizerConfig"):
 
 GET_TOKENIZER_FN = {
     "get_tokenizer_simple_v1": get_tokenizer_simple_v1,
-    "get_tokenizer_tulu_v1": get_tokenizer_tulu_v1,
+    "get_tokenizer_tulu_v1": get_tokenizer_tulu_v1,  # old version, see https://github.com/allenai/open-instruct/pull/570
     "get_tokenizer_tulu_v2": get_tokenizer_tulu_v2,
 }
 
--- a/open_instruct/dataset_transformation.py
+++ b/open_instruct/dataset_transformation.py
@@ -223,40 +223,39 @@ def get_tokenizer_tulu_v1(tc: "TokenizerConfig"):
     )
     # no default pad token for llama!
     # here we add all special tokens again, because the default ones are not in the special_tokens_map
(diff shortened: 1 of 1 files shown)


vwxyzjn (changes requested):
Makes sense to me. This should be compatible with the existing olmo2 and tulu3 recipe, right?


Could you also update this to be `get_tokenizer_tulu_v1.1`? Maybe add a comment in the `TokenizerConfig` with the link to this PR so people know how the version changed.


https://github.com/allenai/open-instruct/blob/74897429b291acc2e579a888348c6f185cbdbec5/open_instruct/dataset_transformation.py#L287

https://github.com/allenai/open-instruct/blob/74897429b291acc2e579a888348c6f185cbdbec5/open_instruct/dataset_transformation.py#L299

hamishivi:
Oh, fair point - split it into a new function. This should be compatible with past recipes -- Tulu 3 and Olmo don't even use the Llama tokenizer (which is the only main logic change), and Tulu 2 only added pad token.

vwxyzjn:
Actually, is the new get tokenizer function backward-compatible? If it is, we can just remove the existing `get_tokenizer_tulu_v1` altogether.

vwxyzjn (approved):
Sounds good. Feel free to merge!

vwxyzjn:
Could you also add a logic:

`assert tokenizer.pad_token_id != tokenizer.eos_token_id`?

I found this to lead to a bunch of bugs and we should try to assert it.

hamishivi:
Added! also changed the logic to `tokenizer.pad_token_id is None or tokenizer.pad_token_id == tokenizer.eos_token_id`, so we overwrite the pad token if it is set as so. For example:
```python
tokenizer = AutoTokenizer.from_pretrained("deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B")
>>> tokenizer.pad_token_id == tokenizer.eos_token_id
True
>>> num_added_tokens = tokenizer.add_special_tokens(
...                 {
...                     "pad_token": "<pad>",
...                 }
...             )
>>> tokenizer.pad_token_id == tokenizer.eos_token_id
False
```

PR #575, opened 2025-02-20
Support `--vllm_num_engines` > 1

vwxyzjn:
About 2x faster. With qwen 1.5B model using `--vllm_num_engines 4`, before:

```
🔥🔥🔥 Generation time: 101.81 seconds
```

after:

```
🔥🔥🔥 Generation time: 48.85 seconds
```

### DIFF
--- a/open_instruct/grpo_vllm_thread_ray_gtrl.py
+++ b/open_instruct/grpo_vllm_thread_ray_gtrl.py
@@ -859,6 +859,14 @@ class PolicyTrainerRayProcess(RayProcess):
             n=args.number_samples_per_prompt,
             stop=args.stop_strings,
         )
+        evaluation_generation_config = SamplingParams(
+            temperature=0.001,
+            top_p=1.0,
+            max_tokens=args.response_length,
+            include_stop_str_in_output=True,
+            n=1,  # since we are doing greedy sampling, don't need to generate more
+            stop=args.stop_strings,
+        )
         # print("setup async queues")
         param_prompt_Q = None
         response_ids_Q = None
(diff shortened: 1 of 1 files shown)


open_instruct/grpo_vllm_thread_ray_gtrl.py
```diff
@@ -880,34 +881,40 @@ def vllm_generate(
...
-                    outputs = ray.get(
-                        llm.generate.remote(
-                            prompt_token_ids=sample_evaluation_prompt_token_ids,
-                            sampling_params=generation_config,
-                            use_tqdm=False,
-                        )
-                    )
-                    # for evaluation, even if we have multiple outputs, we only look at one of them for simplicity
-                    response_ids = [list(output.outputs[0].token_ids) for output in outputs]
+                    response_ids = generate_with_engines(sample_evaluation_prompt_token_ids, generation_config)
```
hamishivi: I think this means now we get all samples at eval time if we set number_samples_per_prompt > 1. Does this affect aour logging?

hamishivi (approved):

open_instruct/grpo_vllm_thread_ray_gtrl.py
vwxyzjn: Good point. 

I added a evaluation generation config with `n=1`, so it should resolve it. 

It should also solve the following error, because `tokenizer.batch_decode(sample_evaluation_prompt_token_ids)` should expect to see similar things`.

image

vwxyzjn:
image

Is about 2x faster

PR #577, opened 2025-02-24
Online RL Optimization

vwxyzjn:
This PR optimizes online RL algorithm speed.

The main change is to remove the following forward. We don't need it as we can use the logprob during the actual forward. For why we can do that, see https://costa.sh/blog-understanding-why-there-isn't-a-log-probability-in-trpo-and-ppo's-objective

https://github.com/allenai/open-instruct/blob/5b5810c57df92046cf9d3b11d42611e2a7cdbabf/open_instruct/ppo_vllm_thread_ray_gtrl.py#L1137-L1139

There is also a minor optimization not to do the unnecessary `broadcast_to_vllm()` during the first iteration 

image


open_instruct/grpo_vllm_thread_ray_gtrl.py
```diff
@@ -954,14 +987,13 @@ def generate_with_engines(prompts: List[List[int]], sampling_params: SamplingPar
         )
         stats_shape = (
             args.num_epochs,
-            args.num_mini_batches * args.number_samples_per_prompt,
+            args.num_mini_batches,
```
vwxyzjn, line 990: This fixes 

image


to 

image

open_instruct/grpo_vllm_thread_ray_gtrl.py
```diff
@@ -79,6 +79,8 @@
 )
 from transformers.integrations import HfDeepSpeedConfig
 from vllm import SamplingParams
+from peft import PeftModel, get_peft_model_state_dict
```
hamishivi, line 82: oh re-adding peft model support? did you test this?

open_instruct/grpo_vllm_thread_ray_gtrl.py
```diff
@@ -634,8 +659,10 @@ def from_pretrained(
             attn_implementation="flash_attention_2",
             use_cache=False,
         )
+        self.policy_vocab_size = self.policy.config.vocab_size
         disable_dropout_in_model(self.policy)
         self.policy.gradient_checkpointing_enable()
+        # from deepspeed.ops.adam import DeepSpeedCPUAdam, FusedAdam
```
hamishivi, line 665: prolly remove?

open_instruct/grpo_vllm_thread_ray_gtrl.py
```diff
@@ -985,7 +1017,7 @@ def generate_with_engines(prompts: List[List[int]], sampling_params: SamplingPar
         queries_next = data[INPUT_IDS_PROMPT_KEY].to(device)
         ground_truths_next = data[GROUND_TRUTHS_KEY]
         datasets_next = data[DATASET_SOURCE_KEY]
-        if accelerator.is_main_process:
+        if self.rank == 0:
```
hamishivi, line 1020: is there a reason for this over accelerator main process?

open_instruct/ppo2.py
hamishivi, line 1: Accidentally committed?

open_instruct/ppo_vllm_thread_ray_gtrl.py
```diff
@@ -908,6 +943,14 @@ def repeat_generator():
             n=args.number_samples_per_prompt,
             stop=args.stop_strings,
         )
+        evaluation_generation_config = SamplingParams(
+            temperature=0.001,
```
hamishivi: why do this?

open_instruct/grpo_vllm_thread_ray_gtrl.py
vwxyzjn: No did not test it, but this seems standard in openrlhf / open reaosner codebase so I just added it back.

open_instruct/grpo_vllm_thread_ray_gtrl.py
vwxyzjn: I am inclined to keep it in case we need to comment it out and test it out something (we wouldn't know where FusedAdam comes from)

open_instruct/grpo_vllm_thread_ray_gtrl.py
vwxyzjn: The `accelerator.is_main_process` is kind of a hack. Since we now extensively using deepspeed stack, I wanted to clean it up a bit.

open_instruct/ppo2.py
vwxyzjn: No this is intentional. ppo2.py applies the KL loss instead a KL penalty in the rewards.

open_instruct/ppo_vllm_thread_ray_gtrl.py
vwxyzjn: For reproducibility and better comparison? If we are using e.g., 0.7 temp, then even the same model would generate different resposnes.

hamishivi: oh, but why 0.001 and not just greedy decode?

open_instruct/grpo_vllm_thread_ray_gtrl.py
hamishivi: sure, nw

hamishivi (approved):

open_instruct/ppo_vllm_thread_ray_gtrl.py
vwxyzjn: Good point. I thought we couldn't specify zero but the docs indicate otherwise. Have just changed it.

image

billxbf:
It's a better implementation than TRL :) Great job! @vwxyzjn

natolambert:
@vwxyzjn is a GOAT

PR #594, opened 2025-03-06
Refactor reward/verifier setup

hamishivi:
Bit of a revamp of the way the `apply_verifiable_rewards` func works:
- We now allow a list of ground truths and answers, instead of just 1.
- We add rewards together in this case
- All verifiers inherit from a basic class that defines the core API.
- We return per-verifier scores along with the added total score, and log out verification scores / rates separately. Note that some verifiers return a continuous value (e.g. the max len sample), so the rate is just counting the times the reward is non-zero.

edit: seems like there is still some bug in logging the rate correctly.... will work it out.

@vwxyzjn, let me know if you want me to make this PR into main, or merge into GRPO-fast and also edit that code for you. I don't mind either way. I need this code for my token control experiments, though!


vwxyzjn:
Looks good. Let's merge this to grpo packing branch probably.

hamishivi:
Closing to re-target packing branch.

PR #595, opened 2025-03-06
Minor fixes to mason + caching

hamishivi:
1. Add the ability to manually specify extra secrets for `mason.py`
2. respect the split argument for datasets. The old code split if you used a non-train split for the eval dataset. This fixes it by instead constructing a dataset dict following the specified splits in the dataset config.

### DIFF
--- a/open_instruct/dataset_transformation.py
+++ b/open_instruct/dataset_transformation.py
@@ -798,28 +798,28 @@ class DatasetTransformationCache:
         config_hash = self.compute_config_hash(dcs, tc)
         repo_name = f"{self.hf_entity}/dataset-mix-cached"
 
+        # NOTE: the cached dataset is always train split
+        DEFAULT_SPLIT_FOR_CACHED_DATASET = "train"
+
         # Check if the revision exists
         if revision_exists(repo_name, config_hash, repo_type="dataset"):
             print(f"✅ Found cached dataset at https://huggingface.co/datasets/{repo_name}/tree/{config_hash}")
             # Use the split from the first dataset config as default
-            return load_dataset(repo_name, split=dcs[0].dataset_split, revision=config_hash)
+            return load_dataset(repo_name, split=DEFAULT_SPLIT_FOR_CACHED_DATASET, revision=config_hash)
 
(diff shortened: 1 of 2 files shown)


open_instruct/dataset_transformation.py
```diff
@@ -807,16 +807,19 @@ def load_or_transform_dataset(self, dcs: List[DatasetConfig], tc: TokenizerConfi
...
-        transformed_datasets = []
+        transformed_datasets = {}
         for dc in dcs:
             dataset = get_dataset_v1(dc, tc)
-            transformed_datasets.append(dataset)
+            split = dc.dataset_split
+            if split in transformed_datasets:
+                transformed_datasets[split] = concatenate_datasets([transformed_datasets[split], dataset])
+            else:
+                transformed_datasets[split] = dataset
```
vwxyzjn: Here is a use case, say I want to mix "train, test, train, test" splits of 4 datasets. Would this implementation only load `"train, train"` of the first and thrid dataset? 



What do you think about making a default split like

https://github.com/allenai/open-instruct/blob/7d1c5d95cf59941283dd53f1c54c7f52452f6784/open_instruct/dataset_transformation.py#L801-L808

and 

https://github.com/allenai/open-instruct/blob/7d1c5d95cf59941283dd53f1c54c7f52452f6784/open_instruct/dataset_transformation.py#L858-L859

vwxyzjn (commented):
Nice addition on the `secret`. The `dataset_transformation.py` might have some issues.

open_instruct/dataset_transformation.py
hamishivi: Ah, I guess i was thinking of this differently, this setup makes more sense.

hamishivi: Okay changed to this, and tested for my use case and it works!

vwxyzjn: Did you push the commit?

hamishivi: lol im dumb

vwxyzjn (approved):

PR #596, opened 2025-03-06
Refactor reward setup + mult rewards at once

hamishivi:
Bit of a revamp of the way the apply_verifiable_rewards func works:

- We now allow a list of ground truths and answers, instead of just 1.
- We add rewards together in this case
- All verifiers inherit from a basic class that defines the core API.
- We return per-verifier scores along with the added total score, and log out verification scores / rates separately. Note that some verifiers return a continuous value (e.g. the max len sample), so the rate is just counting the times the reward is non-zero.

From testing, appears to log correctly. This shouldn't change the actual reward computation from before.

I merged this branch into main and used a linter so it added a bunch of other changes too...

Merging into grpo-fast since it needs to integrate there. Need to debug+test some logging stuff before merging.


open_instruct/grpo_fast.py
```diff
@@ -1597,7 +1632,20 @@ def reward_fn(decoded_responses: List[str], ground_truths: List[str], datasets:
                 np_verifiable_rewards = np.array(verifiable_rewards)
                 metrics["objective/verifiable_reward"] = np_verifiable_rewards.mean()
                 metrics["objective/verifiable_correct_rate"] = (np_verifiable_rewards > 0.0).mean()
-
+                # reshuffle around per_func rewards
+                per_func_lists = {}
+                for reward_dict in per_func_rewards:
+                    for key, value in reward_dict.items():
+                        if key not in per_func_lists:
+                            per_func_lists[key] = []
```
vwxyzjn:
```suggestion
                per_func_lists = defaultdict(list)
                for reward_dict in per_func_rewards:
                    for key, value in reward_dict.items():
```

open_instruct/grpo_fast.py
```diff
@@ -1621,8 +1669,8 @@ def reward_fn(decoded_responses: List[str], ground_truths: List[str], datasets:
                         if float(answer) == float(ground_truths[i]):
                             arithmetic_rewards.append(args.arithmetic_reward)
                             scores[i] += args.arithmetic_reward
-                    except:
-                        pass # it's ok if things went wrong
+                    except:  # noqa
+                        pass  # it's ok if things went wrong
```
vwxyzjn: Could you fix the logic here? It should be

```
                    try:
                        answer = decoded_response[answer_start:answer_end]
                        answer = answer.replace(",", "").strip()
                        if float(answer) == float(ground_truths[i]):
                            arithmetic_rewards.append(args.arithmetic_reward)
                            scores[i] += args.arithmetic_reward
                    except:
                        arithmetic_rewards.append(0)
                        pass # it's ok if things went wrong
```

vwxyzjn (approved):
Generally looks good to me. Some minor comments. Could you fix the PPO and merge?

PR #603, opened 2025-03-11
delete default model mount when not using beaker models for safety eval

saumyamalik:
tested, dataset inputs on jobs look good


hamishivi (approved):
LGTM, Thanks!

PR #608, opened 2025-03-14
Make mason standalone

vwxyzjn:
@hamishivi is this what you had in mind?


hamishivi (approved):
LGTM! Yes, roughly this is what I had in mind.

PR #610, opened 2025-03-14
add arg for eval_priority, default normal

saumyamalik:
I haven't tested it yet...but this should add and pass an arg `eval_priority` in each of 4 training scripts where some form (either from `utils.py` or locally redefined) of `launch_ai2_evals_on_weka` is called

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -281,6 +281,8 @@ class Args:
     """The beaker evaluation tasks to launch"""
     oe_eval_max_length: int = 4096
     """the max generation length for evaluation for oe-eval"""
+    eval_priority: str = "normal"
+    """the priority of auto-launched evaluation jobs"""
 
     def __post_init__(self):
         if self.single_gpu_mode:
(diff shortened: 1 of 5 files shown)


open_instruct/grpo_fast.py
```diff
@@ -281,6 +281,8 @@ class Args:
     """The beaker evaluation tasks to launch"""
     oe_eval_max_length: int = 4096
     """the max generation length for evaluation for oe-eval"""
+    eval_priority: str = "normal"
```
hamishivi: Can you add a `__post_init__` check that these are valid? or `Literal` type like KL. So we don't accidentally put in a typod priority and dont realise until the first intermediate eval.

saumyamalik: good idea, will push, test, and update

vwxyzjn: looks good!

vwxyzjn (approved):

vwxyzjn:
LGTM! Merging as is.

PR #612, opened 2025-03-18
fix epoch logging in grpo_fast

hamishivi:
GRPO fast epoch logging doesn't take into account that we are sampling multiple times from a prompt in each step (and counting that to the episode count). This fixes.

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -1504,7 +1504,7 @@ def main(args: Args, model_config: ModelConfig, reward_fn: Callable):
                     "episode": episode,
                     "training_step": training_step,
                     "val/num_total_tokens": num_total_tokens,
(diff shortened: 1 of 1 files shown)


open_instruct/grpo_fast.py
```diff
@@ -1504,7 +1504,7 @@ def main(args: Args, model_config: ModelConfig, reward_fn: Callable):
                     "episode": episode,
                     "training_step": training_step,
                     "val/num_total_tokens": num_total_tokens,
-                    "epoch": episode / len(train_dataset),
+                    "epoch": episode / len(train_dataset) / args.num_samples_per_prompt_rollout,
```
vwxyzjn: How about

```suggestion
                    "epoch": episode / args.num_samples_per_prompt_rollout / len(train_dataset),
```

PR #614, opened 2025-03-18
Fix reward logging for open_instruct/grpo_vllm_thread_ray_gtrl.py

hamishivi:
reward logging here before was mucked up, since different ranks could have different sets of rewards.
This fixes by making sure there is always some value for each reward type in the dataset logged on each rank. If the reward wasn't seen, this value ends up being NaN, and we ignore it when reducing metrics. We only apply this logic to the rewards, meaning e.g. if loss becomes NaN it will still log as normal.

Testing out here: https://wandb.ai/ai2-llm/open_instruct_internal/runs/4bqxl8io?nw=nwuserhamishivi


hamishivi:
Yeah, open to ways to make this better, but at least nice to have it correct for now.

PR #617, opened 2025-03-19
mask all turns but last for dpo

VictoriaGraf:


hamishivi (approved):
Looks correct to me (and assuming its tested). I sort of wonder if the logic couldn't be simpler? Like can't we just tokenize messages[:-1] and then mask [0, len(tokenized messages]? But this isn't a blocker to merging.

open_instruct/dataset_transformation.py
```diff
@@ -548,10 +615,10 @@ def preference_tulu_tokenize_and_truncate_v1(
     if len(rejected_messages) == 0:
         raise ValueError("rejected messages field is empty.")
 
-    chosen_encoded = sft_tulu_tokenize_and_truncate_v1(
+    chosen_encoded = last_turn_tulu_tokenize_and_truncate_v1(
```
vwxyzjn, line 999: Let's make a `sft_tulu_filter_v1_2` for this. The motivation of versioning the functions is to make sure they are reproducible. Considering this appears to be performance impacting refactoring, let's give it a different name.

natolambert:
Okay in discussions, I agree with above, would be good to have @VictoriaGraf explain the context of her testing (I think she retrained DPO models and didnt see degradation), but should be good to merge.

o1pro code review also happy with it, the bigger picture is how to manage these things for reproducibility.

PR #624, opened 2025-03-25
Add host networking option to mason.py

hamishivi:
As title says. Useful for cases where we want to launch something with mason (e.g., a server of some sort) and make it accessible to other jobs.


hamishivi:
@vwxyzjn Flipped the default!

vwxyzjn (approved):

PR #629, opened 2025-03-26
Vllm upgrade 2

vwxyzjn:
Faster and no memory leak.

image

Also cutdown the size of dockerfiles (9GB) by removing `RUN chmod -R 777 /stage/` 

image


vwxyzjn (commented):
All tested. @hamishivi also fixed an issue with PPO reward logging.

image

PR #635, opened 2025-03-31
Add more configs

vwxyzjn:
Mostly just adding configs here and there and update the results in the docs.


vwxyzjn:
Everything tested. Merging as is.

PR #642, opened 2025-04-08
Grpo log the real batch size and allow update of ref model

vwxyzjn:
In GRPO, as we train we actually get smaller and smaller batch sizes. This is because if we solve a prompt 100% correct or 0% correct, the std of the group is 0. So `adv = (score - score.mean()) / (score.std + 1e-5) = 0 / 1e-5 = 0`, causing 0 gradients.

This PR

1. logs the real batch size ratio: the percentage of the batch that produces zero gradient
2. filter out those zero gradient batches before packing.

image

It also implements a new thing: allow ref model to update


open_instruct/grpo_fast.py
```diff
@@ -214,6 +214,13 @@ class Args:
     """the length of the pack (you should prob set to the max length of the model)"""
     masked_mean_axis: Optional[int] = None
     """the axis to compute the mean of the masked values"""
+    alpha: float = 0.6
+    """The alpha value for doing polyak updates (ref_param = alpha * param + (1 - alpha) * ref_param)
+    reference: [TR-DPO](https://huggingface.co/papers/2404.09656), but it's actually pretty commonly
+    used. E.g., [TD3](https://arxiv.org/abs/1802.09477) uses https://github.com/vwxyzjn/cleanrl/blob/dcc289fc6f0bda492fa7360a155262cf826b12a5/cleanrl/td3_continuous_action.py#L269
+    """
+    ref_policy_update_freq: Optional[int] = None
+    """The frequency to update the reference policy"""
```
hamishivi:
```suggestion
    """How many training steps to take before updating the reference policy."""
```

hamishivi (approved):
Looks good to me, assuming everything is tested!

vwxyzjn:
Yep tested. The following refreshes ref model every 10 steps:

deepspeed stage 2

image

deepspeed stage 3

image

PR #643, opened 2025-04-08
grpo_fast correct eval freq and get eval metrics

mnoukhov:
currently `grpo_fast` 
1. checks for eval results before training (and finds none)
2. then the generation thread will generate eval results after first train results
3. then it may log the eval generation from the 1st step during the 2nd training step

as well
- you do not do a final evaluation after the last step
- there are no metrics saved from the evaluation generations

This PR
- adds evaluation metrics
- switches to a more standard evaluation scheme
  - eval every `eval_freq` so that you also do an evaluation after the last step 
  - i.e. total steps = 10, num_evals = 2 will evaluate after the 5th step and after the 10th (final) step
  - total_steps = 32, num_evals = 4 will evaluate after the 8th, 16th, 24th, and 32nd step

If getting the eval generations from the first step is important, I can add an arg for that

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -1061,7 +1061,7 @@ def vllm_generate_thread(
         inference_results_Q.put((response_ids, finish_reasons))
 
         # Evaluate the model
-        if eval_prompt_token_ids is not None and (training_step - 1) % eval_freq == 0:
+        if eval_prompt_token_ids is not None and training_step % eval_freq == 0:
             response_ids, finish_reasons = generate_with_engines(eval_prompt_token_ids, eval_generation_config)
             evaluation_inference_results_Q.put((response_ids, finish_reasons))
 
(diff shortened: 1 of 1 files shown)


open_instruct/grpo_fast.py
```diff
@@ -1061,7 +1061,7 @@ def generate_with_engines(prompts: List[List[int]], sampling_params: SamplingPar
         inference_results_Q.put((response_ids, finish_reasons))
 
         # Evaluate the model
-        if eval_prompt_token_ids is not None and (training_step - 1) % eval_freq == 0:
+        if eval_prompt_token_ids is not None and training_step % eval_freq == 0:
```
vwxyzjn, line 1064: This seems problematic. Then the initial policy's response would not be logged. Could you revert the change?

vwxyzjn:
Thanks @mnoukhov ! I combined your changes and merged in https://github.com/allenai/open-instruct/pull/651.

PR #651, opened 2025-04-15
enable training from resumption

vwxyzjn:
This PR enables grpo training to resume from checkpoints. It works like this:

### auto-resume:

If the run was preemptied, it could auto-resume. If the run errored out from nccl error, we could "restart" the same beaker job, and the training would continue.

So for the following job, I tried cancel the job and hit the refresh button, and it was able to resume training pretty nicely. We can reasonably assume that resuming from preemption is gonna work nicely, too.

image

### arbitrary resume: we allow the user to specify a `--checkpoint_state_dir` or `--gs_checkpoint_state_dir` to start from an arbitrary checkpoint

```
python update_command_args.py scripts/train/qwen/grpo_fast_7b.sh \
    --cluster ai2/augusta-google-1 \
    --workspace ai2/scaling-rl \
    --image costah/open_instruct_dev_0415_9 \
    --wandb_project_name open_instruct_public \
    --checkpoint_state_dir /weka/oe-adapt-default/allennlp/deletable_checkpoint_states/costah/1744804711_477946 \
    --gs_checkpoint_state_dir gs://ai2-llm/post-training//costah//weka/oe-adapt-default/allennlp/deletable_checkpoint_states/costah/1744804711_477946 \
    --checkpoint_state_freq 200 \
    --priority high | uv run bash
```

image

### Edge cases and limitations:

An edge case was that the upload could fail. For example, imagine the upload or saving failed somehow. In that case, my implementation would try to locate the latest checkpoint folder that "has the most files" as the intact checkpoint to resume.

One limitation is that if preemptied, we do not have that sigterm handling which saves the model on preemption, but we do save the checkpoints periodically so this is prob enough for now.

### Evaluation:

For eval, you do need to gather all the separate run_ids from wandb, but you can grab them by doing a filtering step in wandb

image

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -1498,15 +1498,15 @@ def main(args: Args, tc: TokenizerConfig, model_config: ModelConfig, reward_fn: 
         pg=pg if args.single_gpu_mode else None,
     )
     resume_training_step = ray.get(inits)[0] + 1
-    if resume_training_step > 1:
-        print(f"Resuming training from step {resume_training_step}... Broadcasting weights to vLLM engines.")
-        with Timer("[Main Thread] 🔄 Loading weights using shared memory"):
-            ray.get([m.broadcast_to_vllm.remote() for m in policy_group.models])
     episode = (resume_training_step - 1) * args.num_unique_prompts_rollout * args.num_samples_per_prompt_rollout
     print("======== ✅ all models and vLLM engines initialized =========")
 
     ray.get([m.setup_model_update_group.remote(vllm_engines=vllm_engines) for m in policy_group.models])
     print("======== ✅ model update group setup successfully =========")
+    if resume_training_step > 1:
+        print(f"Resuming training from step {resume_training_step}... Broadcasting weights to vLLM engines.")
+        with Timer("[Main Thread] 🔄 Loading weights using shared memory"):
+            ray.get([m.broadcast_to_vllm.remote() for m in policy_group.models])
 
     # Setup training
     generation_config = SamplingParams(
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -1656,7 +1656,7 @@ def main(args: Args, tc: TokenizerConfig, model_config: ModelConfig, reward_fn: 
                 }
                 scalar_metrics = {}
                 for key, value in metrics.items():
-                    if isinstance(value, float):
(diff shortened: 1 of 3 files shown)


open_instruct/grpo_fast.py
```diff
@@ -1360,7 +1453,7 @@ def main(args: Args, tc: TokenizerConfig, model_config: ModelConfig, reward_fn:
             dataset_local_cache_dir=args.dataset_local_cache_dir,
             dataset_skip_cache=args.dataset_skip_cache,
         )
-        eval_dataset = eval_dataset.shuffle(seed=args.seed)
+        # NOTE: eval dataset should not be shuffled
```
hamishivi: Could we add a flag for this? Sometimes its nice to shuffle the eval dataset e.g. if the dataset is ordered in some way, and we want to observe performance on a random subset (to avoid large inference eval costs during training).

open_instruct/grpo_fast.py
```diff
@@ -1593,32 +1672,90 @@ def main(args: Args, tc: TokenizerConfig, model_config: ModelConfig, reward_fn:
...
+                eval_scores, eval_reward_metrics = reward_fn(
+                    eval_responses,
+                    eval_decoded_responses,
+                    eval_ground_truths,
+                    eval_dataset_names,
+                    eval_finish_reasons,
+                )
+                eval_reward_metrics = {f"eval/{key}": val for key, val in eval_reward_metrics.items()}
+                eval_metrics = {
+                    "eval/scores": np.array(eval_scores).mean(),
```
hamishivi, line 1732: nice!

hamishivi (approved):
LGTM, only one minor comment. Does this work on augusta and jupiter?

open_instruct/grpo_fast.py
vwxyzjn: This is actually @mnoukhov's code in #643. It was a bit more convenient to include his change in this PR. Thanks @mnoukhov!

open_instruct/grpo_fast.py
vwxyzjn: Good point. Just added a flag.

vwxyzjn:
> Does this work on augusta and jupiter?

Haven't tried jupiter but it works on Augusta. e.g., on a 4 node setup:


image


The drop was caused by not broadcasting the policy weights to vllm engine upon resumption. I added that logic as well, and the KL drop disappeared. 

image

PR #654, opened 2025-04-21
vLLM offline inference with tools

vwxyzjn:
Quick prototype. So far it seems to work nicely: GPU utilization is high and the output looks reasonable.

https://github.com/user-attachments/assets/7cd9df0c-3b2a-4458-9351-9c0175bf5bd0


open_instruct/tool_utils/tool_vllm.py
```diff
@@ -0,0 +1,304 @@
...
+
+    def __call__(self, prompt: str) -> ToolOutput:
+        raise NotImplementedError("Subclasses must implement this method")
+
+class PythonCodeTool(Tool):
+    """@vwxyzjn: I recommend using something like a FastAPI for this kind of stuff; 1) you 
+    won't accidentally block the main vLLM process and 2) way easier to parallelize via load balancing."""
+    def __call__(self, prompt: str) -> ToolOutput:
+        # Find Python code blocks using regex
+        re_str = r'<tool>\s*(.*?)\s*</tool>'
```
ljvmiranda921: Wondering if we can parametrize some of the special tokens to ascribe w/ the current template: https://github.com/allenai/open-instruct/blob/toolu/open_instruct/dataset_transformation.py#L251

there might be a chance the template might change though!

open_instruct/tool_utils/tool_vllm.py
```diff
@@ -0,0 +1,304 @@
...
+        # Return all captured outputs as a single string
+        return ToolOutput(output='\n'.join(all_outputs), called=True, success=True)
+
+
+
+class ToolUseLLM(LLM):
+    def __init__(self, tools: dict[str, Tool] = None, sampling_params: SamplingParams = None, *args, **kwargs):
+        self.tools = tools
+        self.sampling_params = sampling_params
+        assert sampling_params.n == 1, "ToolUseLLM only supports n=1, because we could potentially do multiple too calls"
```
hamishivi: Doesn't this make this not really usable for GRPO? One thing you can do is just flatten the output and duplicate the prompts, and then re-construct things back at the output... this is what I did for the rl-rag stuff.

open_instruct/tool_utils/tool_vllm.py
hamishivi: +1 on making the <tool> tag configurable.

open_instruct/tool_utils/tool_vllm.py
```diff
@@ -0,0 +1,304 @@
...
+        
+        if len(code_blocks) == 0:
+            return ToolOutput(output="", called=False, success=False)
+
+        # Only execute the last code block
+        code = code_blocks[-1]
+        try:
+            # Call the FastAPI endpoint to execute the code
+            response = requests.post(
+                "http://phobos-cs-aus-453:1212/execute",
```
hamishivi: Probably best to set via an env var or argument.

hamishivi:
Another higher level thought was that at least for RL-rag I'm currently controlling the number of 'actions' the model is allowed to take via the higher-level generation loop. It might be useful to be able to do that here (e.g. if you only have budget for X calls per query).

open_instruct/tool_utils/tool_vllm.py
vwxyzjn: The tag is actually already configurable. See the immediate line below:

`re_str = re_str.replace('<tool>', self.start_str).replace('</tool>', self.end_str)`

I added a comment there: `# we replace <tool> immediately with the custom start_str`. Let me know if this resolves your use case.

open_instruct/tool_utils/tool_vllm.py
vwxyzjn: Good point. Have addressed this.

open_instruct/tool_utils/tool_vllm.py
vwxyzjn: This is a good point. Though we can still make it work with GRPO via repeated prompt outside of this, it's a bit easier to support `sampling_param.n`. I have modified the code to support it.

vwxyzjn (commented):
Should have addressed all comments. Let me know what you all think!

hamishivi:
I took a stab at integrating this into my RL-Rag code (and so into GRPO) here: https://github.com/allenai/open-instruct/tree/rl-rag-tool/open_instruct

PR #655, opened 2025-04-21
Deepspeed import on cpu fix

vwxyzjn:
Otherwise when launching mason we get 

```
  File "/weka/oe-adapt-default/costah/oi2/.venv/lib/python3.10/site-packages/triton/runtime/driver.py", line 8, in _create_driver
    raise RuntimeError(f"{len(actives)} active drivers ({actives}). There should only be one.")
RuntimeError: 0 active drivers ([]). There should only be one.
```


natolambert (approved):
thanks!

PR #660, opened 2025-04-22
fix: allow not randomizing the output directory

HarikrishnanBalagopal:


vwxyzjn:
Thanks for the PR. What do you think about adding a `checkpoint_state_dir` like done in `grpo_fast.py`?

https://github.com/allenai/open-instruct/blob/b5f60321d8186500d10660e3d7b575ae619a678b/open_instruct/grpo_fast.py#L303-L306

HarikrishnanBalagopal:
> Thanks for the PR. What do you think about adding a `checkpoint_state_dir` like done in `grpo_fast.py`?
> 
> https://github.com/allenai/open-instruct/blob/b5f60321d8186500d10660e3d7b575ae619a678b/open_instruct/grpo_fast.py#L303-L306

Thanks for looking at the PR.

I might be mistaken but the `checkpoint_state_dir` flag looks complicated
https://github.com/allenai/open-instruct/blob/b5f60321d8186500d10660e3d7b575ae619a678b/open_instruct/grpo_fast.py#L345
It involves heuristics based on file counts to guess valid checkpoints and removing invalid checkpoints https://github.com/allenai/open-instruct/blob/102562db4ebc515d2c1cee028a0bc4eac7df39e0/open_instruct/utils.py#L770-L776

The goal of this PR is simply to make the output directory deterministic. This way we can resume a failed run with the exact same config (including `output_dir`) and it will find and resume from the last checkpoint that was written to the directory.

Additionally it's also useful when used as part of a pipeline where you might want full control over the exact output directory path.

hamishivi (approved):
LGTM, thanks! Keeping this optional works.

PR #661, opened 2025-04-24
Fix final eval + other bugs

hamishivi:
- We had an assertion that GRPO requires > 1 sample per prompt. This is true, but running 1 sample per prompt can be interesting to try (basically just REINFORCE), so I removed the assert and added a log warning.
- #651 added some changes where we always eval on the final step, but the generation thread was not updated. This meant that the code would hang waiting for an evaluation generation thread to finish that would never come. This PR should fix that.
- Sometimes `try_load_from_cache` can return `_CACHED_NO_EXIST`, an empty object (see https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/file_download.py#L1380-L1381). This fixes our code to handle that return value.
- We had liger kernel stuff in SFT, but its fallen out. Added it back into the default install things.

### DIFF
--- a/open_instruct/dataset_transformation.py
+++ b/open_instruct/dataset_transformation.py
@@ -66,6 +66,7 @@ from transformers import (
 from transformers.utils.hub import (
     _CACHED_NO_EXIST,
     TRANSFORMERS_CACHE,
+    _CACHED_NO_EXIST,
     extract_commit_hash,
     try_to_load_from_cache,
 )
@@ -86,9 +87,13 @@ def custom_cached_file(model_name_or_path: str, filename: str, revision: str = N
         else:
             return None
     else:
-        return try_to_load_from_cache(
+        resolved_file = try_to_load_from_cache(
             model_name_or_path, filename, cache_dir=TRANSFORMERS_CACHE, revision=revision, repo_type=repo_type
         )
+        # special return value from try_to_load_from_cache
+        if resolved_file == _CACHED_NO_EXIST:
+            return None
+        return resolved_file
 
 
 def get_commit_hash(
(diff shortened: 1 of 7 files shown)


open_instruct/dataset_transformation.py
```diff
@@ -66,6 +66,7 @@
 from transformers.utils.hub import (
     _CACHED_NO_EXIST,
     TRANSFORMERS_CACHE,
+    _CACHED_NO_EXIST,
```
vwxyzjn: There seems to be duplicate import here?

open_instruct/grpo_fast.py
```diff
@@ -1152,7 +1155,11 @@ def generate_with_engines(prompts: List[List[int]], sampling_params: SamplingPar
         inference_results_Q.put((response_ids, finish_reasons))
 
         # Evaluate the model
-        if eval_prompt_token_ids is not None and (training_step - 1) % eval_freq == 0:
+        if (
+            eval_prompt_token_ids is not None
+            and num_evals > 0
+            and ((training_step - 1) % eval_freq == 0 or training_step == num_training_steps)
+        ):
```
vwxyzjn, line 1162: One thing that may also help is to set `evaluation_inference_results_Q = Queue(maxsize=10)` and also a timeout = 100 seconds in the final check (instead of a None that could block forever). Was gonna add just to be on the safe side, but what you have here should also work!

vwxyzjn (approved):
LGTM!

vwxyzjn (approved):

open_instruct/grpo_fast.py
hamishivi: Oh yeah, we should add some timeout anyway, good point.

PR #670, opened 2025-05-11
Tool + RL-rag

hamishivi:
Merging #663 and some rl-rag stuff. 

Also adds prefix cache invalidation for when we update weights, as this wasn't happening before. This meant when doing RL training with a long static prompt (that would not get invalidated), training would result in a worse model when loaded separately, since it was always using the old kv-cache.


hamishivi:
@vwxyzjn would be great to get your feedback on this at some point!

open_instruct/vllm_utils2.py
```diff
@@ -138,9 +138,15 @@ def __init__(self, *args, bundle_indices: list = None, **kwargs):
             os.environ["VLLM_RAY_BUNDLE_INDICES"] = ",".join(map(str, bundle_indices))
             print(f"creating LLM with bundle_indices={bundle_indices}")
 
-        from vllm import LLM
+        if tool_use:
+            # use tool vllm
+            from open_instruct.tool_utils.tool_vllm import ToolUseLLM
```
vwxyzjn: Nice change! Did you end up merging `vllm_utils2.py` and `vllm_utils3.py`?

open_instruct/grpo_fast.py
```diff
@@ -323,6 +325,25 @@ class Args:
     eval_priority: Literal["low", "normal", "high", "urgent"] = "normal"
     """the priority of auto-launched evaluation jobs"""
 
+    # Tool settings
+    tools: Optional[List[str]] = None
+    """If set, use the tool mapped to the string. Currently only supports `search` and `code`"""
+    max_tool_calls: int = 5
```
vwxyzjn: I guess one thing we could do to improve `max_tool_calls` is `max_tool_calls: List[int]`, so that it 1) either has len = 1, in which case, the same `max_tool_calls` applies to all the tools, or 2) it has the same length as tools, in which case each tool gets its own max_tool_calls.

vwxyzjn (approved):
LGTM!

open_instruct/vllm_utils2.py
hamishivi: Ah, I didn't, I should...

open_instruct/grpo_fast.py
hamishivi: done!

open_instruct/vllm_utils2.py
hamishivi: done!

hamishivi:
Comments addressed, running some quick tests to make sure I haven't broken anything in addressing them, and then I'll merge.

open_instruct/grpo_fast.py
```diff
@@ -113,7 +115,7 @@
     maybe_use_ai2_wandb_entity,
     sync_gs_bucket,
 )
-from open_instruct.vllm_utils2 import create_vllm_engines, init_process_group
+from open_instruct.vllm_utils3 import create_vllm_engines, init_process_group
```
natolambert, line 118: Hmmm, weird naming? Meant to not include?

open_instruct/grpo_fast.py
```diff
@@ -323,6 +325,25 @@ class Args:
...
+    mask_tool_use: bool = True
+    """Whether to mask the tool output. By default on."""
+    only_reward_good_outputs: bool = False
+    """Whether to only reward good outputs. By default off. Useful to force the model to use the tool(s)."""
+
+    # rl-rag specific settngs
+    number_documents_to_search: int = 3
+    """The maximum number of documents to retrieve for each query."""
+    search_api_endpoint: Optional[str] = None
+    """The API endpoint for the search engine."""
```
natolambert, line 342: I dream of these args getting super messy because we have way too many.
Maybe should start with a prefix, like this would be `tool_rag_...`

open_instruct/grpo_fast.py
```diff
@@ -347,107 +368,11 @@ def __post_init__(self):
             download_latest_checkpoint_from_gs(self.gs_checkpoint_state_dir, self.checkpoint_state_dir)
         if self.checkpoint_state_dir is not None:
             calibrate_checkpoint_state_dir(self.checkpoint_state_dir)
-
-
-def get_train_ds_config(
```
natolambert, line 352: Guessing stuff like this is just no longer used?

natolambert (approved):
My high level comment is "should the tool use be in the same file as all the other RL." 
Should we abstract it somehow?
Should it be a separate file?

I can't help but think that it would be nice if it was all handled vllm side or something. I wonder how OpenAI etc do this.

Definitely a PR so big that it is a research project. I'm going to revisit the PR to keep learning the codebase, but don't wait for me to move with it.

open_instruct/grpo_fast.py
vwxyzjn: They are just getting imported from above using a shared utility file

open_instruct/grpo_fast.py
hamishivi: I'll rename to utils2, i merged utils3 and 2 and left the 3 name

open_instruct/grpo_fast.py
hamishivi: Good idea! I'm not sure there is a great way to reduce the number of args but we can keep them organised.

open_instruct/grpo_fast.py
hamishivi: Yeah, these got pushed into a utility file.

PR #678, opened 2025-05-15
Raise error when num_processes not set for multi-node jobs.

hamishivi:
Currently if num_proceses is not set for multi-node jobs, then the regex fails silently and the job is submitted, but doesn't actually run as a multi-node job. This adds an explicit check for this (I and Ethan hit this recently).


vwxyzjn (approved):
Lgtm

PR #681, opened 2025-05-21
fix: only skip batches once on resume

garrett361:
Fixes the skip-batch logic so that we only skip `resume_step` batches once. The current logic causes `resume_step` batches to be skipped every epoch.


garrett361:
CC @natolambert , maybe? Looks like you may have touched this section of code last.

hamishivi (approved):
Ah, thanks so much for catching this :D

PR #686, opened 2025-05-22
Revert tool call out of old grpo

hamishivi:
When I did the big merge last week, I accidentally left some of my older tool call work in. This removes that.

TEMP while testing this. Might need to make some additional edits.


natolambert (approved):
This looks right, ofc I haven't tested it.

PR #687, opened 2025-05-24
Configurable async steps

hamishivi:
Adds a cmd line arg to set how many async steps we are allowed to go (simply controlled by queue size).


natolambert (approved):
Has this been tested / set up correctly? Not familiar with the data type so while the code changes look minimal i'm hoping it doesn't cause any other issues?

PR #689, opened 2025-05-26
Overlong filtering + adjustable top-p

hamishivi:
Basic support for overlong filtering and support for adjusting top-p (I found this important for tool use experiments).


natolambert (approved):
LGTM.
Will change of top_p from 0;95 to 1 matter for a default?

Also, I am like, Hamish was reading this stuff: https://x.com/kalomaze/status/1926765123219124665

hamishivi:
It might cause issues with tool use training (I saw qwen sometime give weird generations that caused vllm to error out, producing tokens not in its vocab, which you can fix by lowering top_p to avoid those tokens). But  figure we should default to 1 unless really needed.

PR #692, opened 2025-05-27
initial judge refactor

saurabh111233212:
refactor to allow for LM-judge based "verifer"

- general `VeriferConfig` so that we can configure verifers per training run (builder pattern). Not much in there now but easily extendable to any verifiers we want to configure in the future.
- verifiers now run async/in parallel by default. This should be a small perf bonus for efficient verifers (like math) but hugely important for slow verifiers (like LM as a judge and code execution)

test beaker training runs:
jupiter run [here](https://beaker.allen.ai/orgs/ai2/workspaces/oe-adapt-code/work/01JWD8K9QDY36MRSRKS3PKZ4MW?taskId=01JWD8K9QF3JZ7W6YJAJPKJ5ER&jobId=01JWD8K9XBXRK1CNPYPY5AH4JR)
augusta run [here](https://beaker.allen.ai/orgs/ai2/workspaces/oe-adapt-code/work/01JWD94HC8ACQPH97FM7AP41EE?taskId=01JWD94HCA7RAK8VDWX2DEQBD2&jobId=01JWD94HHM2HH7M5SX24Z4EWT0)


open_instruct/ground_truth_utils.py
```diff
@@ -371,33 +440,182 @@ class UpToMaxLenVerifier(VerifierFunction):
...
+                    import signal
+                    loop.add_signal_handler(signal.SIGTERM, cleanup_client)
+                except (AttributeError, NotImplementedError):
+                    pass
+            
+            return self._client_cache[loop_id]
+            
+        except RuntimeError:
+            return AsyncAzureOpenAI(
+                api_key=os.getenv("AZURE_OPENAI_API_KEY"),
```
hamishivi: might this need to be `username_KEY` instead?

open_instruct/ground_truth_utils.py
```diff
@@ -25,6 +31,40 @@
 
 logger = logging.getLogger(__name__)
 
+@dataclass
+class VerifierConfig:
```
hamishivi, line 44: Can we have a more generic superclass, and then a LlmJudgeVerifierConfig?

open_instruct/model_utils.py
```diff
@@ -230,28 +243,59 @@ def apply_verifiable_reward(
...
             dataset_list = dataset
         assert len(ground_truth_list) == len(dataset_list), "Ground truth and dataset list lengths do not match."
-        # for now, we just assume rewards are additive, rather than more complex functions.
-        reward = 0
-        per_func_reward = {}
+        
+        # Create async tasks for each ground truth/dataset pair
         for gt, ds in zip(ground_truth_list, dataset_list):
-            reward_func = REWARD_FN_MAPPING.get(ds.lower())
+            if ds.endswith('-'):
```
hamishivi: What is this check for?

open_instruct/grpo_fast.py
```diff
@@ -49,6 +49,7 @@
 import threading
```
hamishivi: Can you make the same changes in ppo_fast?

open_instruct/grpo_fast.py
```diff
@@ -250,6 +251,10 @@ class Args:
     verification_reward: float = 10.0
     """the reward value for verifiable responses"""
 
+    # -- llm verifiers reward
+    llm_judge_model: str = "gpt-4o-mini-standard"
```
hamishivi: Do we need this? Can we auto-expand the args based on the verifier configs somehow? Not sure on the best way, but it would be nice to have the verifierConfigs specify args, and fill out the arg list here automatically.

but open to better ideas...

hamishivi (commented):
Looks good, especially if tested. I think we can do some more cleanup maybe?

open_instruct/ground_truth_utils.py
saurabh111233212: nope! The change in `mason.py` reads the `username_AZURE_OPENAI_API_KEY` beaker secret and sets the env variable (without the username prefix)

open_instruct/ground_truth_utils.py
saurabh111233212: That makes sense since we're passing a config to every verifier, but are there any shared fields between verifiers? Otherwise the super class is basically just `dataclass.dataclass`

open_instruct/model_utils.py
saurabh111233212: ah, hacky thing because the dataset I was testing with has these appended to the dataset names. I'll remove from the PR

open_instruct/ground_truth_utils.py
hamishivi: I don't think there are, but it feels weird to me to have these extra fields getting passed to verifiers that care about them at all...

open_instruct/grpo_fast.py
saurabh111233212: Oh, I have it the other way around. The Args fill out the verifier config. (see `def from_args(cls, args)` in `ground_truth_utils.py`) 

this way, the verifier config can be passed to the script from the top-level just like everything else (see `scripts/train/debug/judge.sh`).  

Is that not what we want here?

open_instruct/ground_truth_utils.py
fabrahman, line 44: is this only used for llm judge verifiers and not other? (if so maybe we name it accordingly)

open_instruct/ground_truth_utils.py
```diff
@@ -34,23 +74,47 @@ class VerifierFunction(ABC):
...
+        Subclasses can override this method for truly asynchronous implementation.
+
+        Args:
+            tokenized_prediction (List[int]): Tokenized representation (unused by most verifiers).
+            prediction (str): The model output.
+            label (Any): The ground truth answer or evaluation constraint.
+            query (Optional[str]): The original query.
+
+        Returns:
+            Tuple[float, float]: A tuple containing (score, cost)
```
fabrahman: Update the doc string to VerificationResults.

open_instruct/ground_truth_utils.py
saurabh111233212: @hamishivi I agree. I'll go ahead and create that empty shared config

open_instruct/grpo_fast.py
hamishivi: I guess im thinking about if we added lots of other verifiers (e.g., maybe a code server to run tests on, or a mixture of different lm judges), and had a lot more args. But I think for now yeah lets do it this way!

scripts/train/debug/judge.sh
fabrahman, line 1: so we have to include  `llm_judge_max_tokens`, and `llm_judge_temperature` here if needed?

open_instruct/grpo_fast.py
fabrahman: So this can be removed from here if it's already in VerifierConfigs, right?

open_instruct/ground_truth_utils.py
```diff
@@ -371,33 +440,182 @@ class UpToMaxLenVerifier(VerifierFunction):
...
+    """
+    Verifier that uses a language model's judgement to score a response.
+    """
+    _client_cache = {}
+
+    def __init__(self, judge_type: str, verifier_config: VerifierConfig) -> None:
+        super().__init__(f"general-{judge_type}", verifier_config=verifier_config, weight=1.0)
+        self.prompt_template = JUDGE_PROMPT_MAP[judge_type]
+    
+    def _get_client(self):
```
fabrahman: So this still creates the client per batch, right?

fabrahman: This only handle AzureOpenA. how about other closed/open source models not deployed in azure?

open_instruct/ground_truth_utils.py
```diff
@@ -371,33 +440,182 @@ class UpToMaxLenVerifier(VerifierFunction):
...
+
+class LMJudgeVerifier(VerifierFunction):
+    """
+    Verifier that uses a language model's judgement to score a response.
+    """
+    _client_cache = {}
+
+    def __init__(self, judge_type: str, verifier_config: VerifierConfig) -> None:
+        super().__init__(f"general-{judge_type}", verifier_config=verifier_config, weight=1.0)
+        self.prompt_template = JUDGE_PROMPT_MAP[judge_type]
```
fabrahman, line 509: asserting if judge_type is implemented?

open_instruct/ground_truth_utils.py
```diff
@@ -371,33 +440,182 @@ class UpToMaxLenVerifier(VerifierFunction):
...
+        model_name = model.replace("-standard", "") #azure OAI models have -standard in the name
+        return PRICE_PER_TOKEN[model_name]["input"] * response.usage.prompt_tokens + PRICE_PER_TOKEN[model_name]["output"] * response.usage.completion_tokens
 
+    async def async_call(self, tokenized_prediction: List[int], prediction: str, label: str, query: str) -> VerificationResult:
+        """
+        Asynchronous version of __call__ that properly handles the async OpenAI client.
+        """
+        client = self._get_client()
+        
+        prompt = self.prompt_template.format(input=query, output=prediction, label=label)
```
fabrahman: so we're not modifying the prompt to add "response_format specification" as we did previously using this class method [here](https://github.com/allenai/open-instruct/blob/47a42c724516a42c192831e92239c5926fcdb73a/open_instruct/judges/base.py#L102-L128)?
we can potentially add `response_format={"type": "json_object"}` as an arg to chat_completion function but I'm not sure how much removing that build_message function will impact the llm judge response format

open_instruct/grpo_fast.py
```diff
@@ -1816,7 +1831,8 @@ def reward_fn(
 
         if args.apply_verifiable_reward:
             with Timer("[Data Preparation Thread] Calculating rewards -- 🏆 Applying verifiable reward"):
-                verifiable_rewards, per_func_rewards = apply_verifiable_reward(
+                verifiable_rewards, per_func_rewards = await apply_verifiable_reward(
+                    reward_fn_mapping,
```
fabrahman, line 1881: I don't see queries being added here

open_instruct/grpo_fast.py
fabrahman, line 1: Initially, I decoded queries and then use `extract_user_query` to extract the original query string (removing template), but my function only handled r1_style template.
I'd suggest to get the query text from the data_transformation instead. I guess we should use the `DEFAULT_SFT_MESSAGES_KEY` ?

fabrahman: Like adding:

messages_next = data_next[DEFAULT_SFT_MESSAGES_KEY]
similar to queries_next etc and use that instead of decoding and extracting

open_instruct/ground_truth_utils.py
```diff
@@ -371,33 +440,182 @@ class UpToMaxLenVerifier(VerifierFunction):
...
+        
+        max_retries = 3
+        retry_delay = 1.0
+        
+        for attempt in range(max_retries):
+            try:
+                response = await client.chat.completions.create(
+                    model=self.verifier_config.llm_judge_model,
```
fabrahman, line 583: where are you passing temperature and max_token?
btw I realized for more recent models we should use `max_completion_tokens` instead of max_tokens in the chat_completion function.

fabrahman (commented):
Left several comments. Let's discuss after you check them. Thanks

open_instruct/ground_truth_utils.py
saurabh111233212: yep good call, I went with Hamish's idea to have a blank general config and then subclass for specific verfiers (like LM Judge, Code, etc)

scripts/train/debug/judge.sh
saurabh111233212: yep that's correct, otherwise they'll be set to their defaults

open_instruct/ground_truth_utils.py
saurabh111233212: will need to add support for that, maybe through litellm or maybe through a client map or something. TBD

open_instruct/ground_truth_utils.py
saurabh111233212: we initialize a verifier for each type. It'll fail out if a type in the dataset is not implemented

open_instruct/grpo_fast.py
saurabh111233212: yep sorry, fixed

open_instruct/ground_truth_utils.py
saurabh111233212: fixed, ty

open_instruct/grpo_fast.py
saurabh111233212: No Args is what's parsed from the grpo_fast script at the top level, and whatever you pass is used to populate the VeriferConfigs

PR #704, opened 2025-06-04
better IF

ValentinaPy:
I fixed the verifier functions (now using the same setup as the original IFEval verifiers). I also added 29 new training constraints and verifier functions. We can now use the original 25 constraints + verifier functions from IFEval and the additional new 29 constraints. I also changed the IFEvalVerifier class, so that it now can handle multiple constraints at once!
(in a later pull request I will add my IF verifiers + reward model code)
Lastly, I added an example script for how I ran IF-GRPO training.


open_instruct/ground_truth_utils.py
```diff
@@ -235,31 +237,36 @@ class IFEvalVerifier(VerifierFunction):
...
+        super().__init__("ifeval", weight=1.0)
 
-    def __call__(
-        self, tokenized_prediction: List[int], prediction: str, label: Union[str, Dict], query: Optional[str] = None
-    ) -> VerificationResult:
-        constraint = label
+    def __call__(self, tokenized_prediction: List[int], prediction: str, label: Union[str, Dict]) -> bool:
+        instruction_dict = instructions_registry.INSTRUCTION_DICT
+        constraint_dict = ast.literal_eval(label)
+        constraint_dict = constraint_dict[0]
```
hamishivi, line 270: Should this be under a check? Old ifeval data wasn't in a list...

open_instruct/IFEvalG/requirements.txt
```diff
@@ -0,0 +1,4 @@
+absl
```
hamishivi: Can we add this to pyproject.toml and the requirements at the top level?

hamishivi (commented):
Some comments but looks good! Also would be good to lint (make quality/make style)

open_instruct/ground_truth_utils.py
```diff
@@ -235,31 +237,36 @@ class IFEvalVerifier(VerifierFunction):
...
     specified in the constraint.
 
-    The constraint may be a JSON string or a dictionary containing a key
-    'func_name' used to lookup the evaluation function.
+    The constraint(s) are a list of constraint ids.
+    This list is found under the key "instruction_id" in the ground_truth dict.
+
     """
 
-    def __init__(self, verifier_config: Optional[VerifierConfig] = None) -> None:
```
hamishivi, line 259: please keep the verifier config in the init

hamishivi (approved):
LGTM, assuming you've tested it runs fine!

PR #705, opened 2025-06-04
SFT -> OlmoCore data conversion script

tyler-romero:
Can be run with a command like:
```
gantry run --cluster ai2/phobos-cirrascale --allow-dirty --timeout -1 -y --budget ai2/oe-training --install "curl -LsSf https://astral.sh/uv/install.sh | sh && /root/.local/bin/uv sync" --weka=oe-training-default:/data -- /root/.local/bin/uv run python scripts/data/convert_sft_data_for_olmocore.py --add_bos --output_dir /data/tylerr/data/sft/tulu-3-sft-olmo-2-mixture-0225-olmocore
```

I'm sure theres a cleaner way that doesnt involve my hacky uv usage but I'm still learning my way around all the systems.


pyproject.toml
```diff
@@ -7,7 +7,7 @@ requires-python = ">=3.10,<3.12"
 dependencies = [
     "accelerate>=1.1.1",
     "antlr4-python3-runtime==4.11",
-    "bitsandbytes>=0.44.1",
+    "bitsandbytes>=0.44.1; platform_system != 'Darwin'",
```
tyler-romero, line 10: make `uv sync` work on mac

scripts/data/convert_sft_data_for_olmocore.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def write_memmap(filename, data, dtype):
+        mmap = np.memmap(filename, mode="w+", dtype=dtype, shape=(len(data),))
+        mmap[:] = data
+        mmap.flush()
+        return mmap
+
+    write_memmap(f"{args.output_dir}/token_ids.npy", token_ids, np.uint32)
+    write_memmap(f"{args.output_dir}/labels.npy", labels, np.int32)
+    write_memmap(f"{args.output_dir}/attention_mask.npy", attention_mask, np.int32)
```
epwalsh: I don't know how big these datasets are, but if they are large enough then it would make sense to write out multiple numpy files like we do for pretraining datasets.

tyler-romero: The dataset that was used for Olmo2 SFT is only 2GB for the token_ids.npy file. I went ahead and added support for writing out multiple numpy files if larger datasets are used in the future though.

epwalsh (approved):

natolambert (commented):
I'm happy with these if Hamish and Pete are ;)

PR #707, opened 2025-06-06
Fix oe-eval bracket

hamishivi:
bugfix. auto evals broken without this.


saumyamalik (commented):
good catch, looks good

PR #710, opened 2025-06-06
vLLM hosted judge + general verifier

hamishivi:
Changes needed to use: https://huggingface.co/TIGER-Lab/general-verifier

Also refactor setup a bit: we now can specify an extractor for a given judge/prompt. This hopefully provides some more flexibility with templates in the future.

I think for now, lets just do the simple/dumb thing of we launch a judge separately, and then ping it repeatedly. This is annoying if we want to develop on judges, but I think makes sense for now.

see scripts/train/rlvr/judge_general_verifier.sh for how to run this yourself.

Example run: https://wandb.ai/ai2-llm/open_instruct_internal/runs/78ow09w2


saurabh111233212 (approved):
nice! I like the extractor map thing, makes sense to me.

PR #712, opened 2025-06-08
Update uv command + tokenizer fixes

hamishivi:
found a slightly better uv command to generate requirements, just edited to fix file.

Also:
- allow vllm to use a custom tokenizer. This might be useful in future, but not sure. Added anyway. Either way, this keeps things more consistent (if the user specifies a custom tokenizer, vllm will now respect that)
- Guard against the padding token appearing in our tokenized inputs in RLVR tokenize. Turns out this was happening and was the cause of my vllm errors. I don't know why exactly its happened, but adding this guard is reasonable anyway. We assume a pad token later (when constructing the batch), so we want to keep this extra pad token around.


saurabh111233212 (approved):
lgtm assuming it's tested

PR #713, opened 2025-06-10
Fix judge templates

hamishivi:
format call stuffed up them, i think this should fix.


fabrahman:
This is working fine!
thanks

PR #715, opened 2025-06-12
DSCHF Fixes

hamishivi:
We need to keep dschf around to keep deepspeed behaving properly, which is important for large (32b) runs. this assigns the object to a variable and so keeps it alive.

Yes, its very spooky.


natolambert (approved):
im just wondering how you found this

PR #722, opened 2025-06-16
Various bag of fixes.

hamishivi:
1. Remove thinking traces for IFEval and GSM8k from output.
2. Add 'world padding'. This will prevent the issue of not having enough data to run a gradient step after packing in some cases, at the cost of just wasting some compute. This is mainly useful if you want to run multiple models with similar hypers, and one model just isnt training. In other cases, if you see this happen on runs, you should really just increase the batch size or reduce packing length or similar.
3. Add process_output by default to our eval runs. This means the thinking trace is stripped from the output before being passed to the evaluation, which makes AlpacaEval/IFEval style evaluations actually make sense for thinking models (we shouldn't evaluate the thinking process, just the final actual answer).

### DIFF
--- a/open_instruct/ground_truth_utils.py
+++ b/open_instruct/ground_truth_utils.py
@@ -282,7 +282,7 @@ class IFEvalVerifier(VerifierFunction):
         constraint_dict = constraint_dict[0]
         if isinstance(constraint_dict, str):
             constraint_dict = json.loads(constraint_dict)
-        answer = remove_thinking_section(prediction)
+        answer = remove_thinking_section(prediction).strip()
         instruction_keys = constraint_dict["instruction_id"]
         args_list = constraint_dict["kwargs"]
         rewards = []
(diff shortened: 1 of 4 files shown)


ValentinaPy (approved):
LGTM, I tested it in a run yesterday and it works great!

saurabh111233212:
I approved - still a bit confused when you get into situations where batch size is less than world size but looks fine to me. Maybe should be a flag that defaults to off if it's wasting compute? I don't feel strongly abt this

hamishivi:
To explain why we might need world padding for posterity:

If we pack such that we only have 2 batches, but 8 gpus, then currently the code just skips the entire step. Sometimes we end up with models that output really short outputs and we end up always skipping training. This 'fixes' by just making fake batches that don't contribute gradients for the rest of the gpus to process. this wastes compute, so I'm flag-gating it.... generally if you hit this scenario you should just increase batch size / decrease packing length, but sometimes if you want to do sweeps across models with the same hypers it can be useful to use this.

PR #729, opened 2025-06-19
add long-form rl-rag reward

RulinShao:
# Reward Design

### Global Rewards
* Rubric rewards. LLM-as-a-judge
    * For each training/test sample, we generate per-sample rubrics for evaluation. For example, the rubric could be “the starting of the answer should introduce the background of XXX”. We use LLM-as-a-judge for rubric evaluation, which outputs a scalar score. We (weighted) average over all rubrics as the final rubric reward.
* Global rubric rewards. LLM-as-a-judge
    * This is a placeholder for a general rubric evaluation prompt for all examples.
* Length rewards (optional).
    * We use this reward to steer the generation towards the ideal length.

### Citation Rewards.
* Citation rewards (F1, prompts adapted from [LongCite](https://arxiv.org/abs/2409.02897)).
    * Citation factuality.
        * Check if the generated citation id is a valid id present in context. Return #valid_ids / #total_ids as the reward.
    * Citation quality Avg(F1=2(R*P)/(R+P)) LLM-as-a-judge
            * For each claim, compute the recall and precision.
    * Recall.
        * When the claim has no citation:
            * Return 1 if the model thinks this sentence does not require citation; return 0 otherwise.
        * When the claim has citations:
            * Return 1 if the claim is fully supported by corresponding citations;
            * Return 0.5 if the claim is partially supported by corresponding citations;
            * Return 0 if the claim is not supported by the citations.
    * Precision.
        * For citations, return 1 if it is relevant, else 0.

### DIFF
--- a/open_instruct/search_rewards/README.md
+++ b/open_instruct/search_rewards/README.md
@@ -1,0 +1,265 @@
+# Search Rewards Module
+
+This module provides a comprehensive reward scoring system for evaluating responses to paper search queries. It uses a rubric-based approach with multiple scoring components to assess the quality of answers.
+
+## Overview
+
+The reward function evaluates responses based on several criteria including:
+- **Length appropriateness** - Optimal response length (300-600 words)
+- **Expertise level** - Alignment with expected user expertise
+- **Citations and excerpts** - Proper citation usage and supporting quotes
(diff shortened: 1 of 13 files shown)


hamishivi:
Closing since we have actual dr tulu repo now :D

PR #731, opened 2025-06-20
Ruff and linting in pyproject.toml

mnoukhov:
I've moved all the linting settings from isort and black into `pyproject.toml` and I've also added support for `ruff` which has become the default linter, formatter, and import sorter in many python projects. It also allow for easy adding autoformat on save.

I set the parameters of ruff so it matches our current `black`, `isort --profile black`, `flake8`, and `autoflake`. 
In doing so, I've also fixed two small errors in `open_instruct/ground_truth_utils.py` and `open_instruct/merge_lora.py` 

To do autoformat on save in VSCode, you can install the [Ruff extension](https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff) and make it auto-format, auto-lint, and auto-organize on save by changing `settings.json` with

```json
{
  "[python]": {
    "editor.formatOnSave": true,
    "editor.codeActionsOnSave": {
      "source.fixAll": "explicit",
      "source.organizeImports": "explicit"
    },
    "editor.defaultFormatter": "charliermarsh.ruff"
  }
}
```

small notes: 
- flake8's W503 doesn't need to be ignored in ruff because it is excluded by default
- I've kept the formatting to files under `open_instruct/` as with the current `make style` and `make quality` commands

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,8 @@ export PYTHONPATH = open_instruct
 check_dirs := open_instruct 
 
 style:
-	python3 -m black --line-length 119 --target-version py310 $(check_dirs)
-	python3 -m isort $(check_dirs) --profile black
+	python3 -m black $(check_dirs)
+	python3 -m isort $(check_dirs)
 
 quality:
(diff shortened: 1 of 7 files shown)


hamishivi (approved):
thanks!

PR #732, opened 2025-06-20
Div zero fixes

hamishivi:
Two fixes:
1. grad accumulation steps could hit zero if you asked for more minibatches than we had batch sizes after packing. This guards against that so we always just at least do one batch. If you ask for multiple minibatches but we have 1 batch, we just do one batch.
2. guards against a division by zero error in the packing code. If you have a batch with all 0/1 reward, we filter everything out. This actually runs fine until the metrics call, which errors since len(responses) is 0 and we try to divide by it. Later in the code, we do handle the case where the packing is empty fine (B == 0, we skip the batch or do world padding).

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -753,7 +753,8 @@ class PolicyTrainerRayProcess(RayProcess):
         to_device_inplace(collated_position_ids, self.device)
         to_device_inplace(collated_advantages, self.device)
         to_device_inplace(collated_response_masks, self.device)
-        accumulation_steps = math.ceil(len(collated_query_responses) / num_mini_batches - 0.5)
+        # accumulation steps should always be at least 1
+        accumulation_steps = max(math.ceil(len(collated_query_responses) / num_mini_batches - 0.5), 1)
         leftover = len(collated_query_responses) % accumulation_steps
         if leftover > 0:
             collated_query_responses = collated_query_responses[0:-leftover]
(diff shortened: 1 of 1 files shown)


open_instruct/grpo_fast.py
```diff
@@ -753,7 +753,8 @@ def train(
         to_device_inplace(collated_position_ids, self.device)
         to_device_inplace(collated_advantages, self.device)
         to_device_inplace(collated_response_masks, self.device)
-        accumulation_steps = math.ceil(len(collated_query_responses) / num_mini_batches - 0.5)
+        # accumulation steps should always be at least 1
+        accumulation_steps = max(math.ceil(len(collated_query_responses) / num_mini_batches - 0.5), 1)
```
vwxyzjn, line 757: Is this the case where one device's `len(collated_query_responses)` is 0? Also what is this `0.5` in `num_mini_batches - 0.5`?

hamishivi: It covers the case where we have ask for more minibatches than we have in collated_query_responses such that `len(collated_query_responses) / num_mini_batches` is small (e.g. 0.1)!

The -0.5 is just a way to do rounding done in the PR (#698), could swap out with Math.round()... basically, we want to get the rounded number of accumulation steps, and then throw out the remaining (hopefully not much) data. Before the logic wasn't right if `len(collated_query_responses) % num_mini_batches != 0`.

hamishivi:
Closing since we'll merge these changes in as part of #733

PR #733, opened 2025-06-22
fix context len

fabrahman:
We often hit the context length limit when using the local judges:

`RequestError: ContextWindowExceededError: Hosted_vllmException - This model's maximum context length is 8192 tokens. However, you requested 10840 tokens (8792 in the messages, 2048 in the completion). Please reduce the length of the messages or completion. None`

What this PR has:

1. add a new judge config for `llm_judge_max_context_length` which is the maximum context length for the local judge
2. checks the full judge message length and truncate if Len >  `llm_judge_max_context_length` -  `llm_judge_max_tokens` (considering some safe margin)
3. Slightly modified the ref-based chat_template 

Note: I released the local judge (specifically Qwen3-32b) generate responses in thinking style, having thinking tokens between <think> tags, in addition to "REASONING" in the json format we ask in the judge prompt. We might want to tweak the Owen chat template to prevent repetitive reasoning content.

### DIFF
--- a/open_instruct/context_window_checker.py
+++ b/open_instruct/context_window_checker.py
@@ -56,9 +56,18 @@ def get_encoding_for_model(model_name: str):
 
     # Models that use gpt2 encoding (including OLMo and other AI2 models)
     gpt2_models = [
-        "qwen", "llama", "llama2", "llama3", "mistral", "codellama",
-        "olmo", "olmoe", "olmo2",  # AI2 OLMo models use GPTNeoX tokenizer (GPT-2 based)
-        "allenai/olmo", "allenai/olmoe", "allenai/olmo2"  # Full model names
+        "qwen",
+        "llama",
+        "llama2",
+        "llama3",
+        "mistral",
+        "codellama",
+        "olmo",
+        "olmoe",
+        "olmo2",  # AI2 OLMo models use GPTNeoX tokenizer (GPT-2 based)
+        "allenai/olmo",
+        "allenai/olmoe",
+        "allenai/olmo2",  # Full model names
     ]
     if any(model in model_name_lower for model in gpt2_models):
         return tiktoken.get_encoding("gpt2")
--- a/open_instruct/ground_truth_utils.py
+++ b/open_instruct/ground_truth_utils.py
@@ -642,7 +642,7 @@ class LMJudgeVerifier(VerifierFunction):
         for attempt in range(max_retries):
             # judges the quality of a response
             try:
-                system_prompt = "Do not generate text between the <think> and </think> tags." # "You are a concise assistant who gives very short explanations before giving a quality score."
+                system_prompt = "Do not generate text between the <think> and </think> tags."  # "You are a concise assistant who gives very short explanations before giving a quality score."
                 messages = build_messages(prompt, system_prompt)
 
                 # Faeze: check if the request would exceed context window
--- a/open_instruct/judge_utils.py
+++ b/open_instruct/judge_utils.py
@@ -206,16 +206,16 @@ def extract_json_score_with_fallback(score_str: str) -> "tuple[str, float]":
     try:
         # Strip markdown code blocks if present
         cleaned_str = score_str.strip()
-        if cleaned_str.startswith('```json'):
+        if cleaned_str.startswith("```json"):
             cleaned_str = cleaned_str[7:]  # Remove ```json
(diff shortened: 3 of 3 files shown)


configs/judge_configs/general_verifier_judge.yaml
```diff
@@ -24,7 +26,8 @@ tasks:
       cluster:
         - ai2/saturn-cirrascale
     resources:
-      gpuCount: 1
+      gpuCount: 2
     context:
       priority: high
       preemptible: False
+#  hamishivi/general-verifier
```
hamishivi: remove this line?

open_instruct/ground_truth_utils.py
```diff
@@ -600,7 +601,9 @@ def parse_completion(self, completion):
             return reasoning, score
 
         try:
-            content = completion.choices[0].message.content
+            # remove anything between <think> and </think> including the tags using regex
+            pattern = r"<think>\s*.*?\s*</think>\s*"
```
hamishivi, line 606: Some models aslo wrap their answer in `<answer>...</answer>`. Lets also remove those?
```
content = content.replace("<answer>", "").replace("</answer>", "")
```

open_instruct/ground_truth_utils.py
```diff
@@ -635,14 +638,60 @@ async def async_call(
         retry_delay = 1.0
 
         for attempt in range(max_retries):
+            # judges the quality of a response
             try:
-                messages = build_messages(prompt)
+                system_prompt = "Do not generate text between the <think> and </think> tags." # "You are a concise assistant who gives very short explanations before giving a quality score."
```
hamishivi: 🤣

open_instruct/ground_truth_utils.py
```diff
@@ -635,14 +638,60 @@ async def async_call(
...
+                        truncate_messages_to_fit_context,
+                    )
+
+                    context_check_available = True
+                except ImportError:
+                    logger.warning("Context window checker not available. Proceeding without context checking.")
+                    context_check_available = False
+
+                # Check if the request would exceed context window
+                if context_check_available:
```
hamishivi, line 662: should we also check if we have a compatible tokenizer? For example, I don't think litellm supports the olmo tokenizer.

scripts/eval/oe-eval.sh
hamishivi, line 1: please put this back!

open_instruct/context_window_checker.py
```diff
@@ -0,0 +1,367 @@
...
+
+        if remaining_tokens <= 0:
+            logger.warning("System messages alone exceed judge context window. Keeping only system messages.")
+            return system_messages
+
+        # Truncate other messages to fit
+        truncated_messages = system_messages.copy()
+        current_tokens = system_tokens
+
+        for msg in other_messages:
```
hamishivi, line 198: it seems currently this truncates from the right, do we want that behaviour? I wonder if we should truncate from the left after the system message? Not too strong opinions here.

hamishivi (commented):
A few comments but happy this seems to be working!

configs/judge_configs/general_verifier_judge.yaml
fabrahman: done

open_instruct/ground_truth_utils.py
fabrahman: done

scripts/eval/oe-eval.sh
fabrahman: done

open_instruct/context_window_checker.py
fabrahman: Hmmm, we can potentially do that.

open_instruct/ground_truth_utils.py
fabrahman: isn't OLMo using `GPTNeoXTokenizerFast`? I just modified to use gpt-2 tokenizer for olmo. wdyt?
I guess having a safety margin will prolly help with tokenizer differences?

fabrahman:
I've pushed changes to address most of the comments. 
@hamishivi  Let me know if anything else needs attention

open_instruct/ground_truth_utils.py
hamishivi: sounds good!

hamishivi (approved):

PR #736, opened 2025-06-25
Save tokenizer files alongside olmocore dataset

tyler-romero:
This helps track how the data was created and these files are useful when converting back to hf format downstream.

### DIFF
--- a/scripts/data/convert_sft_data_for_olmocore.py
+++ b/scripts/data/convert_sft_data_for_olmocore.py
@@ -10,16 +10,17 @@ Usage:
         --tokenizer_name_or_path allenai/OLMo-2-1124-7B \
         --add_bos \
         --dataset_mixer_list allenai/tulu-3-sft-olmo-2-mixture-0225 1.0 \
-        --output_dir /weka/oe-adapt-default/tylerr/tulu-3-sft-olmo-2-mixture-0225-olmocore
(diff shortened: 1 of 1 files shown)


scripts/data/convert_sft_data_for_olmocore.py
```diff
@@ -10,16 +10,17 @@
         --tokenizer_name_or_path allenai/OLMo-2-1124-7B \
         --add_bos \
         --dataset_mixer_list allenai/tulu-3-sft-olmo-2-mixture-0225 1.0 \
-        --output_dir /weka/oe-adapt-default/tylerr/tulu-3-sft-olmo-2-mixture-0225-olmocore
+        --output_dir ./data/tulu-3-sft-olmo-2-mixture-0225-olmocore
```
jacob-morrison, line 13: did you mean to change this directory to ./data/? I don't see you using the same directory anywhere else so I'm not sure

jacob-morrison (approved):
lgtm overall, just one question about a directory

scripts/data/convert_sft_data_for_olmocore.py
tyler-romero: Yeah good Q, since the internal command is below I didnt want to specify /weka there

PR #739, opened 2025-06-30
code stdio format

saurabh111233212:
infra and verifier for executing code in the `stdio` format (e.g. using `print` and `input` to read/write to console instead of implementing a function)

this is useful for scripting, and about half of livecodebench examples are of this format, so we prob wanna train with it. 

Test job [here](https://beaker.allen.ai/orgs/ai2/workspaces/oe-adapt-code/work/01JZ10DJSMB1J0XEZ7EZ8RA355?taskId=01JZ10DJSQJND11DC72QG94GT7&jobId=01JZ10DJXAZQBX9SCZM10B3QYZ), wandb [here](https://wandb.ai/ai2-llm/open_instruct_internal/runs/3iss4wgd)

### DIFF
--- a/scripts/data/azure_batch/check_azure_batch_status.py
+++ b/scripts/data/azure_batch/check_azure_batch_status.py
@@ -48,8 +48,7 @@ def print_status(job: dict) -> None:
     errors = job.get("errors", [])
     print(f'{errors=}')
     error_str = "-" if not errors else ";".join(
-        (f"{e.get('code') or e.get('error_code')}:{e.get('message')}" if isinstance(e, dict) else str(e))
-        for e in errors
+        f"{e.get('code') or e.get('error_code')}:{e.get('message')}" for e in errors
     )
     line = (
         f"{job['status']}"


open_instruct/ground_truth_utils.py
```diff
@@ -825,9 +826,18 @@ def build_all_verifiers(args) -> Dict[str, VerifierFunction]:
...
 
-        instance = subclass(subclass.get_config_class().from_args(args))
+        verifier_config = subclass.get_config_class().from_args(args)
+        instance = subclass(verifier_config)
         verifiers[instance.name.lower()] = instance
 
+        # add the code_stdio verifier
+        if subclass == CodeVerifier:
+            stdio_config = copy.deepcopy(verifier_config)
+            stdio_config.code_api_url = stdio_config.code_api_url.replace("/test_program", "/test_program_stdio")
```
hamishivi, line 836: Would there be a case where we want to run stdio verifier and regular code verifier together, and they end  up pointing to different places?

hamishivi (approved):
Looks good, one minor comment.

Do you think it would be worth unifying the old code verify and this stdio verify stuff or nah?

open_instruct/ground_truth_utils.py
saurabh111233212: @hamishivi 
I don't think what you're describing could happen with the way we're training rn. 
That said, I don't love the way we're branching rn. 
The `fn` and `stdio` paths do need to run through different execution infra at some point, but could flow better by maybe moving the branching further down - like inside the execution endpoint itself. 

This seems doable but not 100% straightforward, so prob not worth it IMO. Happy to give it a go if you disagree, though

PR #740, opened 2025-06-30
Padding-free SFT

garrett361:
This PR adds the ability to perform padding-free SFT, which reduces the memory costs and increases throughput when `per_device_train_batch_size>1`.  The model itself must support padding-free training to properly use this feature (Llama and bamba models support this, instance; see https://github.com/huggingface/transformers/pull/35861 for some typical throughput improvements).

Pass `--padding-free True` to use.

### DIFF
--- a/open_instruct/finetune.py
+++ b/open_instruct/finetune.py
@@ -367,6 +367,7 @@ class FlatArguments:
         default=False,
         metadata={"help": "Whether to use padding-free collation via TensorDataCollatorWithFlattening"},
     )
+
     def __post_init__(self):
         if self.reduce_loss not in ["mean", "sum"]:
             raise ValueError("reduce_loss must be either 'mean' or 'sum'")
--- a/open_instruct/finetune.py
+++ b/open_instruct/finetune.py
@@ -43,7 +43,6 @@ from accelerate import Accelerator, DataLoaderConfiguration
 from accelerate.logging import get_logger
 from accelerate.utils import InitProcessGroupKwargs, set_seed
 from huggingface_hub import HfApi
-from padding_free_collator import TensorDataCollatorWithFlattening
 from peft import LoraConfig, TaskType, get_peft_model, prepare_model_for_kbit_training
 from rich.pretty import pprint
 from torch.utils.data import DataLoader
@@ -64,6 +63,7 @@ from open_instruct.dataset_transformation import (
     visualize_token,
 )
 from open_instruct.model_utils import push_folder_to_hub, save_with_accelerate
+from open_instruct.padding_free_collator import TensorDataCollatorWithFlattening
 from open_instruct.utils import (
     ArgumentParserPlus,
     clean_last_n_checkpoints,
--- a/open_instruct/finetune.py
+++ b/open_instruct/finetune.py
@@ -43,6 +43,7 @@ from accelerate import Accelerator, DataLoaderConfiguration
 from accelerate.logging import get_logger
 from accelerate.utils import InitProcessGroupKwargs, set_seed
 from huggingface_hub import HfApi
+from padding_free_collator import TensorDataCollatorWithFlattening
 from peft import LoraConfig, TaskType, get_peft_model, prepare_model_for_kbit_training
 from rich.pretty import pprint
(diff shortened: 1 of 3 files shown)


vwxyzjn:
Is this packing?

garrett361:
Yeah, think it's the same @vwxyzjn . 

Basically rather than padding out uneven examples which are batched together, you just concatenate them and add additional info to say where example boundaries are.

garrett361:
Same idea [as here](https://huggingface.co/blog/packing-with-FA2)

garrett361:
FYI I'm double checking a possible issue with this at the moment; will report back after I have more info.

garrett361:
> I'm double checking a possible issue

Okay, I was debugging some bad-looking padding-free SFT training curves, but it turned out to be an issue with the `mamba` kernels which was [recently fixed here](https://github.com/state-spaces/mamba/commit/a6a1dae6efbf804c9944a0c2282b437deb4886d8). So, not an issue on the part of this commit.

garrett361:
Some plots of padding-free training. All curves use 8xA100s and have `--per_device_train_batch_size 2`. The only difference between any pair of curves on a plot is whether `--padding_free` is `true` or `false`.

First, tuluv3 training on a [BambaForCausalLM](https://huggingface.co/docs/transformers/main/en/model_doc/bamba#transformers.BambaForCausalLM) model. Loss curves are nearly identical (they will never be precisely the same due to numerics), and the throughput of the `--padding_free true` run is ~40% higher.
 
Scherm­afbeelding 2025-07-03 om 3 17 41 PM


And then the same experiment with [LlamaForCausalLM](https://huggingface.co/docs/transformers/main/model_doc/llama#transformers.LlamaForCausalLM). Similar results: loss curves are similar and `--padding_free true` througput is ~40% higher.

Scherm­afbeelding 2025-07-03 om 3 18 08 PM

CC @vwxyzjn @hamishivi

open_instruct/finetune.py
```diff
@@ -325,6 +326,10 @@ class FlatArguments:
 
     sync_each_batch: bool = False
     """Optionaly sync grads every batch when using grad accumulation. Can significantly reduce memory costs."""
+    padding_free: bool = field(
```
hamishivi: minor nit: could we call this `packing` instead of `padding_free`? To make it clearer to end-users what the feature is.
```suggestion
    packing: bool = field(
```

open_instruct/finetune.py
```diff
@@ -607,10 +612,18 @@ def main(args: FlatArguments, tc: TokenizerConfig):
         model.gradient_checkpointing_enable()
 
     # DataLoaders creation:
+    if args.padding_free:
```
hamishivi:
```suggestion
    if args.packing:
```

hamishivi (approved):
Thanks! So it seems like if the concatted sequence goes over the maximum supported length of the model, there might be issues? I guess in the ppo/grpo packing implementations, our code instead tries to fit samples into a given max length, and does use multiple microbatches if it has to. 

Generally happy to merge this in since its optional, just also might be nice to name it `packing` since I think that's a more common term that people will understand.

open_instruct/finetune.py
garrett361: yep, that is fine

garrett361: done

open_instruct/finetune.py
garrett361: done

garrett361:
> So it seems like if the concatted sequence goes over the maximum supported length of the model, there might be issues?

Maybe? I'm not sure what the max supported length affects, apart from RoPE embeddings? Or I guess similar mechanisms like alibi or just fixed positional embeddings. For the models I've used (llama, bamba) padding-free/packing with, there's no problem because the explicitly sequence dependent bits adjust accordingly. Certainly this only works for some model classes, which is why the collator raises a warning.

> I guess in the ppo/grpo packing implementations, our code instead tries to fit samples into a given max length, and does use multiple microbatches if it has to.

I'm only very familiar with the SFT `finetune.py` script at the moment, so can't yet speak to how you might use this for ppo/grpo packing.

PR #741, opened 2025-06-30
Pull unit tests out into separate files

finbarrtimbers:
Also updates the unittest workflow so that it runs all of these tests, rather than just `test_code_utils.py`.


finbarrtimbers:
As far as I can tell, the test was actually failing at head! There was a [bug](https://chatgpt.com/share/e/6862fd9f-ac08-800d-84ea-0aed142fe364).

saurabh111233212:
Oh wow, nice catch! Yeah, the reliability guard is probably not what we should be doing, but it's easy and mostly works, so need some good reason to get past the inertia. 

Good stuff. Have you run any tests that put up a server and call the same server multiple times? Just to make sure reliability guard is restored

finbarrtimbers:
> Oh wow, nice catch! Yeah, the reliability guard is probably not what we should be doing, but it's easy and mostly works, so need some good reason to get past the inertia.
> 
> Good stuff. Have you run any tests that put up a server and call the same server multiple times? Just to make sure reliability guard is restored

Added a bunch more tests!

saurabh111233212 (approved):
niiice. this is really important hygiene - thanks for doing this!

PR #747, opened 2025-07-01
rm run_name, add_seed_and_date_to_exp_name

garrett361:
This PR proposes two changes to the SFT `finetune.py` script:
1. Remove the currently-unused `run_name` arg
2. Allow the user to avoid appending the seed and date to the `exp_name`

For the first point, `run_name` is never actually used in the script. Instead, it is immediately overwritten by the `exp_name` value:
https://github.com/allenai/open-instruct/blob/82632ddcee48d58bd6ff6b86dda6337174d53707/open_instruct/finetune.py?plain=1#L416
So, I propose removing this no-op arg. 

For the second point, it is not always desirable to add the seed and start time to the experiment name.  Some reasons:
1. The date causes the experiment name to change when resuming an experiment.
2. In multi-node scenarios, the date can be different on each node and the date affects the name of the checkpoint directory, which I believe can cause issues with checkpointing and restart.


hamishivi (approved):

garrett361:
Thanks, @hamishivi!

PR #748, opened 2025-07-01
additional_model_arguments

garrett361:
This PR adds a catch-all `--additional_model_arguments` flag which allows the user to pass an arbitrary dictionary of arguments which are used to initialize the model.  Example usage:

```
--additional_model_arguments '{"z_loss": 1e-7, "some_boolean_key": true, "some_int_key": 1}'
```

Example use case: change the details of the RoPE implementation for the model.

### DIFF
--- a/open_instruct/finetune.py
+++ b/open_instruct/finetune.py
@@ -27,6 +27,7 @@ try:
 except Exception:
     pass
 # isort: on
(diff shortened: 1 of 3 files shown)


hamishivi (approved):
Looks good, thanks!

PR #749, opened 2025-07-01
Write out label_mask instead of labels in olmocore tokenization script

tyler-romero:


open_instruct/dataset_transformation.py
```diff
@@ -1180,9 +1317,44 @@ def __post_init__(self):
...
         self.dataset_range = dataset_range
-        if self.dataset_range > len(self.dataset):
-            raise ValueError("Dataset range exceeds dataset length")
-        self.dataset = self.dataset.select(range(self.dataset_range))
+        original_size = len(self.dataset)
+        self.original_dataset_size = original_size
+        
+        if self.dataset_range <= original_size:
+            # Normal case: select subset
+            self.dataset = self.dataset.select(range(self.dataset_range))
```
hamishivi: nit-ish: can we shuffle the dataset or indices a bit when we sub-select? I think random subselection is a better default than taking first n.

open_instruct/model_utils.py
```diff
@@ -404,14 +404,26 @@ def save_with_accelerate(
...
+    chat_template_name: str = "tulu",
 ) -> None:
     """`model_attribute_to_save` is for used to save PPO's policy instead of the full model"""
     # set the generation config to an empty setting to be safe.
     # we usually do greedy decoding for generation, so this should be okay.
     # otherwise, we get an error thrown at save time.
-    model.generation_config = transformers.GenerationConfig(
-        temperature=None, top_p=None, eos_token_id=tokenizer.eos_token_id, bos_token_id=tokenizer.bos_token_id
-    )
+    if "olmo" in chat_template_name:
```
hamishivi, line 419: can we add some logging that this is firing? I worry that someone might not realise this is firing when they have some model name like "coolmoment"

open_instruct/model_utils.py
```diff
@@ -404,14 +404,26 @@ def save_with_accelerate(
...
+        # New chat template has no bos token, and two eos tokens: <|im_end|> and <|endoftext|>
+        model.generation_config = transformers.GenerationConfig(
```
hamishivi, line 422: does this saving change also need to be added to the grpo/ppo scripts?

jacob-morrison: I think it does, only for from-zero RL experiments. if we're RL'ing on top of a previously SFT'd/DPO'd model, it should already have the updated generation config, but I updated all the save functions for PPO/GRPO to be safe.

open_instruct/model_utils.py
jacob-morrison: done! it's based on the chat template name, but it can't hurt to be safe

open_instruct/dataset_transformation.py
jacob-morrison: yup, good call. this actually got me to simplify the logic, so now we just always call the same select_samples function, and it'll handle the <1.0 case as well as the >=1.0 case

hamishivi (approved):
Bit hard to see the changes with all the changes from main but I think my main comments got addressed! I see this is merging into the chat template PR so can give that one a look over once this is merged too.

PR #755, opened 2025-07-02
Add entropy calc + fix clipping logic.

hamishivi:
1. Entropy of our policy is a useful to track! This adds the ability to track it (optional, default on).
2. Our clipping logic was wrong for multiple minibatches in the fast script! Clipping should come into play we take any gradient step, but our fast scripts only did clipping on multiple num_epochs, rather than multiple mbsz. This is probably why I've found it unstable... mega thanks to Teng for pointing this out.


open_instruct/grpo_fast.py
```diff
@@ -775,10 +794,39 @@ def train(
                     ref_logprob = torch.masked_fill(ref_logprob, ~response_mask[:, 1:], INVALID_LOGPROB)
                     collated_ref_logprobs.append(ref_logprob)
                     torch.cuda.empty_cache()
+        # if we have multiple minibatches, we need to calculate the old logprobs for each minibatch
+        # following gtrl scripts in just doing this on the current active policy, rather than use the logprobs
+        # from the generator (note that async mode means these are a bit diff!)
+        old_logprobs = [None for _ in range(len(collated_query_responses))]
+        if num_mini_batches > 1:
```
mnoukhov, line 801: honestly, it is a slight performance hit but probably not a huge one to just always get the old_logprobs here. could we worth switching to that for clear code

PR #759, opened 2025-07-04
Adds a test for the generator for `grpo_fast.py`

finbarrtimbers:
Test verifies we can perform basic operations with the generator. The test runs on GPU with all the normal code as an integration test. On CPU, it doesn't do anything. The goal is to give me something fast I can run while implementing async generation.

Note: I know that it's not generally good to do exact string matching on LLM outputs, but I want to verify that we're not changing anything accidentally. I'm open to suggestions on how to do this better. I did run this 50x on a GPU machine and they all passed, so there shouldn't be stochasticity.

### DIFF
--- a/open_instruct/test_grpo_fast.py
+++ b/open_instruct/test_grpo_fast.py
@@ -89,7 +89,8 @@ class TestGrpoFastVLLM(unittest.TestCase):
         # Decode the response
         generated_text = tokenizer.decode(response_ids[0], skip_special_tokens=True)
 
-        self.assertEqual(generated_text, '\n\nThe""')
+        self.assertIsInstance(generated_text, str)
+        self.assertGreater(len(generated_text), 0)
 
         generate_thread.join(timeout=5)
 


open_instruct/test_grpo_fast.py
```diff
@@ -0,0 +1,99 @@
...
+            num_engines=1,
+            tensor_parallel_size=1,
+            enforce_eager=True,
+            tokenizer_name_or_path=tokenizer_name,
+            pretrain=tokenizer_name,
+            revision="main",
+            seed=42,
+            enable_prefix_caching=False,
+            max_model_len=512,
+            vllm_gpu_memory_utilization=0.5,  # Use less GPU memory for testing
```
hamishivi, line 50: not sure we neeeeed to use lower memory since the test is running on its own(?) but for a really small model probably doesn't matter.

hamishivi (approved):
Seems good, discussed a bit on slack.

PR #760, opened 2025-07-08
Update to new eval suite

natolambert:
First PR with UV, dang


scripts/eval/oe-eval.sh
```diff
@@ -64,6 +65,7 @@ while [[ "$#" -gt 0 ]]; do
         --revision) REVISION="$2"; shift ;;
         --max-length) MAX_LENGTH="$2"; shift ;;
         --unseen-evals) UNSEEN_EVALS="true" ;;
```
hamishivi: Maybe we should just remove unseen evals flag if we are swapping to specifying the task suite by name?

PR #762, opened 2025-07-09
Add script filtering "I was made by"

natolambert:


scripts/data/filtering_and_updates/filter_dataset_by_keywords.py
```diff
@@ -0,0 +1,217 @@
+#!/usr/bin/env python3
```
hamishivi, line 1: nit: it'd be good to have a short description and example usage at the top of the script.

PR #757, opened 2025-07-03
For `grpo_fast`, refactored main() to use functions.

finbarrtimbers:
Previously, the `main()` definition was a 508 line monolith. This PR breaks it up into a bunch of separate function calls, which takes it down to NN lines. 

Similarly, I move the `reward_fn` definition out of the `if __name__ == '__main__'` block to make that more readable.

PR #764, opened 2025-07-09
Update grpo_fast debug script

hamishivi:
- Using Qwen means we usually actually get some reward
- lower vllm usage % to make it not oom


mnoukhov:
already happening in #757 fyi

PR #765, opened 2025-07-09
Add new olmo chat templates, and improve data mixing/tokenization

jacob-morrison:
Adds new olmo-core-compatible chat templates. Includes:
* New olmo template with support for function-calling. Includes a basic hard-coded system prompt, and appends "You do not have access to any functions" to any SFT examples that do not include functions.
* Thinker version of the above template, has `<think>` included in the generation prompt
* R1-style thinker template
These 3 templates mirror our current Tulu templates

Also includes some necessary changes to the `--add_bos` logic, to handle the new chat template which does not have a bos token.

Includes a few other QoL fixes:
* Fixes a bug in the olmocore tokenization script re: label mask
* Logs dataset-level statistics during data mixing and tokenization
* Supports easy upsampling during data mixing


open_instruct/dataset_transformation.py
```diff
@@ -273,6 +273,160 @@ def visualize_token_role(tokens: list[int], masks: list[int], tokenizer: PreTrai
...
         "{% endfor %}"
     ),
+    # olmo-core-compatible chat templates:
+    "olmo": (
+        "{% for message in messages %}"
+        "{% if message['role'] == 'system' %}"
+            "{% if message.get('functions', none) is not none %}"
+                "{{ '<|im_start|>system\n' + message['content'] + '\n' + '<functions>' + message['functions'] + '</functions><|im_end|>\n' }}"
+            "{% else %}"
+                "{{ '<|im_start|>system\n' + message['content']  + '<|im_end|>\n' }}"
```
hamishivi: Do you think its worth setting <|im_end|> as a stop token by default in evals/rl generation?

open_instruct/dataset_transformation.py
```diff
@@ -273,6 +273,160 @@ def visualize_token_role(tokens: list[int], masks: list[int], tokenizer: PreTrai
...
+        "{% endif %}"
+        "{% if loop.last and add_generation_prompt %}"
+            "{{ '<|im_start|>assistant\n' }}"
+        "{% endif %}"
+        "{% endfor %}"
+    ),
+    "olmo_thinker": (
+        "{% for message in messages %}"
+        "{% if message['role'] == 'system' %}"
+            "{% if message.get('functions', none) is not none %}"
```
hamishivi: Should we also have the system message from olmo_toolu in olmo and olmo_thinker?

open_instruct/dataset_transformation.py
natolambert: probably? We do this for eos now?

jacob-morrison: I'm hoping it's not necessary, if we include something like this in our generation_config: https://huggingface.co/Qwen/Qwen3-8B/blob/main/generation_config.json#L4-L7. If we set multiple eos tokens, from what I understand vllm will stop generations if it sees any eos tokens or a stop sequence? But if that doesn't work then we can just include `<|im_end|>`

open_instruct/dataset_transformation.py
natolambert: Only two exist:
1. olmo_toolu
2. olmo_thinker
I don't think we need a third, except for research, but not models we intend to release

open_instruct/dataset_transformation.py
natolambert: seems reasonable

open_instruct/dataset_transformation.py
jacob-morrison: I'm not sure. I was debating either including the system prompt or just not including the function-calling bits in the template without a system prompt, but it's a no op if the dataset doesn't have instances with functions, so I thought it couldn't hurt. I'd say we shouldn't include the default system prompt in `olmo`, just so we have an easy option to have no system prompt. 

For olmo thinker, are you still exploring tool use in RL? If so, what do you think? If not, I'm fine either way.

jacob-morrison:
> Could you also add maybe some docs or write something in the PR describing the chat templates with some details? e.g. how function works, system message, thinker vs non-thinker etc.

Yeah can do, I'll add it to the PR for now

open_instruct/dataset_transformation.py
```diff
@@ -533,9 +690,11 @@ def get_tokenizer_tulu_v2_2(tc: "TokenizerConfig"):
             # OLMo newer models use this tokenizer
             if tokenizer.bos_token is None:
                 tokenizer.bos_token = tokenizer.eos_token
-                assert tc.add_bos, (
-                    "For OLMo with GPTNeoX, you must add bos token to the beginning of the input sequence."
-                )
+                if "olmo" not in tc.chat_template_name:
```
natolambert, line 792: Is this backward compatible? E.g. if someone tries to reproduce an olmo 2 run and messes something up, I don't think this will trigger?

jacob-morrison: If you're replicating an olmo 2 run, you'd use the tulu chat template right, which should trigger the old functionality? These new templates are really for future models, but I didn't want to put `olmo3` in the name lol

natolambert: oh i see, NOT in

open_instruct/dataset_transformation.py
```diff
@@ -273,6 +273,160 @@ def visualize_token_role(tokens: list[int], masks: list[int], tokenizer: PreTrai
...
+    ),
+    "olmo_thinker_r1_style": (
+        "A conversation between User and Assistant. "
+        "The user asks a question, and the Assistant solves it. "
+        "The assistant first thinks about the reasoning process in "
+        "the mind and then provides the user with the answer. "
+        "The reasoning process and answer are enclosed within <think> </think> "
+        "and <answer> </answer> tags, respectively, "
+        "i.e., <think> reasoning process here </think> "
+        "<answer> answer here </answer>."
```
mnoukhov: I did a little bit of prompt tuning with qwen on simple tasks and found a couple things that could make this template a bit shorter and slightly nicer. Mainly
- fixing grammar (e.g. "think in the mind")
- I put the tags closer to the thing they represent and remove "respectively"
- change the capitalization of User and Assistant to match template (lowercase)

```
"A conversation between user and assistant. "
"The user asks a question, and the assistant solves it. "
"The assistant first thinks and reasons about the question "
"and after thinking provides the user with the answer. "
"The reasoning process is enclosed in <think> </think> tags "
"and the answer are enclosed in <answer> </answer> tags "
"so the full response is <think> reasoning process here </think> "
"<answer> answer here </answer>."
```

could be worth changing but I'll defer to @hamishivi

open_instruct/dataset_transformation.py
```diff
@@ -273,6 +273,160 @@ def visualize_token_role(tokens: list[int], masks: list[int], tokenizer: PreTrai
...
+        "{% if loop.last and add_generation_prompt %}"
+            "{{ '<|im_start|>assistant\n' }}"
+        "{% endif %}"
+        "{% endfor %}"
+    ),
+    "olmo_toolu": (
+        "{% set has_system = messages|selectattr('role', 'equalto', 'system')|list|length > 0 %}"
+        "{% if not has_system %}"
+            "{{ '<|im_start|>system\nYou are OLMo 2, a helpful function-calling AI assistant built by Ai2. Your date cutoff is November 2024, and your model weights are available at https://huggingface.co/allenai. You do not currently have access to any functions.<|im_end|>\n' }}"
+        "{% endif %}"
```
natolambert: We want a way that interleaves this with the hardcoded data...

open_instruct/dataset_transformation.py
tyler-romero: Yeah I agree w/ Jacob. When we're in the HF universe we should try to avoid special casing things and just get our configs right.

hamishivi: Can we add some code to make sure the generation config gets set correctly then?

open_instruct/dataset_transformation.py
hamishivi: No strong opinion! Probably having grammar and casing fixed at least is good.

open_instruct/dataset_transformation.py
hamishivi: If I'm reading the chat templates right, currently we only inject a system prompt for the toolu chat template. I guess I'm asking, should we consider having a default system prompt for the others to be a bit consistent?

hamishivi:
@soldni 

> It is weird to have system prompt backed in

this is what existing e.g. qwen3 instruct models do. Doesn't it make sense to have this so that off-the-shelf generation behaves nicely, esp if we train with system prompts?

> I am not sure why we have different templates for thinking and tool use. One template plz.

Lots of discussion for this on slack. The thinking prompt needs a different generation prompt (with `<think>`), merging the tool use and thinking templates beyond this is reasonable imo. Qwen relies on extra args during the chat template formatting to toggle this... we could swap to this but I don't have super strong feelings. It feels a bit weird.

> Please include function to bake template into a tokenizer

This already happens in our scripts when training happens

open_instruct/dataset_transformation.py
jacob-morrison: That'll be in olmo core sft -- that PR is still in progress, so I'll update that one separately

open_instruct/dataset_transformation.py
```diff
@@ -384,13 +357,13 @@ def visualize_token_role(tokens: list[int], masks: list[int], tokenizer: PreTrai
...
-        "the mind and then provides the user with the answer. "
-        "The reasoning process and answer are enclosed within <think> </think> "
-        "and <answer> </answer> tags, respectively, "
-        "i.e., <think> reasoning process here </think> "
+        "A conversation between user and assistant. "
+        "The user asks a question, and the assistant solves it. "
+        "The assistant first thinks and reasons about the question "
+        "and after thinking provides the user with the answer. "
+        "The reasoning process is enclosed in <think> </think> tags "
+        "and the answer are enclosed in <answer> </answer> tags "
```
mnoukhov: I accidentally left a mistake, should be "answer **is** enclosed"

jacob-morrison: fixed!

hamishivi:
Have we come to a decision on these @jacob-morrison ?

jacob-morrison:
> Have we come to a decision on these @jacob-morrison ?

yeah, my preference is we just have these 3 chat templates for now. it's simplest imo to treat these as drop-in replacements for our current templates, and so far they seem to be working fine in sft and dpo. if we just stick with these, anything else you think I should add to the pr?

open_instruct/dataset_transformation.py
jacob-morrison: I updated save_with_accelerate to also save the new eos token ids during sft, dpo and reward modeling

open_instruct/dataset_transformation.py
```diff
@@ -273,6 +273,132 @@ def visualize_token_role(tokens: list[int], masks: list[int], tokenizer: PreTrai
         "{% endif %}"
         "{% endfor %}"
     ),
+    # olmo-core-compatible chat templates:
+    # TODO: unify these 3 chat templates and send variables through the tokenizer's apply_chat_template kwargs
+    "olmo": (
```
tyler-romero, line 405: nit: should we call this olmo2 instead? Presumably olmo3 will need a different default?

open_instruct/dataset_transformation.py
```diff
@@ -597,7 +728,7 @@ class TokenizerConfig:
     tokenizer_revision: Optional[str] = None
     trust_remote_code: bool = False
     use_fast: bool = True
-    chat_template_name: str = "tulu"  # TODO: should I give an option to force override?
+    chat_template_name: str = "olmo"  # TODO: should I give an option to force override?
```
tyler-romero: is the comment here still relevant?

open_instruct/model_utils.py
```diff
@@ -404,14 +404,21 @@ def save_with_accelerate(
...
     # set the generation config to an empty setting to be safe.
     # we usually do greedy decoding for generation, so this should be okay.
     # otherwise, we get an error thrown at save time.
-    model.generation_config = transformers.GenerationConfig(
-        temperature=None, top_p=None, eos_token_id=tokenizer.eos_token_id, bos_token_id=tokenizer.bos_token_id
-    )
+    if "olmo" in chat_template_name:
+        # New chat template has no bos token, and two eos tokens: <|im_end|> and <|endoftext|>
+        model.generation_config = transformers.GenerationConfig(
+            temperature=None, top_p=None, eos_token_id=[100265, 100257]
```
tyler-romero: hardcoding token ids worries me, what if the vocab size changes for olmo4 or something? Seems like a sharp edge.

open_instruct/dataset_transformation.py
jacob-morrison: nope! olmo3 can use the same one; I've been testing olmo3 sft with this chat template and it's been fine

open_instruct/dataset_transformation.py
jacob-morrison: good point, it is not!

open_instruct/model_utils.py
jacob-morrison: yeah good point, updated to just grab the IDs for the relevant tokens:

```
>>> [tokenizer.convert_tokens_to_ids("<|im_end|>"), tokenizer.convert_tokens_to_ids("<|endoftext|>")]
[100265, 100257]
```

open_instruct/dataset_transformation.py
tyler-romero: But this one hardcodes OLMo 2!

tyler-romero (approved):
LGTM!

open_instruct/dataset_transformation.py
jacob-morrison: Ah good catch! I just updated it to `OLMo`, and we'll just need to finalize the hard-coded prompt (e.g. the date too) before we release

open_instruct/grpo_fast.py
```diff
@@ -983,8 +983,19 @@ def save_checkpoint_state(self, checkpoint_state_dir: str, client_state: Dict[st
                     checkpoint_state_dir, args.gs_checkpoint_state_dir
                 )
 
-    def save_model(self, output_dir: str) -> None:
+    def save_model(self, output_dir: str, chat_template_name: str, tokenizer: PreTrainedTokenizer) -> None:
         model_to_save = self.model
+        if "olmo" in chat_template_name:
+            # New chat template has no bos token, and two eos tokens: <|im_end|> and <|endoftext|>
+            model_to_save.generation_config = GenerationConfig(
+                temperature=None,
```
hamishivi: should we set a default temp / top_p? And maybe its good to put this GenerationConfig as a constant in a util somewhere so we only have to change it in one place for all scripts?

open_instruct/dataset_transformation.py
```diff
@@ -205,6 +206,132 @@ def visualize_token_role(tokens: list[int], masks: list[int], tokenizer: PreTrai
...
+        "{% endif %}"
+        "{% if loop.last and add_generation_prompt %}"
+        "{{ '<|im_start|>assistant\n' }}"
+        "{% endif %}"
+        "{% endfor %}"
+    ),
+    "olmo_thinker": (
+        "{% set has_system = messages|selectattr('role', 'equalto', 'system')|list|length > 0 %}"
+        "{% if not has_system %}"
+        "{{ '<|im_start|>system\nYou are OLMo 2, a helpful function-calling AI assistant built by Ai2. Your date cutoff is November 2024, and your model weights are available at https://huggingface.co/allenai. You do not currently have access to any functions. <functions></functions><|im_end|>\n' }}"
```
hamishivi: Olmo 2?

jacob-morrison: ah, I had already fixed this but merging in the other PR reverted it. fixed again

open_instruct/grpo_fast.py
jacob-morrison: yes to the constant -- I added a function to handle that.

default temp/top_p, probably? at least for release. not sure what good values would be though (esp. because this is for both thinking and not-thinking), so I'd say for now let's leave it as is and continue to use whatever settings we use during eval/etc

hamishivi: Sure, sounds good!

hamishivi (approved):

PR #767, opened 2025-07-09
next olmo and rl from base

mnoukhov:
- [x] update requirements and switch to new transformers
    - note that we are sticking to ray 2.46.0 because 2.48.0 gives an error for our local vllm install
- [x] install and compile custom olmo vllm locally
- [x] create a dockerfile for custom vllm
  - caveat: it is currently 58GB :/ I think we're adding a couple extra Gs by cloning vllm into the docker but that seems unavoidable imo
- [x] create a test script to run and eval olmo on standard data mix

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,6 @@ COPY open_instruct open_instruct
 
 # Install custom vllm for olmo3
 RUN git clone -b shanea/olmo3 https://github.com/2015aroras/vllm.git vllm_olmo3
-RUN VLLM_USE_PRECOMPILED=1 uv pip install -e vllm_olmo3
 
 # Install dependencies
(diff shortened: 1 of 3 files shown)


scripts/train/rlvr/grpo_olmo3.sh
```diff
@@ -0,0 +1,67 @@
+exp_name="grpo_fast_olmo3"
+dataset_mix="saurabh5/rlvr_acecoder 56878 hamishivi/rlvr_orz_math_57k_collected 56878 hamishivi/tulu_3_rewritten_400k_string_f1_only_v2 56878 allenai/IF_multi_constraints_upto5 56878"
```
saurabh111233212: nit: could you swap out the acecoder data for `saurabh5/rlvr_acecoder_filtered`

PR #779, opened 2025-07-13
Add waiting time metric to Wandb

finbarrtimbers:
Add `eval/wait_time_between_evals` metric to Wandb to measure the idle time between evaluation runs.

### DIFF
--- a/EVAL_WAIT_TIME_IMPLEMENTATION.md
+++ b/EVAL_WAIT_TIME_IMPLEMENTATION.md
@@ -1,0 +1,100 @@
+# Eval Wait Time Metric Implementation
+
+## Overview
+
+This implementation adds a new metric to track the time between eval jobs in the `grpo_fast.py` script. The metric measures how much time the eval job is waiting between runs, which helps identify potential bottlenecks in the evaluation pipeline.
+
+## Implementation Details
+
+### 1. Global Variable for Tracking
+
+Added a global variable to track the last eval finish time:
+
+```python
+# Global variable to track the last eval finish time
+# This is used to measure the time between when one eval job finishes and the next one is queued
+last_eval_finish_time = None
+```
+
+### 2. Recording Eval Finish Time
+
+Modified the `maybe_evaluate` function to record when an eval job finishes:
+
+```python
+# Record the time when eval job finishes
+# This timestamp is used to calculate the wait time for the next eval job
+global last_eval_finish_time
+last_eval_finish_time = time.time()
+logger.info("[Main Thread] 📊 Evaluation job finished")
+```
+
+### 3. Measuring Wait Time
+
+Modified the `vllm_generate_thread` function to measure and log the wait time when the next eval job is queued:
+
+```python
+# Record the time when eval job is queued and measure wait time since last eval finished
+global last_eval_finish_time
+current_time = time.time()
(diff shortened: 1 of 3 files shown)


EVAL_WAIT_TIME_IMPLEMENTATION.md
```diff
@@ -0,0 +1,109 @@
+# Eval Wait Time Metric Implementation
```
hamishivi, line 1: Could we put this file into a documentation folder or similar? I feel like also we don't necessarily need the full file, maybe just an explanation of the flag.

open_instruct/grpo_fast.py
```diff
@@ -1165,6 +1168,21 @@ def generate_with_engines(prompts: List[List[int]], sampling_params: SamplingPar
...
+            current_time = time.time()
+            if eval_timing_info is not None and eval_timing_info.get("last_eval_finish_time") is not None:
+                eval_wait_time = current_time - eval_timing_info["last_eval_finish_time"]
+                # Log the eval wait time to wandb if tracking is enabled
+                # This metric measures how much time the eval job was waiting between runs
+                # Note: We can't access args here, so we'll log to wandb directly if available
+                try:
+                    import wandb
+                    if hasattr(wandb, 'run') and wandb.run is not None:
+                        wandb.log({"eval/wait_time_between_evals": eval_wait_time}, step=training_step)
```
hamishivi, line 1181: I don't quite understand why we want to track time between evals. We can get an arb number of training steps inbetween evals, so isn't this usually just telling us how long n training steps is taking?

EVAL_WAIT_TIME_IMPLEMENTATION.md
finbarrtimbers: Ah, yeah, sorry. I'm trying out Cursor's background agents, and it put this here. Let me mark the PR (and the other Cursor ones) as draft until I clean it up. 

I totally agree with you.

open_instruct/grpo_fast.py
finbarrtimbers: yeah, that's fair. I'm trying to understand how we're doing evals generally. I think this is the wrong thing to measure.

hamishivi: Yeah fair its a bit unclear. There are sort of two eval loops that run (copying from a slack message i wrote recently):

1. in-loop evals, which just re-use the generation and reward/verifier code. I’m not crazy about these but they can be useful to observe how model generations change over training, and can give you an idea of val reward. Controlled by num_evals, which works out the total number of training steps via the episode count and then sets eval_freq accordingly. https://github.com/allenai/open-instruct/blob/main/open_instruct/grpo_fast.py#L1459. I’m actually not really a fan of this way of doing it but don’t feel strongly enough to change it haha.

2. oe-eval evals, which are downstream and launched as separate jobs, and the actual final numbers we usually care about. This is tied to save_freq AND requires you setting try_launch_beaker_eval_jobs_on_weka to True (even if on augusta, we should rename this arg). We could add a further check at e.g. https://github.com/allenai/open-instruct/blob/main/open_instruct/grpo_fast.py#L1785. Note that we can’t really untie this since the oe-eval jobs need some sort of path.

PR #788, opened 2025-07-15
Fixes the quality workflow

finbarrtimbers:
Right now, if the quality workflow finds fixes to all of the formatting issues, it returns exit code zero (i.e. a success).

But that's not actually success because it means that there are still changes to be done! 

This pull request checks that make style and make quality won't do anything and fails otherwise.

### DIFF
--- a/open_instruct/benchmark_generators.py
+++ b/open_instruct/benchmark_generators.py
@@ -257,21 +257,13 @@ def run_generation_batch(
     inference_results_Q: queue.Queue, param_prompt_Q: queue.Queue, prompts: list[list[int]]
 ) -> dict[str, Any]:
     """Run generation for a batch of prompts and measure performance."""
-    from open_instruct.vllm_utils3 import PromptRequest
-    
+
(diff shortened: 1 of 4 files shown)


open_instruct/benchmark_generators.py
```diff
@@ -257,13 +257,21 @@ def run_generation_batch(
     inference_results_Q: queue.Queue, param_prompt_Q: queue.Queue, prompts: list[list[int]]
 ) -> dict[str, Any]:
     """Run generation for a batch of prompts and measure performance."""
-
```
hamishivi: Are these changes intentional?

hamishivi (approved):
Seems good, not sure if you wanted to included the PromptRequest changes but they seem fine either way!

open_instruct/benchmark_generators.py
finbarrtimbers: Ugh, no. Removed. Thanks for the catch.

PR #793, opened 2025-07-16
fix Dockerfile.uv, install nginx in uv

mnoukhov:
it was only being installed in regular Dockerfile


saurabh111233212 (approved):
ooh nice, sorry abt that!

PR #794, opened 2025-07-17
Remove Unused DPO Function

fabianlim:
In https://github.com/allenai/open-instruct/pull/772 we have introduced an unused function (that only lives in our fork). 

For cleanliness sake this PR removes it.


open_instruct/dataset_transformation.py
```diff
@@ -978,41 +978,6 @@ def preference_tulu_tokenize_and_truncate_v1_2(
 def preference_tulu_filter_v1(row: Dict[str, Any], tokenizer: PreTrainedTokenizer):
     return any(x != -100 for x in row[CHOSEN_LABELS_KEY]) and any(x != -100 for x in row[REJECTED_LABELS_KEY])
```
hamishivi, line 980:
```suggestion


```

PR #795, opened 2025-07-17
Updates the benchmark script so that we save results to weka for easy data analysis

finbarrtimbers:
Otherwise, I was parsing the benchmark logs by hand.


saurabh111233212 (approved):
very nice!! ty

PR #796, opened 2025-07-17
Now, we run individual prompts through the queue.

finbarrtimbers:
Changes the queue setup so that we run individual prompts through the queues rather than batches. This will avoid the need for load balancing and should enable us to get some nice speed ups once we enable the async engine (next PR!).

Benchmark results! From HEAD: 

```
Results (excluding first batch):
Average tokens/second: 1014.84
Average MFU: 3.96%
Average generation time per batch: 2.02s
Average new tokens per sample: 2048.0 tokens
Wasted compute % (variable response length): 0.00%
```

From this PR:

```
Results (excluding first batch):
Average tokens/second: 614.95
Average MFU: 2.40%
Average generation time per batch: 1.67s
Average new tokens per sample: 1024.0 tokens
Wasted compute % (variable response length): 0.00%%
```


mnoukhov:
I'm confused why (1) the length of samples goes down and also (2) the time per batch goes down if the speed goes down too

finbarrtimbers:
> I'm confused why (1) the length of samples goes down and also (2) the time per batch goes down if the speed goes down too

@mnoukhov it's because I accidentally changed the batch size between runs! That's not the length, but the total number of tokens generated in the batch. Nice catch.

open_instruct/vllm_utils3.py
```diff
@@ -204,34 +205,63 @@ def process_from_queue(
...
+                eval_results = self._generate_batch_individual(eval_prompts, eval_sampling_params, None)
+                for eval_result in eval_results:
+                    eval_result.is_eval = True
+                    # Put eval results in separate queue if available
+                    if self.eval_results_queue is not None:
+                        self.eval_results_queue.put(eval_result)
+                    else:
+                        self.results_queue.put(eval_result)
 
     def _generate_batch(
```
mnoukhov, line 237: is this being deprecated? if so, maybe worth adding a comment about that

open_instruct/vllm_utils3.py
```diff
@@ -204,34 +205,63 @@ def process_from_queue(
...
-            # Process training prompts
-            result = self._generate_batch(request.prompts, sampling_params, request.dataset_index)
-            self.results_queue.put(result)
+            # Collect individual prompts to form a batch
+            prompts_batch = []
+            dataset_indices_batch = []
+            eval_prompts = None
+
+            # Pull prompts until we have a full batch or queue is empty
+            while len(prompts_batch) < (batch_size or 1):
```
mnoukhov: probably simpler to make `batch_size` have a default value of 1 instead of doing the `or 1` here

open_instruct/vllm_utils3.py
```diff
@@ -204,34 +205,63 @@ def process_from_queue(
...
-
-            # Process training prompts
-            result = self._generate_batch(request.prompts, sampling_params, request.dataset_index)
-            self.results_queue.put(result)
+            # Collect individual prompts to form a batch
+            prompts_batch = []
+            dataset_indices_batch = []
+            eval_prompts = None
+
+            # Pull prompts until we have a full batch or queue is empty
```
mnoukhov: Seems like this can have more prompts than batch_size (unless .get only gets one at a time?) 

I'm also confused when would you have a batch that is not full but due to timeout you will take what you get?

mnoukhov (commented):
Did a quick pass and seems good. Definitely a good precursor to async

open_instruct/vllm_utils3.py
finbarrtimbers: Ok, I have fixed this! Now, `PromptRequest` only contains a single prompt, and I have removed the timeout, so it will block until it gets a full batch.

open_instruct/vllm_utils3.py
finbarrtimbers: You're right, removed.

open_instruct/vllm_utils3.py
finbarrtimbers: I wasn't careful enough with Claude Code and let it add a new function instead of modifying `_generate_batch`. Thanks for catching that.

open_instruct/vllm_utils3.py
mnoukhov: Makes sense, you can remove the comment about `non-blocking`

open_instruct/grpo_fast.py
```diff
@@ -1209,7 +1201,10 @@ def data_preparation_thread(
         # Accumulate results from multiple vLLM engines into a single training batch
         with Timer("🚀 [Data Preparation Thread] Getting response ids"):
             result, queries, ground_truths, datasets = accumulate_inference_batches(
-                inference_results_Q, pending_queries_map, args.vllm_num_engines, training_step
+                inference_results_Q,
+                pending_queries_map,
+                args.num_unique_prompts_rollout * args.num_samples_per_prompt_rollout,
```
mnoukhov, line 1206: is it possible that at the end of an epoch or something you will have fewer than this number of results? or are we dropping the last batch

open_instruct/grpo_fast.py
```diff
@@ -2100,9 +2086,17 @@ def main(args: Args, tc: TokenizerConfig, model_config: ModelConfig, num_eval_sa
     reward_fn = make_reward_fn(args)
 
     # Start vLLM engines to process from queues
+    batch_size_per_engine = (
+        args.num_unique_prompts_rollout * args.num_samples_per_prompt_rollout
+    ) // args.vllm_num_engines
```
mnoukhov, line 2095: do we know if this is equally divisible? and is it alright if its not?

mnoukhov (approved):
other than the little questions about specific batch sizes and divisibility, seems good to me

open_instruct/grpo_fast.py
finbarrtimbers: It should be divisible; I changed the code to raise a `ValueError` if it isn't, and we can handle that then.

open_instruct/grpo_fast.py
finbarrtimbers: `ShufflingIterator` guarantees that they're all the same size: 

```
 class ShufflingIterator:
      def __init__(self, data: np.ndarray, batch_size: int, seed: Optional[int] = None):
          self.data = data.copy()
          self.batch_size = batch_size
          self.index = 0
          self.rng = np.random.default_rng(seed)
          self.rng.shuffle(self.data)

          # Ensure the effective dataset size is divisible by batch_size                                                  
          self.effective_size = len(self.data) - (len(self.data) % batch_size)

      def __iter__(self) -> Iterator[List[int]]:
          return self

      def __next__(self) -> List[int]:
          if self.index >= self.effective_size:
              self.index = 0
              self.rng.shuffle(self.data)

          end_index = self.index + self.batch_size
          batch = self.data[self.index : end_index].tolist()
          self.index = end_index

          return batch
```

PR #797, opened 2025-07-17
Improve mason.py by allowing to pass local models for GCP and bubble up dataset cache errors

mnoukhov:
1. We can pass local weka models to be uploaded to GCP for augusta (instead of just hf models)

The downside is the commit_hash = "None"

2. If there are errors when running our script to cache the dataset, these are bubbled up and the script shows those errors and we don't launch the job


saurabh111233212 (approved):
very nice QoL improvement: ty!
was just earlier this week when I avoided launching on augusta bc a model was on weka and I didn't feel like uploading to hf 🫢

PR #798, opened 2025-07-18
binary reward for code

saurabh111233212:


open_instruct/ground_truth_utils.py
```diff
@@ -830,7 +832,11 @@ def make_request():
             result = await asyncio.to_thread(make_request)
             passes = result["results"]
             pass_rate = sum(passes) / len(passes) if passes else 0.0
-            return VerificationResult(score=pass_rate)
+            if self.is_binary_reward:
```
hamishivi: maybe we could make the threshold configurable?

saurabh111233212: yeah good idea i think the code block would look like 
```python
if pass_rate >= self.pass_rate_reward_threshold
    score = pass_rate
else:
    score = 0.0
```

PR #800, opened 2025-07-18
Adds flashinfer dep.

finbarrtimbers:
With this, we can use flashinfer as the backend for vllm. To do so, set the attention backend environment variable: 

```
export VLLM_ATTENTION_BACKEND=FLASHINFER
```

In a mason script, add `--env VLLM_ATTENTION_BACKEND=FLASHINFER` before the `--`.

In my benchmarking, this leads to a 15% speedup. [Benchmark at HEAD](https://beaker.allen.ai/orgs/ai2/workspaces/oe-eval/work/01K0DP2BEJYX5B5SQ6XNX6J8AD?taskId=01K0DP2BERNT8TQ53RY36NBYPB&jobId=01K0EJGKNWEYY3NJZR209674MT): 

```
Results (excluding first batch):
Average tokens/second: 3084.47
Average MFU: 4.41%
Average generation time per batch: 1105.47s
Average new tokens per sample: 3409088.75 tokens
Wasted compute % (variable response length): 34.84%
```

With this change: 

```
Results (excluding first batch):
Average tokens/second: 3500.95
Average MFU: 5.00%
Average generation time per batch: 1072.42s
Average new tokens per sample: 3748289.0 tokens
Wasted compute % (variable response length): 28.36%
```

(note that this got a shorter generation time per batch despite generating more tokens, on average: 3.4M vs 3.7M.)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,7 +55,7 @@ explicit = true
 
 # flash-attn related setups
 [project.optional-dependencies]
-compile = ["flash-attn>=2.8.0.post1",
+compile = ["flash-attn>=2.8.0.post2",
 		   "flashinfer-python>=0.2.7.post1"]
 liger = ["liger-kernel>=0.5.4"]
 code = [
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ pip install --upgrade pip "setuptools<70.0.0" wheel
 # TODO, unpin setuptools when this issue in flash attention is resolved
 pip install torch==2.6.0 torchvision==0.21.0 --index-url https://download.pytorch.org/whl/cu124
 pip install packaging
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -55,7 +55,8 @@ explicit = true
 
 # flash-attn related setups
 [project.optional-dependencies]
-compile = ["flash-attn>=2.8.0.post1"]
+compile = ["flash-attn>=2.8.0.post1",
```
mnoukhov: I know its not exactly related but can we fix the flash-attn to be post2 like it is in the dockerfile

mnoukhov (approved):

pyproject.toml
finbarrtimbers: Yeah I can do that!

PR #805, opened 2025-07-21
perf penalty

saurabh111233212:
simple perf penalty


open_instruct/code/code_utils.py
```diff
@@ -30,61 +36,56 @@
...
-    """Run a program against a list of tests, if the program exited successfully then we consider
-    the test to be passed. Note that you SHOULD ONLY RUN THIS FUNCTION IN A VIRTUAL ENVIRONMENT
-    as we do not gurantee the safety of the program provided.
+def encode_tests(tests: list) -> str:
+    if not tests:
+        return ""
+    pickled_data = pickle.dumps(tests)
+    compressed_data = zlib.compress(pickled_data)
+    b64_encoded_data = base64.b64encode(compressed_data)
+    return b64_encoded_data.decode("utf-8")
```
hamishivi, line 49: Is this to make sure the tests are uploaded to the server quickly?

open_instruct/code/api.py
```diff
@@ -38,7 +38,7 @@
 
 class TestRequest(BaseModel):
     program: str
-    tests: List[Any]
+    tests: Any
     max_execution_time: float = 1.0
```
hamishivi, line 42: Is 1.0 a reasonable default here?

hamishivi (approved):
Seems good assuming you've got a test run. Had some clarification questions

open_instruct/code/code_utils.py
saurabh111233212: yeah needed to compress the tests when training on LiveCodeBench, those long tests get long! The decoding logic handles tests in all formats so should be fine. 

also will be needed as we scaled up length of our real test-casess

open_instruct/code/api.py
saurabh111233212: I think so. LiveCodeBench uses 6.0, but they have much crazier test cases. 
We can revisit when we get there, but I think this is more than enough. Could probably be smaller

PR #806, opened 2025-07-21
Provide an option to truncate samples longer than max_seq_length

dangxuanhong:
- In SFTraining, we may want to retain only complete conversations/messages. This PR adds support for filtering out truncated samples (i.e., when their conversations/messages are longer than max_seq_length after tokenization). 
- This PR also logs the number of samples removed after applying the sequence of data transformation steps specified via `--dataset_transform_fn`. It also logs token counts for each dataset. These stats. are useful for monitoring and debugging the training process.


dangxuanhong:
(Closed due to submitting to a wrong main branch)

PR #814, opened 2025-07-22
Fix disk space error on image creation

hamishivi:
We are running out of space on the github runner when building the docker image, hopefully this fixes this.


finbarrtimbers (approved):
This looks good, feel free to go ahead with it. I do wonder though if there's a way to avoid creating the offending dir.

hamishivi:
I think they're just default on the github runners, not sure if theres a more minimal runner available...

PR #819, opened 2025-07-23
fix truncating

garrett361:
Small fixes for #33 to ensure it doesn't break DPO.


garrett361:
made against the wrong branch, again.

PR #821, opened 2025-07-24
Fix head so grpo_fast.py can run.

finbarrtimbers:
Fixes head so that experiments run. 

Runs:
- [single GPU script](https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01K18VFK0DBRS53YG62DN9P0E6?)
- [Saurabh's experiment](https://beaker.allen.ai/orgs/ai2/workspaces/oe-adapt-code/work/01K18VCDE60ZZY1H096ZRSCWR4/logs?jobId=01K18VCDKHMPQ76VHKB2Q1Z4PN)

There were a bunch of minor bugs that weren't caught because the script I was running to test the changes didn't rebuild the docker image, as I didn't understand that was needed. Those have been fixed. 

Additionally, we were accidentally double inserting into the queues, as we were [inserting the original batch after creating it](https://github.com/allenai/open-instruct/blob/main/open_instruct/grpo_fast.py#L2158), and then doing so again in [sync_weights_and_prepare_prompts](https://github.com/allenai/open-instruct/blob/main/open_instruct/grpo_fast.py#L1736). This has been fixed, making the logic in `sync_weights_and_prepare_prompts` much simpler.

The bigger problem is the way we were syncing the weights. Previously we would split the queries into batches and send each to a specific engine, calling `engine.generate.remote(queries)`. Later on, we'd sync the weights, in `broadcast_to_vllm`. But when I switched to use the queue, we had the actors running a while loop that pulls off the queue the whole time, doing `engine.generate.process_from_queue(prompt_q, results_q)`. As this never yielded control, that would deadlock. 

As a result, this PR changes `process_from_queue` to exit when the queue is done, and then will sync the weights synchronously. It should make the changes easy to switch to insert individual prompts and then to run the async engine.


open_instruct/grpo_fast.py
```diff
@@ -2069,11 +2187,11 @@ def main(args: Args, tc: TokenizerConfig, model_config: ModelConfig, num_eval_sa
...
 
     # Initialize Ray before creating Ray objects
-    ray.init(dashboard_host="0.0.0.0")  # enable debugging from a different machine (e.g., phobos)
+    ray.init(dashboard_host="0.0.0.0")
 
-    # Create Ray queues
-    inference_results_Q = ray_queue.Queue(maxsize=args.async_steps)
-    param_prompt_Q = ray_queue.Queue(maxsize=args.async_steps)
+    # Create Ray queues.
+    inference_results_Q = ray_queue.Queue(maxsize=args.async_steps + 1)
```
hamishivi: Should it be +1? I thought we only need `async_steps` number of slots in the queue.

.github/workflows/tests.yml
```diff
@@ -24,4 +24,4 @@ jobs:
       - name: Set up Python
         run: uv python install
       - name: Run unit tests
-        run: uv run pytest
\ No newline at end of file
+        run: uv run pytest -vs
```
hamishivi: intentional change?

open_instruct/grpo_fast.py
```diff
@@ -142,6 +144,134 @@
 INVALID_LOGPROB = 1.0
 
 
+# ---- Runtime leak detection -----------------------------------------------------------------
```
hamishivi: Nit: could we move this into a utils file?

run_hamish_repro.sh
```diff
@@ -0,0 +1,64 @@
+uv run python mason.py \
```
hamishivi: Intended to commit these scripts? Please move them into a scripts folder if so!

hamishivi (commented):
I ran this in a debug session and it ran fine for me, I think there's just a little cleanup needed before it's good to merge, assuming we have a succesful larger-scale test.

.github/workflows/tests.yml
finbarrtimbers: Debugging. But I think we should change this anyway. Will send you a PR.

open_instruct/grpo_fast.py
finbarrtimbers: Yes! Done.

run_hamish_repro.sh
finbarrtimbers: Removed!

open_instruct/grpo_fast.py
```diff
@@ -230,10 +231,10 @@ class Args:
     The returned output will not contain the stop strings."""
 
     # Algorithm
-    async_mode: bool = True
+    async_mode: Optional[bool] = None
```
hamishivi: If we are going to deprecate this, let's just remove it and add the description to the async_steps arg?

open_instruct/grpo_fast.py
```diff
@@ -1109,42 +1114,124 @@ def get_bundle_index(rank, num_gpus_per_node):
...
-    # Collect results from all engines
+    # Collect results from all engines with non-blocking progress bar
     results = []
     all_queries = []
     all_ground_truths = []
     all_datasets = []
 
-    for batch_idx in range(vllm_num_engines):
-        # Get result from queue
+    for i in tqdm(
```
hamishivi, line 1203: nit: Could we add back in the total generation time timer too? It can be useful to keyword search it in the logs for checking if we should allocate more/less compute to inference.

finbarrtimbers: Oh, yes, absolutely.

open_instruct/grpo_fast.py
finbarrtimbers: Sounds good!

PR #823, opened 2025-07-24
Silences TORCH_CUDA_ARCH_LIST warning by passing the env var through.

finbarrtimbers:
Before this PR, we saw lots of annoying warnings like this: 

```
2025-07-24T13:32:07.377Z (LLMRayActor pid=21608) /stage/.venv/lib/python3.12/site-packages/torch/utils/cpp_extension.py:2356: UserWarning: TORCH_CUDA_ARCH_LIST is not set, all archs for visible cards are included for compilation. 
2025-07-24T13:32:07.377Z (LLMRayActor pid=21608) If this is not desired, please set os.environ['TORCH_CUDA_ARCH_LIST'].
```
Now they're gone, as we set `TORCH_CUDA_ARCH_LIST` through to Ray. We are able to guarantee the correct value for `TORCH_CUDA_ARCH_LIST` as the main process loads torch.

[Example successful Beaker run without logs](https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01K1BCJ07WJ994BQB2SNM85FZ4?taskId=01K1BCJ083ZTSAB5NEDVBXBW91&jobId=01K1BCJ0BQ89BGB2R3A7B5F6X6).

### DIFF
--- a/open_instruct/vllm_utils3.py
+++ b/open_instruct/vllm_utils3.py
@@ -302,6 +302,23 @@ class LLMRayActor:
         return True
 
 
+def get_cuda_arch_list() -> str:
+    """Get CUDA compute capabilities and format them for TORCH_CUDA_ARCH_LIST."""
+    if not torch.cuda.is_available():
+        return ""
+    
+    cuda_capabilities = []
+    for i in range(torch.cuda.device_count()):
+        major, minor = torch.cuda.get_device_capability(i)
+        cuda_capabilities.append(f"{major}.{minor}")
+    
+    # Remove duplicates and sort
+    cuda_capabilities = sorted(set(cuda_capabilities))
+    cuda_arch_list = ";".join(cuda_capabilities)
+    print(f"Detected CUDA compute capabilities: {cuda_capabilities}, setting TORCH_CUDA_ARCH_LIST={cuda_arch_list}")
+    return cuda_arch_list
+
+
 def create_vllm_engines(
     num_engines: int,
     tensor_parallel_size: int,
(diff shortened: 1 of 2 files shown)


scripts/train/debug/single_gpu_on_beaker.sh
```diff
@@ -2,7 +2,7 @@
 
 # Get the Beaker username to construct the image name
 BEAKER_USER=$(beaker account whoami --format json | jq -r '.[0].name')
-BEAKER_IMAGE="${BEAKER_USER}/open-instruct-integration-test"
+BEAKER_IMAGE="${1:-${BEAKER_USER}/open-instruct-dev}"
```
hamishivi: nit: I sort of like calling it integration-test? Sometimes I use -dev. Or even the commit hash in the name, something that is unlikely to override/clash with an existing image.

open_instruct/vllm_utils3.py
```diff
@@ -377,7 +394,9 @@ def create_vllm_engines(
                 num_gpus=num_gpus,
                 scheduling_strategy=scheduling_strategy,
                 # VLLM v1 multiprocessing is required due to https://github.com/vllm-project/vllm/issues/15349
-                runtime_env=ray.runtime_env.RuntimeEnv(env_vars={"VLLM_ENABLE_V1_MULTIPROCESSING": "0"}),
+                runtime_env=ray.runtime_env.RuntimeEnv(
+                    env_vars={"VLLM_ENABLE_V1_MULTIPROCESSING": "0", "TORCH_CUDA_ARCH_LIST": get_cuda_arch_list()}
```
hamishivi, line 398: Is there a reason we might want to unset `TORCH_CUDA_ARCH_LIST`? Should we give users the option to?

hamishivi (approved):
I think this should be okay, although I wonder if we should make it configurable. E.g. for interactives, sometimes you end up using diff gpus for testing (L40 vs A100 vs H100), so there it maybe this would cause an issue? Not entirely sure tho.

scripts/train/debug/single_gpu_on_beaker.sh
finbarrtimbers: I'm happy to keep it then!

open_instruct/vllm_utils3.py
finbarrtimbers: If we unset it, then torch will generate code for all architectures that are supported by the currently visible devices.

That is desirable if there is a mismatch between the main process and the target one in terms of device. We currently do not support that heterogeneity in workload afaik, so I don't think there is a reason to unset it.

hamishivi: Yeah, I guess I was just thinking about the case where you initially install your dev build on a machine with one gpu type and then want to run it on some different device (e.g. I installed on jupiter, but I can only get some gpus on neptune). Then having the ability to unset this, or we only set it in e.g. mason (which isnt used for interactives) makes more sense.

But probably we can see if people hit this issue!

PR #825, opened 2025-07-24
look at changes

jacob-morrison:


open_instruct/dataset_transformation.py
```diff
@@ -1787,9 +1787,13 @@ def get_cached_dataset_tulu_with_statistics(
         )
     elif dataset_cache_mode == "hf":
         cache = DatasetTransformationCache(config_hash=dataset_config_hash, hf_entity=hf_entity)
-    return cache.load_or_transform_dataset(
-        dcs, tc, dataset_skip_cache=dataset_skip_cache, return_statistics=return_statistics
-    )
+
+    dataset, statistics = cache.load_or_transform_dataset(dcs, tc, dataset_skip_cache=dataset_skip_cache)
+
+    if return_statistics:
```
hamishivi: I don't really like mutable returns with python functions, because it makes debugging what a function is doing down the line much harder. Can we instead keep the `get_cached_dataset_tulu` as before, and then the statistics script just uses the sub-function that also returns the stats?

That or we set it so wherever the dataset function is called, we also take the statistics out too (which may be none).

open_instruct/dataset_transformation.py
```diff
@@ -1818,7 +1822,7 @@ def get_cached_dataset_tulu(
         dataset_local_cache_dir,
         dataset_skip_cache,
         return_statistics=False,
-    )[0]
+    )
```
hamishivi: see comment above.

hamishivi (changes requested):
The removing logic seems good, but less sure on these mutable returns.

open_instruct/dataset_transformation.py
jacob-morrison: That's fine with me! updated

jacob-morrison:
test GRPO job to make sure everything's good to go: https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01K0ZDKD1N3ABZ9JMJ70HE8WD6?taskId=01K0ZDKD1VQATYD8SW6XCDKKBR&jobId=01K0ZDKD7Y68RMRZR2SKNA0BHA

hamishivi:
Seemed like the job errored, also now there's a merge conflict?
I think the error is probably unrelated to your change, but it'd be good to do the run in a pure_docker_mode setup with an image corresponding to this code if possible, thanks!

PR #830, opened 2025-07-25
Adds an integration test, which launches a beaker experiment that runs `grpo_fast` on a single GPU.

finbarrtimbers:
The integration test is triggered when:

1. After a PR is merged (to avoid blocking, as it can take ~10 minutes to run).
2. Every 24 hours (assuming there have been changes to the repo).

The goal here is to make it hard/impossible to break the repo without anyone knowing.

### DIFF
--- a/open_instruct/code/api.py
+++ b/open_instruct/code/api.py
@@ -19,6 +19,7 @@ or
 
 curl -X GET http://localhost:1234/health
 curl -X POST http://localhost:1234/test_program -H "Content-Type: application/json" -d '{"program": "def add(a, b): return a + b", "tests": ["assert add(1, 2) == 3", "assert add(-1, 1) == 0", "assert add(0, 0) == 1"], "max_execution_time": 1.0}'
+curl -X POST http://localhost:1234/test_program_stdio -H "Content-Type: application/json" -d '{"program": "import sys\nfor line in sys.stdin.read().splitlines():\n    print(int(line.strip()) + 1)", "tests": [{"input": "1\n", "output": "2\n"}, {"input": "100\n", "output": "101\n"}], "max_execution_time": 1.0}'
 """
 
 import logging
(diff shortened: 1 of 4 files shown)


.github/workflows/beaker-experiment.yml
```diff
@@ -0,0 +1,216 @@
+name: Beaker Experiment Launch
+
+on:
```
hamishivi, line 3: Can we test that this PR works and temporarily add back running on the PR? I clicked through the commits on the PR and didn't see a working test run...

edit: although I guess for this to happen we need to get main fixed first!

pyproject.toml
```diff
@@ -87,6 +87,7 @@ dev = [
     "pytest>=8.3.4",
     "ruff>=0.11.13",
 	"parameterized>=0.9.0",
```
hamishivi: nit: indentation looks weird in github ui?

scripts/train/build_image_and_launch.sh
```diff
@@ -0,0 +1,36 @@
...
+set -e
+
+image_name=open-instruct-integration-test
+
+# Build the Docker image exactly like push-image.yml does
+docker build \
+    --build-arg BUILDKIT_INLINE_CACHE=1 \
+    --build-arg CUDA=12.1.0 \
+    --build-arg TARGET=cudnn8-devel \
+    --build-arg DIST=ubuntu20.04 \
```
hamishivi, line 11: Oop, I think the CUDA/TARGET/DIST args aren't needed anymore after I updated the dockerfile. Don't need to change in this pr but we should

finbarrtimbers: Sounds good.

pyproject.toml
finbarrtimbers: Fixed! Nice catch.

hamishivi (approved):
Approved since it seems the test is working. Remember to comment out the bit to trigger on each PR! (dont remove entirely in case we want to test out a PR with this test in the future)

.github/workflows/beaker-experiment.yml
finbarrtimbers: Done!

finbarrtimbers:
> Approved since it seems the test is working. Remember to comment out the bit to trigger on each PR! (dont remove entirely in case we want to test out a PR with this test in the future)

Done!

PR #831, opened 2025-07-26
remove_non_tensor_columns

garrett361:
In https://github.com/allenai/open-instruct/pull/765 additional string metadata was added to the dataset which conflicts with the application of the `DataCollatorForSeq2Seq` collator function, which expects only tensor data.   This makes `finetune.py` fail (when `--packing False`) with errors like:
```
Traceback (most recent call last):
  File "/proj/data-eng/swanand/sft_dpo/venv_sft_dpo/venv/venv_sft_dpo/lib/python3.10/site-packages/transformers/tokenization_utils_base.py", line 767, in convert_to_tensors
    tensor = as_tensor(value)
  File "/proj/data-eng/swanand/sft_dpo/venv_sft_dpo/venv/venv_sft_dpo/lib/python3.10/site-packages/transformers/tokenization_utils_base.py", line 729, in as_tensor
    return torch.tensor(value)
ValueError: too many dimensions 'str'
```
This error arises when trying to create a tensor from a list of strings, e.g. `torch.tensor(["hello"])`. 

This PR adds a utility for filtering non-tensor columns out of the dataset before using and uses the filter in both the sft and dpo scripts.

CC @hamishivi @jacob-morrison 

The issue specifically is the addition of the `DATASET_ORIGIN_KEY` [metadata here](https://github.com/garrett361/open-instruct/blob/a730b24f0c5929a69f3bb5116343e405fe4f76df/open_instruct/dataset_transformation.py?plain=1#L1459-L1463). 

I believe this wrapper is only needed for `finetune.py` and `dpo_cache_tune.py`, and not needed for `reward_modeling.py` or `reward_modeling_eval.py`, but have only tested `finetune.py` with these fixed e2e.

### DIFF
--- a/open_instruct/dataset_transformation.py
+++ b/open_instruct/dataset_transformation.py
@@ -44,6 +44,7 @@ import copy
 import hashlib
 import json
 import multiprocessing
+import warnings
 import os
 from dataclasses import asdict, dataclass, field
 from functools import cached_property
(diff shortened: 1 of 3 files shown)


hamishivi:
Thanks for noticing and the PR, I believe this is fixed by #825 (which I just merged).

garrett361:
Yes, hadn't seen that one! Thanks.

PR #837, opened 2025-07-29
Switches LLMRayActor to use the LLMEngine instead of LLM. 

finbarrtimbers:
This change by itself should not change the observed behaviour of the system at all, but it makes subsequent refactoring to have finer grained control over updates and to support continual processing of prompts much easier.

The [implementation](https://github.com/vllm-project/vllm/blob/755fa8b657e3666cc93b08a6d2b9a50d0f46c37e/vllm/entrypoints/llm.py#L1634
) of `vllm.LLM.generate` uses a python while loop around `engine.step()`, basically doing:

```
def generate(requests):
    for request in requests:
        self.llm_engine.add_request(request)
    while self.llm_engine.has_unfinished_requests():
        step_outputs = self.llm_engine.step()
        for output in step_outputs:
            if output.finished:
                outputs.append(output)
    return outputs
```

Here, we change our usage to do basically that (with a check for a weight update). 

The actor controls when we do weight updates, [allowing for inflight weight updates, like pipelinerl](https://huggingface.co/blog/ServiceNow/pipelinerl).

Runs:

1. Single GPU GRPO: [beaker](https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01K2QSV6MAEB6D2PND78R43GB8?taskId=01K2QSV6MD9W2RJ25YXWRA12AZ&jobId=01K2QSV6R68A3NTFDAJKMEQC9Q).

2. Multi-node GRPO: [beaker](https://beaker.org/ex/01K2QV330F4041NXA178JH4M4Q), [wandb](https://wandb.ai/ai2-llm/open_instruct_internal/runs/xhlk40uv).

3. Single GPU GRPO w/ tools: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01K2QHY91TBYHVTEG3GFWJQ0FG?taskId=01K2QHY92165PQG6P6XQ5PE9AP&jobId=01K2QHY95MFTPE5KRXV641DAC8), [Wandb](https://wandb.ai/ai2-llm/open_instruct_internal/runs/1sl72w03?nw=nwuserfinbarrt).

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -1988,11 +1988,12 @@ def generate_thread(args, vllm_engines, resume_training_step, stop_event, genera
             num_processed = sum(int(result) for result in processed_results)
             # Suppress timing output if nothing was processed
             if num_processed == 0:
-                timer.noop = 1
-        try:
-            generate_metrics_Q.put_nowait({"time/generation": timer.duration})
-        except Full:
-            logger.warning("[Generate Thread] generate metrics queue full, skipping metric")
+                timer.noop = True
+        if num_processed > 0:
+            try:
+                generate_metrics_Q.put_nowait({"time/generation": timer.duration})
+            except Full:
+                logger.warning("[Generate Thread] generate metrics queue full, skipping metric")
     logger.info("[Generate Thread] 🛑 Stopping generation thread")
 
 


open_instruct/vllm_utils3.py
```diff
@@ -190,55 +205,128 @@ def __init__(
             print(f"creating LLM with bundle_indices={bundle_indices}")
 
         if tool_use:
+            # TODO: Need to update ToolUseLLM to use LLMEngine as well
```
hamishivi: Could we make this change in this PR? I can give some example script to test.

open_instruct/grpo_fast.py
```diff
@@ -1807,19 +1811,24 @@ def create_model_and_optimizer(
...
         top_p=args.vllm_top_p,  # prevent rare out-of-vocab tokens with qwen
         max_tokens=args.response_length,
         include_stop_str_in_output=True,
         skip_special_tokens=False,
         n=args.num_samples_per_prompt_rollout,
         stop=args.stop_strings,
+        # IMPORTANT: Set output_kind to FINAL_ONLY to ensure vLLM V1 properly handles n>1
+        # With the default CUMULATIVE mode, vLLM V1 returns separate outputs for each
+        # completion, making it difficult to aggregate them correctly. FINAL_ONLY mode
+        # ensures all n completions are returned together in a single output.
```
hamishivi, line 1872: Is there some documentation link for this?

open_instruct/vllm_utils3.py
```diff
@@ -190,55 +205,128 @@ def __init__(
...
-            return 1  # Successfully processed a request
-        except queue.Empty:
-            self.logger.warning("[LLMRayActor] No request in the queue to process. Returning from process_from_queue.")
-            return 0  # No request to process
-        except queue.Full:
-            self.logger.warning(f"[LLMRayActor] Results queue is full. Skipping insert. {request.is_eval=}.")
-            return 0
-
-    def _generate_batch(
+        if self.tool_use:
```
hamishivi: I feel like a nicer way to do this might just to have the toolvllm subclass / be an alternate class to llmrayactor and override the generation loop?

open_instruct/vllm_utils3.py
```diff
@@ -190,55 +205,128 @@ def __init__(
...
+                    if output.finished:
+                        outputs.append(output)
+
+            # Sort outputs by prompt index (i)
+            # Request IDs are now in format: batch_{step}_{i}
+            outputs.sort(key=lambda x: int(x.request_id.split("_")[-1]))
+
+            # Debug logging
+            total_responses = sum(len(output.outputs) for output in outputs)
+            self.logger.info(
```
hamishivi: Similar to past PRs, I wonder if we can make this logging explicitly debug level and not log by default, since it seems like a lot.

hamishivi (commented):
Generally I think this looks good, but I think we should try to keep the toolvLLM compatibility going now rather than later. Also, both your job links link to jobs that failed?

For the tool stuff, test by:
- running `python open_instruct/tool_utils/tool_vllm.py` directly.
- running a script with tool use (I'll try to update the comment with one by EOD)

open_instruct/grpo_fast.py
finbarrtimbers: Not that I can find. There is this: 

https://github.com/vllm-project/vllm/blob/e789cad6b8b5d2a01aa6521b9208bb8d6501ee5b/vllm/entrypoints/llm.py#L1651

Should I add a link?

open_instruct/vllm_utils3.py
finbarrtimbers: Much nicer! I'll do that.

open_instruct/vllm_utils3.py
finbarrtimbers: I just removed this. Not worth including.

open_instruct/grpo_fast.py
hamishivi: ah... I guess I just worry how much we have to edit this if vllm changes in the future and how tricky it might be to track these sorts of things. I feel like the link isnt thaaat informative for this.

open_instruct/grpo_fast.py
```diff
@@ -2261,17 +2264,22 @@ def check_threads_healthy(
...
-    # Push sentinel to signal data_preparation_thread to stop
+    logger.info("Signaling all actors to stop...")
+    ray.get(actor_manager.set_should_stop.remote(True))
+    logger.info("✅ Signaled all actors to stop")
+
     logger.info("Pushing shutdown sentinel to queues...")
-    if queues and len(queues) > 0:
-        # Push sentinel to the first queue (inference_results_Q)
-        queues[0].put(ShutdownSentinel(), timeout=1)
+    # Push sentinel to the first queue (inference_results_Q)
```
hamishivi, line 2351: why remove the queues check?

open_instruct/tool_utils/tool_vllm.py
```diff
@@ -131,404 +120,73 @@ def find_sum_of_a():
...
-            if real_req_id not in merged_outputs:
-                merged_outputs[real_req_id] = concat_outputs[req_id]
-            else:
-                merged_outputs[real_req_id].outputs.append(concat_outputs[req_id].outputs[0])
-        final_outputs = sorted(
-            merged_outputs.values(), key=lambda x: (int(x.request_id.split("-")[0]), int(x.request_id.split("-")[1]))
+    def generate(self, prompts=None, sampling_params=None, prompt_token_ids=None, use_tqdm=True):
+        """Simple generate method that processes tools synchronously after generation."""
+        # Use the underlying vLLM generate
+        outputs = self.llm.generate(
```
hamishivi: Sometimes we get multiple tool calls in a generation + want to continue generating after the tool call, I don't think that happens here?

open_instruct/grpo_fast.py
```diff
@@ -1907,32 +1923,16 @@ def load_data_from_packing_thread(packed_sequences_Q: Queue, num_total_tokens: i
...
-                desc="[Generate Thread] Waiting for vLLM engines to return",
-                bar_format="{l_bar}{bar}{r_bar}\n",
-                disable=not args.verbose,
-            ) as pbar:
-                for future in futures.as_completed(engine_futures):
-                    processed_results.append(future.result())
-                    pbar.update(1)
+            processed_results = ray_get_with_progress(
+                [engine.process_from_queue.remote(timeout=20) for engine in vllm_engines],
+                desc="[Generate Thread] Waiting for vLLM engines to process",
```
hamishivi, line 1985: This is quite noisy, can we make this verbose only?

hamishivi (commented):
Left some comments. Also pushed a test script to `debug/tool_grpo_fast.sh` for testing tool vllm integrations! Currently using the tool integration seems to hang. Happy to spend some time debugging it tomorrow.

open_instruct/grpo_fast.py
finbarrtimbers: Yeah I share your concern. Not sure what the best answer is.

open_instruct/grpo_fast.py
finbarrtimbers: Done!

open_instruct/grpo_fast.py
finbarrtimbers: I think it's redundant! Happy to add it back

open_instruct/grpo_fast.py
```diff
@@ -1958,42 +1975,25 @@ def load_data_from_packing_thread(packed_sequences_Q: Queue, num_total_tokens: i
...
-            time.sleep(1)
-        else:
-            try:
-                generate_metrics_Q.put_nowait({"time/generation": timer.duration})
-            except Full:
-                logging.warning("[Generate Thread] generate metrics queue full, skipping metric")
-
+            else:
+                try:
+                    generate_metrics_Q.put_nowait({"time/generation": timer.duration})
```
hamishivi: `duration` actually doesn't exist until we exit the timer context, so the indentation here breaks the code. Could do `timer.start_time - time.perf_counter()` or just put this outside the context?

open_instruct/grpo_fast.py
```diff
@@ -1923,24 +1932,32 @@ def sync_weights_and_prepare_prompts(
...
         ray_get_with_progress(
-            ray_refs,
-            desc=f"[Main Thread] Broadcasting weights to vLLM engines at training step {training_step}",
+            [m.broadcast_to_vllm.remote() for m in policy_group.models],
+            desc=f"[Main thread] Broadcasting weights to vLLM engines at training step {training_step}",
             enable=args.verbose,
         )
 
+        ray.get(actor_manager.set_should_stop.remote(False))
+        logger.debug(f"[Main Thread] Set should_stop to False after weight sync at step {training_step}")
```
hamishivi, line 1945: tiniest nit in the world: `Main Thread` vs `Main thread` naming is not consistent (capital on 't').

open_instruct/grpo_fast.py
```diff
@@ -1810,8 +1810,10 @@ def create_model_and_optimizer(
             else:
                 raise ValueError(f"Unknown tool: {tool}")
 
+    actor_manager = vllm_utils3.ActorManager.remote()
```
hamishivi, line 1813: Is the ActorManager giving us anything the stop_event isn't? Maybe we could combine the two to make stopping signalling simpler?

scripts/train/debug/tool_grpo_fast.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
```
hamishivi, line 1: Could we also add a debug script without mason that runs directly on 1 GPU? This should work:
```bash
uv run open_instruct/grpo_fast.py \
    --dataset_mixer_list hamishivi/tulu_3_rewritten_100k_with_tool_prompt 64 \
    --dataset_mixer_list_splits train \
    --dataset_mixer_eval_list hamishivi/tulu_3_rewritten_100k_with_tool_prompt 16 \
    --dataset_mixer_eval_list_splits train \
    --max_token_length 512 \
    --max_prompt_token_length 512 \
    --response_length 512 \
    --pack_length 1024 \
    --per_device_train_batch_size 1 \
    --num_unique_prompts_rollout 16 \
    --num_samples_per_prompt_rollout 4 \
    --model_name_or_path Qwen/Qwen3-1.7B \
    --apply_verifiable_reward true \
    --temperature 0.7 \
    --ground_truths_key ground_truth \
    --chat_template_name r1_simple_chat_postpend_think_tools \
    --learning_rate 3e-7 \
    --total_episodes 200 \
    --deepspeed_stage 2 \
    --num_epochs 1 \
    --num_learners_per_node 1 \
    --vllm_tensor_parallel_size 1 \
    --beta 0.01 \
    --seed 3 \
    --local_eval_every 1 \
    --vllm_sync_backend gloo \
    --single_gpu_mode \
    --vllm_gpu_memory_utilization 0.3 \
    --save_traces \
    --vllm_enforce_eager \
    --gradient_checkpointing \
    --tools search code \
    --search_api_endpoint "http://saturn-cs-aus-232.reviz.ai2.in:44177/search" \
    --code_tool_api_endpoint https://open-instruct-tool-server-10554368204.us-central1.run.app/execute \
    --push_to_hub false \
    # --with_tracking

```

open_instruct/grpo_fast.py
finbarrtimbers: Ah, that was a merge error. Fixed.

open_instruct/grpo_fast.py
finbarrtimbers: Good nit! Fixed.

scripts/train/debug/tool_grpo_fast.sh
finbarrtimbers: Done.

open_instruct/grpo_fast.py
finbarrtimbers: Resolved on Slack.

scripts/train/debug/tool_grpo_fast.sh
finbarrtimbers: And this succeeds!

PR #839, opened 2025-07-29
if there's an error in generate thread, bubble up to main to kill

mnoukhov:
use threading stop event to check whether generate thread is dead

checks for exception in futures, I think that's the right approach

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -1881,15 +1881,15 @@ def generate_thread(
             ) as pbar:
                 for future in futures.as_completed(engine_futures):
                     try:
-                        if future.exception() is not None:
-                            raise future.exception()
+                        # if future.exception() is not None:
+                        #     raise future.exception()
                         processed_results.append(future.result())
-                    except Exception as e:
-                        logger.error(f"[Generate Thread] Error processing from queue: {e}")
-                        stop_event.set()
-                        break
-                    finally:
                         pbar.update(1)
+                    except Exception as e:
+                        logger.error("[Generate Thread] Error processing from queue")
+                        stop_event.set()
+                        raise e
+
             num_processed = sum(int(result) for result in processed_results)
             logger.info(
                 f"[Generate Thread] 🚀 Processed results from all vLLM engines ({num_processed}/{len(processed_results)} processed batches)"
@@ -2366,17 +2366,16 @@ def main(args: Args, tc: TokenizerConfig, model_config: ModelConfig, num_eval_sa
 
         # The generate_thread is now handling vLLM processing asynchronously
 
-        try:
-            collated_data, data_thread_metrics, num_total_tokens = load_data_from_packing_thread(
-                packed_sequences_Q, num_total_tokens, stop_event
-            )
-        except Exception as e:
+        collated_data, data_thread_metrics, num_total_tokens = load_data_from_packing_thread(
+            packed_sequences_Q, num_total_tokens, stop_event
+        )
+
+        if stop_event.is_set():
             cleanup_training_resources(
                 stop_event,
                 [packing_thread, generation_thread],
                 [inference_results_Q, param_prompt_Q, evaluation_inference_results_Q],
             )
-            raise Exception(e)
 
         # Ai2 logic: we use /output to store the artifacts of the job, so we
         if collated_data is None:
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -1826,9 +1826,25 @@ def sync_weights_and_prepare_prompts(
     return queries_next, ground_truths_next, datasets_next, dataset_indices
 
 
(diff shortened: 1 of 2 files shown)


open_instruct/grpo_fast.py
```diff
@@ -1826,9 +1826,25 @@ def sync_weights_and_prepare_prompts(
     return queries_next, ground_truths_next, datasets_next, dataset_indices
 
 
-def load_data_from_packing_thread(packed_sequences_Q: Queue, num_total_tokens: int):
+def load_data_from_packing_thread(
+    packed_sequences_Q: Queue, num_total_tokens: int, stop_generate_event: threading.Event
```
finbarrtimbers: If we're passing stop_generate_event through to multiple threads, let's just call it `stop_event`?

open_instruct/grpo_fast.py
```diff
@@ -1826,9 +1826,25 @@ def sync_weights_and_prepare_prompts(
...
+                if packed_data is not None:
+                    # got data continue
+                    break
+            except Empty:
+                # keep polling for data
+                continue
+
+        if stop_generate_event.is_set():
+            logger.error("[Main Thread] generate thread has died, cancelling")
+            raise Exception("Generate thread has an exception and died, see error above")
```
finbarrtimbers: `stop_generate_event` is only set when we want to stop the generate thread normally, so I don't think we should raise an exception.

What do you think about wrapping the whole function in the while loop here? 

Something like:

```
with Timer("[Main Thread] 📦 Getting packed sequences from thread"):
    while not stop_generate_event.is_set():
        try:
            packed_data = packed_sequences_Q.get(timeout=2.0)
        except Empty:
            # Keep waiting for more data unless stop_event is set.
            continue
        data_thread_metrics = packed_data["metrics"]
        B = packed_data["B"]
        collated_data = packed_data["collated_data"]
        num_total_tokens += packed_data["num_new_tokens"]
        if B == 0:
            logger.warning("[Main Thread] 🤡 After packing, there is not enough data to train")
            return None, data_thread_metrics, num_total_tokens
        return collated_data, data_thread_metrics, num_total_tokens
```

open_instruct/grpo_fast.py
```diff
@@ -1873,15 +1889,24 @@ def generate_thread(
                 bar_format="{l_bar}{bar}{r_bar}\n",
             ) as pbar:
                 for future in futures.as_completed(engine_futures):
-                    processed_results.append(future.result())
-                    pbar.update(1)
+                    try:
+                        if future.exception() is not None:
```
finbarrtimbers: I don't think we need to check the exception. If `future.exception is not None` then calling `future.result()` [will raise the exception anyway](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.result). 

In that case I think we should call stop_event and then re-raise:

```
try:
    processed_results.append(future.result())
except Exception as e:
   logger.error(f"[Generate Thread] Error processing from queue: {e}")
   stop_event.set()
   raise e
```

No need for finally, if we're crashing we don't care about the progress bar.

open_instruct/grpo_fast.py
```diff
@@ -2350,9 +2375,19 @@ def main(args: Args, tc: TokenizerConfig, model_config: ModelConfig, num_eval_sa
...
 
-        collated_data, data_thread_metrics, num_total_tokens = load_data_from_packing_thread(
-            packed_sequences_Q, num_total_tokens
-        )
+        try:
+            collated_data, data_thread_metrics, num_total_tokens = load_data_from_packing_thread(
+                packed_sequences_Q, num_total_tokens, stop_generate_event
+            )
+        except Exception as e:
+            cleanup_training_resources(
```
finbarrtimbers, line 2377: Let's set stop_event here? I think that's more appropriate than in load_data_from_packing_thread.

Also you can just do `raise e`.

finbarrtimbers (changes requested):

open_instruct/grpo_fast.py
mnoukhov: done!

open_instruct/grpo_fast.py
mnoukhov: Implemented and added a return None. None None when stop_event.is_set breaks the while loop

open_instruct/grpo_fast.py
mnoukhov: Good point, done!

open_instruct/grpo_fast.py
mnoukhov: I don't think we're setting stop event either here or in load_data_from_packing thread. Generally setting it in the threads (generate_thread). Since load_data_from_packing_thread isn't throwing an error, there's no specific thing to raise so I made it a RuntimeError

mnoukhov:
Closing in favour of #856 which manages this using threadpoolexecutor

PR #841, opened 2025-07-29
Refactors the eval setup in `grpo_fast.py`

finbarrtimbers:
A few changes:

1. Instead of passing around three variables (`queries`, `ground_truths`, `datasets`) I made them one dataclass
2. Instead of having special cases for the eval prompts, I now insert them into the queue like the other prompts.
3. I wrapped all calls to ray.get with a new function, `ray_get_with_progress`, which reports the progress of the calls. This makes it much easier to identify deadlocks and timing issues in the logs. 

Experiments:

- single GPU run: [beaker](https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01K1EGYK8MN6X184MJVS94E53C?taskId=01K1EGYK8T6MQCX3WABGYE6R5D&jobId=01K1EGYKCMCHGSDEBRWCFH19P8).
- multi-node run: [beaker](https://beaker.allen.ai/orgs/ai2/workspaces/oe-adapt-code/work/01K1EGTSCHYCR3QY51YFXBTY7F?taskId=01K1EGTSCRBBWYD6A3KCHFP7WQ&jobId=01K1EGTSGGM43H61JZKA91C60E), [wandb](https://wandb.ai/ai2-llm/open_instruct_internal/runs/1vu06t0t).


open_instruct/grpo_fast.py
```diff
@@ -1988,19 +2000,28 @@ def maybe_evaluate(
...
 ):
     """Optionally evaluate the model."""
     try:
         # timeout 0.01 if this is the last training step or we're not evaluating
         # otherwise, wait to get the last evaluation generations (long timeout just in case)
         timeout = 0.01 if (training_step < args.num_training_steps or args.local_eval_freq < 0) else 100
-        eval_result = evaluation_inference_results_Q.get(timeout=timeout)
+
+        # Accumulate evaluation results from all vLLM engines
+        eval_result, eval_batch = accumulate_inference_batches(
```
hamishivi, line 2046: Does this mean we block waiting for eval results to come in now?

finbarrtimbers: No; we pass `timeout` through to `accumulate_inference_batches`, so that will raise.

open_instruct/grpo_fast.py
```diff
@@ -1294,20 +1330,26 @@ def data_preparation_thread(
...
         # Pack sequences
         if args.num_samples_per_prompt_rollout > 1:
-            queries = [item for item in queries for _ in range(args.num_samples_per_prompt_rollout)]
-            ground_truths = [item for item in ground_truths for _ in range(args.num_samples_per_prompt_rollout)]
-            datasets = [item for item in datasets for _ in range(args.num_samples_per_prompt_rollout)]
+            batch = Batch(
+                queries=flatten([[item] * args.num_samples_per_prompt_rollout for item in batch.queries]),
+                ground_truths=flatten([[item] * args.num_samples_per_prompt_rollout for item in batch.ground_truths]),
+                datasets=flatten([[item] * args.num_samples_per_prompt_rollout for item in batch.datasets]),
+                indices=flatten([[item] * args.num_samples_per_prompt_rollout for item in batch.indices])
```
hamishivi: nit: I sort of prefer the older double-loop list setup to flatten+list multiplication, since its clearer whats going on in the double-loop case.

open_instruct/grpo_fast.py
```diff
@@ -749,9 +771,9 @@ def broadcast_to_vllm(self):
                     if torch.distributed.get_rank() == 0:
                         torch.distributed.broadcast(param.data, 0, group=self.model_update_group)
         if torch.distributed.get_rank() == 0:
-            ray.get(refss)
+            ray_get_with_progress(refss, desc="Broadcasting weights to vLLM")
```
hamishivi: Looking at your logs, I saw that it seemed that generation and weight broadcasting was happening at the same time? Do you know what's going on there?
```bash
2025-07-30 21:35:21
[Generate Thread] Waiting for vLLM engines to process:  12%|█▎        | 2/16 [01:41<11:00, 47.18s/it]
2025-07-30 21:35:21
(LLMRayActor pid=3176, ip=10.95.0.99) INFO 07-30 21:35:21 [block_pool.py:316] Successfully reset prefix cache
2025-07-30 21:35:21

2025-07-30 21:35:21
Broadcasting weights to vLLM:   6%|▋         | 340/5424 [01:00<13:55,  6.08it/s]4<09:23, 40.27s/it]
2025-07-30 21:35:21
Broadcasting weights to vLLM:   8%|▊         | 451/5424 [01:00<08:19,  9.95it/s]
2025-07-30 21:35:21
Broadcasting weights to vLLM:  10%|█         | 564/5424 [01:00<05:12, 15.53it/s]
2025-07-30 21:35:21
Broadcasting weights to vLLM:  12%|█▏        | 677/5424 [01:00<03:22, 23.41it/s]
2025-07-30 21:35:33
Broadcasting weights to vLLM:  12%|█▏        | 677/5424 [01:12<03:22, 23.41it/s]
2025-07-30 21:35:39
[Generate Thread] Waiting for vLLM engines to process:  19%|█▉        | 3/16 [01:59<07:20, 33.88s/it]
```

open_instruct/grpo_fast.py
hamishivi: Does that mean if we have an eval on e.g. the last training step, the main training loop could finish and exit, missing waiting for the final eval to come in?

open_instruct/utils.py
```diff
@@ -67,6 +69,29 @@
 
 DataClassType = NewType("DataClassType", Any)
 
+
+def ray_get_with_progress(ray_refs: List[ray.ObjectRef], desc: str = "Processing") -> List[Any]:
```
hamishivi: nit: can we make the bar configurable a bit, e.g. only plays on a verbose mode? Looking at your wandb runs, it seems like then we log for e.g. every weight sync, which is useful for debugging but clutters up the log a bunch.

It also might be a bit weird at times, for example, the weight sync often looks longer than it actually is as it sits there and waits for the last generations to finish (although right now this doesnt quite seem to be the case, see previous review comment), which makes the loading bar kinda look funky (sitting around for a few minutes at 0%, then makes progress).

hamishivi (commented):
Some more comments! Also, for testing this, it would be good to run a larger script, but adjust num_episodes to some lower amount and maybe up the num_evals amount so the in-loop evals code runs a few times. I saw that it didn't run at all for the larger wandb run (as num episodes was really high and num_evals was 5).

open_instruct/grpo_fast.py
finbarrtimbers: So, the generate thread never stops. It continually loops over the engines to see if they want to process anything. The behaviour currently is that when `process_from_queue` is called, they will pull a batch from the queue and process it. This times out, so if there's nothing in the queue, they just return. 

Previously ([code before any of my changes](https://github.com/allenai/open-instruct/blob/d18b7e943118260f274fd1b3b28a7e2c11ba38ef/open_instruct/grpo_fast.py#L1103)), the behaviour was similar, in that the generate thread would hang until something new was pushed to `param_prompt_Q`. The difference is that we give control of the queue to the actors now. 

This is kind of a dumb way to do things! In subsequent PRs, I'll be changing this loop so that [1) the actor will keep pulling batches until a flag is set on the learner ](https://github.com/allenai/open-instruct/pull/837) and [2) we'll insert individual prompts in the queue and have the actor stream prompts in/out](https://github.com/allenai/open-instruct/compare/async-updates...continual-processing). When PR 2 is in, that will allow for inflight weight updates (if we want) and for continual processing, which should lead to up to 50% higher throughput. 

Anyway, all this to say, the message in generate thread should perhaps be "waiting for vLLM engines to return" or something like that; we're waiting for `process_from_queue` to return, which will happen when either it is done processing the batch, or times out waiting for the queue. The weight update happens asynchronously to the engines when each engine returns control (as due to the GIL ray doesn't let multiple calls happen to an actor simultaneously). I believe this is the same behaviour as happened before any of my changes; there, we had `vllm_generate_thread` running until the queue was empty, and weight updates would also happen asynchronously once `generate_with_engines` returned and the lock on the actors was released. 

happy to go into more detail if any of this is unclear.

open_instruct/grpo_fast.py
finbarrtimbers: What about if I gave it a name, like this? 

```
def repeat_each(seq, k):
    return [x for item in seq for _ in range(k) for x in (item,)]
...
queries = repeat_each(batch.queries, args.num_samples_per_propmt_rollout)
ground_truths = repeat_each(batch.ground_truths, args.num_samples_per_propmt_rollout)
datasets = repeat_each(batch.datasets, args.num_samples_per_propmt_rollout)
indices = repeat_each(batch.indices, args.num_samples_per_propmt_rollout)
```

I just want to make clear what's going on to the casual observer.

finbarrtimbers:
> Some more comments! Also, for testing this, it would be good to run a larger script, but adjust num_episodes to some lower amount and maybe up the num_evals amount so the in-loop evals code runs a few times. I saw that it didn't run at all for the larger wandb run (as num episodes was really high and num_evals was 5).

Ugh, nice catch. Will do.

finbarrtimbers:
> Some more comments! Also, for testing this, it would be good to run a larger script, but adjust num_episodes to some lower amount and maybe up the num_evals amount so the in-loop evals code runs a few times. I saw that it didn't run at all for the larger wandb run (as num episodes was really high and num_evals was 5).

@hamishivi do you have a specific script in mind?

open_instruct/grpo_fast.py
finbarrtimbers: Somewhat sidestepping your question, I think it's the same exact behaviour as before, right? 

Directly answering your question: 

- `maybe_evaluate` is part of the core training loop, so the call to `accumulate_inference_batches` will block the main training loop for `timeout` seconds. 
- `timeout` is 0.01 during training, and 100 otherwise, so it shouldn't timeout at the end unless 100s pass without results

I believe the comment is wrong and has been wrong for a while. Maybe I'm misunderstanding! The behaviour here looks the same as [Costa's original PR](https://github.com/allenai/open-instruct/commit/b5f60321d8186500d10660e3d7b575ae619a678b#diff-ae23a957db4e48f8e20bc1dd47dfef1414fc0b800674f67bc21d3ed94788b154R1712).

finbarrtimbers: Image

open_instruct/grpo_fast.py
hamishivi: Thanks for the explanation! I guess its weird because the logs suggest inflight weight updates (although I assume its just weird flushing behaviour....?). Or is it that these are interleaved logs from separate vllm engines (one is done and starts the broadcast, while another is still generating?)?

open_instruct/utils.py
hamishivi: discuss offline. 
I think making "Resetting vLLM prefix cache" and "Broadcasting weights to vLLM" only viewable in debug mode is best. Mixed feelings about the generation bar, probably its just useful to keep?

open_instruct/grpo_fast.py
hamishivi: Sure, thats good too!

hamishivi:
@finbarrtimbers maybe reduce the larger script to ~10k episodes and set 100 evals? Doesn't have to run through, but then at least an eval should fire after a few steps. (aside: I feel we should change num_evals so the user sets after how many steps it runs, rather than setting the total num evals over the run).

open_instruct/grpo_fast.py
finbarrtimbers: Interleaved logs is my understanding. When `generate_thread` is running, it's constantly acquiring and releasing the locks on each of the engines as it loops through `vllm_refs`, which is the exact same thing that the weight updates are doing. This has been the source of some deadlocks in my development and is nasty to debug (hence the progress bars).

finbarrtimbers: I will update the log message.

open_instruct/test_utils.py
```diff
@@ -75,6 +76,26 @@ def test_time_parser_used_in_get_beaker_dataset_ids(self):
...
+            ("empty_sequence", [], 3, []),
+            ("integers", [1, 2, 3], 2, [1, 1, 2, 2, 3, 3]),
+            ("mixed_types", ["a", 1, None, True], 2, ["a", "a", 1, 1, None, None, True, True]),
+            ("single_element", ["x"], 5, ["x", "x", "x", "x", "x"]),
+        ]
+    )
+    def test_repeat_each(self, name, sequence, k, expected):
+        """Test the repeat_each function with various inputs."""
+        result = repeat_each(sequence, k)
+        self.assertEqual(result, expected)
```
hamishivi, line 96: really minor nit would be that I'd wanna test if mutating a sequence item after repeat each doesn't change the repeated versions, but this is really minor.

open_instruct/grpo_fast.py
```diff
@@ -1842,45 +1898,27 @@ def load_data_from_packing_thread(packed_sequences_Q: Queue, num_total_tokens: i
...
-                )
+                engine.process_from_queue.remote(num_training_steps, resume_training_step, timeout=0.1)
                 for engine in vllm_engines
             ]
             engine_futures = [ref.future() for ref in engine_refs]
             processed_results = []
             with tqdm(
                 total=len(vllm_engines),
-                desc="[Generate Thread] Waiting for vLLM engines to process",
+                desc="[Generate Thread] Waiting for vLLM engines to return",
```
hamishivi, line 1915: oh, nice change!

hamishivi (approved):
I think this is good to go assuming the test run works!

open_instruct/utils.py
finbarrtimbers: Done.

open_instruct/test_utils.py
finbarrtimbers: Done.

open_instruct/grpo_fast.py
finbarrtimbers: Thank you!

finbarrtimbers:
Experiments are now running:

- ✅ [Big boy](https://beaker.allen.ai/orgs/ai2/workspaces/oe-adapt-code/work/01K20A5R6RFE8YVHSJ9V2D6JXC?) (with 10k episodes and 100 evals), [Wandb](https://wandb.ai/ai2-llm/open_instruct_internal/runs/55g5rw98)
- ✅ [Single GPU](https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01K20AA9SDMWDX58N51TJ2PC7G?taskId=01K20AA9SMNPR85HAX7TTE441T&jobId=01K20AA9X5ZJMA9RXN7GGYPMJC) (with tracking enabled)
- ✅ [Single GPU, but merged to HEAD](https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01K20AE155CJ8H59P7H6V4EED2?taskId=01K20AE15AA9EAQS4VCKV6ZWCX&jobId=01K20AE191VSS11900NTN56C70)

finbarrtimbers:
Still waiting on the big script to run, but we are getting eval responses. See screenshot (from the single GPU job).
Screenshot 2025-08-06 at 1 05 40 PM

PR #849, opened 2025-08-04
force manual set of local model name for upload to GS

mnoukhov:
currently we're using the basename of a local model but that frequently is `step_23943` 

instead force the user to set `--gs_model_name` and set the `commit_hash` to be an md5 hash of the model's local path


saurabh111233212 (approved):
nice, lgtm

PR #851, opened 2025-08-04
use a proper async client for code verification

saurabh111233212:
need these changes to use AWS lambda-based code verifiers, which are needed because larger test cases eat up all the CPU memoery on the local servers when we use that apporach

these changes don't break the local server thing, tested [here](https://wandb.ai/ai2-llm/open_instruct_internal/runs/46f2wyit/). 
^Compare this with reference run from the old impl (what's in main rn) [here](https://wandb.ai/ai2-llm/open_instruct_internal/runs/dnvtuklj)

and this enables AWS lambda solution, tested [here](https://wandb.ai/ai2-llm/open_instruct_internal/runs/9nqd72ov)

### DIFF
--- a/open_instruct/ground_truth_utils.py
+++ b/open_instruct/ground_truth_utils.py
@@ -845,7 +845,6 @@ class CodeVerifier(VerifierFunction):
                 return response.json()
 
             result = await asyncio.to_thread(make_request)
-
             passes = result["results"]
             pass_rate = sum(passes) / len(passes) if passes else 0.0
             score = 0.0 if pass_rate < self.pass_rate_reward_threshold else pass_rate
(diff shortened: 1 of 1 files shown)


open_instruct/ground_truth_utils.py
```diff
@@ -805,15 +827,25 @@ async def async_call(
...
+                    self.verifier_config.code_api_url,
+                    json=payload,
+                    headers={"Content-Type": "application/json"},
+                    timeout=http_timeout,
                 )
                 response.raise_for_status()
                 return response.json()
 
             result = await asyncio.to_thread(make_request)
+
```
finbarrtimbers: Accidental whitespace?

finbarrtimbers (approved):

open_instruct/ground_truth_utils.py
saurabh111233212: oops ! fixed

PR #853, opened 2025-08-06
Log Train vs Generation time to see GPU idling

mnoukhov:
Log Training time, Generation time, to figure out how much we're idling GPUs

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -49,6 +49,7 @@ import json
 import logging
 import math
 import os
+import queue
 import shutil
 import socket
 import threading
@@ -1850,12 +1851,14 @@ def generate_thread(
     num_training_steps,
     resume_training_step,
     stop_event,
+    generate_metrics_Q,
 ):
     """Thread function that repeatedly calls process_from_queue on vllm engines."""
     logger.info("[Generate Thread] 🚀 Starting generation thread")
 
     while not stop_event.is_set():
-        with Timer("🔥 Generation time"):
+        timer = Timer("🔥 Generation time")
+        with timer:
             engine_refs = [
                 engine.process_from_queue.remote(
                     generation_config,
@@ -1884,6 +1887,8 @@ def generate_thread(
         if num_processed == 0:
             # If no batches were processed, sleep for a short time to avoid busy waiting
             time.sleep(1)
+        else:
+            generate_metrics_Q.put({"generation_time": timer.duration})
 
     logger.info("[Generate Thread] 🛑 Stopping generation thread")
 
@@ -2300,6 +2305,7 @@ def main(args: Args, tc: TokenizerConfig, model_config: ModelConfig, num_eval_sa
     packing_thread.start()
 
     # Create and start the generate thread
+    generate_metrics_Q = Queue(maxsize=args.async_steps)
     stop_generate_event = threading.Event()
     generation_thread = threading.Thread(
         target=generate_thread,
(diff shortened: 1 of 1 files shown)


hamishivi (approved):
Fixed up one or two minor things and now seems to run fine with the debug script. The times in the metrics dict seem to match the times that get logged to stdout when running the debug script.

open_instruct/grpo_fast.py
```diff
@@ -1896,12 +1899,14 @@ def load_data_from_packing_thread(packed_sequences_Q: Queue, num_total_tokens: i
...
-def generate_thread(vllm_engines, local_eval_every, num_training_steps, resume_training_step, stop_event):
+def generate_thread(
+    vllm_engines, local_eval_every, num_training_steps, resume_training_step, stop_event, generate_metrics_Q
+):
     """Thread function that repeatedly calls process_from_queue on vllm engines."""
     logger.info("[Generate Thread] 🚀 Starting generation thread")
 
     while not stop_event.is_set():
-        with Timer("🔥 Generation time") as _gen_timer:
+        with Timer("🔥 Generation time") as timer:
```
finbarrtimbers, line 1913: nice!!

mnoukhov:
fixed lots of stuff, currently running and finishing fine!
only issue is training time seems to be very fast but `trainer_idling` is always 0

looking at nvitop, gpu utilization is full for the trainer so maybe this is correct, just need to figure out what's happening to use the GPU if not training

finbarrtimbers:
Nice!!

Can you calculate tokens per second? I can then give you a snippet of code
to turn that into MFU.

open_instruct/grpo_fast.py
```diff
@@ -1963,67 +1975,73 @@ def one_training_step(
...
 
-    return average_metrics
+    average_metrics = {k: sum(m[k] for m in metrics_list) / len(metrics_list) for k in metrics_list[0]}
+    total_time = time.time() - start_time
+    metrics = {
+        "episode": episode,
+        "training_step": training_step,
+        "val/num_total_tokens": num_total_tokens,
+        "epoch": episode / args.num_samples_per_prompt_rollout / len(train_dataset),
+        "tokens_per_second": num_total_tokens / total_time,
```
mnoukhov, line 2029: @finbarrtimbers we already have tokens_per_second, but I can put it under `time/` 
how can I add MFU?

finbarrtimbers: Two options: 

1. Approximate the flops with 3 * num_params * bytes_per_param.
2. Can calculate them using [this function](https://github.com/allenai/open-instruct/blob/1275d3682cccbf9f457cd094291d4bddc085d4aa/open_instruct/benchmark_generators.py#L218) and then multiply by 3 (backwards pass requires 3x as many flops as forwards pass). Also an approximation but should be accurate. You *could* directly calculate the flops for the training step, but the flops counter is inaccurate with torch.compile so I wouldn't bother.

finbarrtimbers: Oh, and then divide by GPU flops: https://github.com/allenai/open-instruct/blob/main/open_instruct/benchmark_generators.py#L33

mnoukhov: does this account for the generation time as well as training time?

finbarrtimbers: This is the time for a single training step. 

For inference, you can use num_params * bytes_per_param ([reference](https://kipp.ly/transformer-inference-arithmetic/)).

mnoukhov: prime-rl does  `flop_per_token = 6 * num_params + 12 * l * h * q * t` instead of `3*num_params * bytes per param` which I assume is 6 because 2 bytes per bf16 and then 12* is the generation. Wyt?

https://github.com/PrimeIntellect-ai/prime-rl/blob/9657a591added96d2671f32f9307bd4d5d92939c/src/prime_rl/trainer/perf.py#L75

PR #854, opened 2025-08-06
Fix to beaker integration test.

finbarrtimbers:
Currently broken. 

Bisect results:


requirements.txt
```diff
@@ -1,891 +1,253 @@
 # This file was autogenerated by uv via the following command:
 #    uv export --format requirements-txt --no-hashes --all-extras --no-emit-project
 absl-py==2.3.0
-    # via tensorboard
```
hamishivi, line 4: why did these comments get removed?

finbarrtimbers: Discussed via Slack (I have no idea why).

PR #860, opened 2025-08-07
Added test script.

finbarrtimbers:
I've been using this to test with, so thought I'd check it in.


finbarrtimbers:
> Seems good to me. I guess if you wanted to use less resources, could use 2 nodes (1 train / 1 vllm) and adjust some of the batch size stuff. Similarly with output length (could reduce to 4k/8k) -- but if this doesn't stick around waiting for long I think its fine.

Great suggestion! I just switched it to use 2 nodes and halved the number of vllm engines and the number of learners. 

Switched output length to 4096 and `num_unique_prompts_rollout` to 128.

PR #859, opened 2025-08-07
Change `LLMRayActor` to continually process individual prompts

finbarrtimbers:

PR #861, opened 2025-08-07
Change num_evals so the user sets after how many steps it runs, rather than setting the total num evals over the run.

finbarrtimbers:
Removes the `num_evals` arg and instead makes users use the `local_eval_every` flag which is the number of steps after which we should run an eval.

Currently, we set `num_training_steps = total_episodes / (num_unique_prompts_rollout * num_samples_per_prompt_rollout)`. I have used this to update `local_eval_every` based on values for `num_evals` in the checked in scripts.

Successful single GPU experiment: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01K255N2YB122MZKQR6QZ2ZN39?taskId=01K255N2YHNFR99TGQS5FP48WF&jobId=01K255N33026EYZCW996ATSHNM).

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -214,10 +214,8 @@ class Args:
     """RUNTIME VALUE: The number of processes (GPUs) to use"""
     num_training_steps: Optional[int] = None
     """RUNTIME VALUE: The number of training_steps to train"""
-    num_evals: int = 10
-    """this sets how many in-loop evals we do during training. in-loop evals reuse the generation/reward verifier setup."""
-    local_eval_freq: Optional[int] = None
-    """this controls the number of in-loop evals, which reuses the generation/reward verifier setup. don't set this directly, but set via num_evals."""
+    local_eval_every: int = 100
+    """Run evaluation after this many training steps. This controls in-loop evals, which reuse the generation/reward verifier setup."""
     save_freq: int = 200
     """How many train steps to save the model"""
     allow_world_padding: bool = False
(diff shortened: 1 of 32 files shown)


open_instruct/grpo_fast.py
```diff
@@ -214,10 +214,8 @@ class Args:
     """RUNTIME VALUE: The number of processes (GPUs) to use"""
     num_training_steps: Optional[int] = None
     """RUNTIME VALUE: The number of training_steps to train"""
-    num_evals: int = 10
-    """this sets how many in-loop evals we do during training. in-loop evals reuse the generation/reward verifier setup."""
-    local_eval_freq: Optional[int] = None
-    """this controls the number of in-loop evals, which reuses the generation/reward verifier setup. don't set this directly, but set via num_evals."""
+    local_eval_every: int = 100
+    """Run evaluation after this many training steps. This controls in-loop evals, which reuse the generation/reward verifier setup."""
```
hamishivi:
```suggestion
    """Run evaluation after this many training steps. This controls in-loop evals, which reuse the generation/reward verifier setup. Set to -1 to disable."""
```

hamishivi (approved):
I wonder if we should make this change in the other scripts too? But not big deal either way.

finbarrtimbers:
> I wonder if we should make this change in the other scripts too? But not big deal either way.

My current plan is to start merging the other scripts into grpo_fast.py (or at least start sharing more code between them, including the `Args` class) once I get https://github.com/allenai/open-instruct/pull/859 merged.

So I agree that we should do that.

hamishivi:
> So I agree that we should do that.

makes sense!

PR #864, opened 2025-08-07
adding wandb to all distributed jobs

mnoukhov:
currently doesn't work, we get a timeout on the second wandb.init


mnoukhov:
closing as either impossible or very annoying due to Ray

PR #866, opened 2025-08-08
Disables wandb in the Beaker experiment integration test.

finbarrtimbers:
Previously, [jobs were failing](https://github.com/allenai/open-instruct/actions/runs/16835010576/job/47692192758) because I [changed the underlying script to enable Wandb by default](https://github.com/allenai/open-instruct/commit/ff0049edd1bdc7e4645a80e8a5ce393663e82526).

Now, the [integration test succeeds](https://github.com/allenai/open-instruct/actions/runs/16836413976/job/47696880915).


mnoukhov:
I think this is supposed to `WANDB_MODE=disabled` at least according to the website https://docs.wandb.ai/guides/track/environment-variables/, does WANDB_DISABLED=true work?

finbarrtimbers:
> I think this is supposed to `WANDB_MODE=disabled` at least according to the website https://docs.wandb.ai/guides/track/environment-variables/, does WANDB_DISABLED=true work?

Thank you! Updated.

finbarrtimbers:
Ahhh it doesn't matter at all because this env var isn't passed to the
beaker script!

PR #868, opened 2025-08-08
Update logging in accumulate_inference_batches

finbarrtimbers:
Only lets accumulate_inference_batches update when there's at least one iteration done.


hamishivi (commented):
From a quick test this still seemed to log the progress bars, but I made a change in #871 to avoid this logging entirely unless verbose mode is set, so maybe lets close this PR in favour of that?

PR #871, opened 2025-08-08
Avoid pickle-imports slowing things down

hamishivi:
I noticed that sometimes we end up re-importing deepspeed due to pickling/unpickling of the PromptRequest/etc classes. This moves those classes into their own file with minimal imports so we avoid this. From local testing, this saves > 10 seconds every some number of generations!

I also moved some of the progress bars into verbose mode, since I feel like they were cluttering things a bit? but open to thoughts here.

### DIFF
--- a/.github/workflows/beaker-experiment.yml
+++ b/.github/workflows/beaker-experiment.yml
@@ -87,7 +87,7 @@ jobs:
       - name: Install dependencies
         run: |
           # Install development dependencies needed for mason.py
-          uv sync --only-group dev
+          uv sync
 
       - name: Delete huge unnecessary tools folder
         run: |
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+[![Beaker Experiment Launch](https://github.com/allenai/open-instruct/actions/workflows/beaker-experiment.yml/badge.svg)](https://github.com/allenai/open-instruct/actions/workflows/beaker-experiment.yml) [![build_open_instruct](https://github.com/allenai/open-instruct/actions/workflows/push-image.yml/badge.svg)](https://github.com/allenai/open-instruct/actions/workflows/push-image.yml)
+
 # Training Open Instruction-Following Language Models
 
 This repo serves as an open effort on instruction-tuning and post-training popular pretrained language models on publicly available datasets. We release this repo and will keep updating it with:
--- a/scripts/train/build_image_and_launch.sh
+++ b/scripts/train/build_image_and_launch.sh
@@ -30,7 +30,7 @@ fi
 
 # Install Python dependencies
 echo "Installing dependencies with uv..."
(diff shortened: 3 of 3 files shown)


open_instruct/grpo_fast.py
```diff
@@ -1925,22 +1924,29 @@ def generate_thread(vllm_engines, local_eval_every, num_training_steps, resume_t
...
+            if args.verbose:
+                with tqdm(
+                    total=len(vllm_engines),
+                    desc="[Generate Thread] Waiting for vLLM engines to return",
+                    bar_format="{l_bar}{bar}{r_bar}\n",
+                ) as pbar:
+                    for future in futures.as_completed(engine_futures):
+                        processed_results.append(future.result())
+                        pbar.update(1)
+            else:
```
mnoukhov: you can just do `tqdm(...disable=not args.verbose)`

open_instruct/grpo_fast.py
```diff
@@ -1236,12 +1230,17 @@ def accumulate_inference_batches(
...
-        bar_format="{l_bar}{bar}{r_bar}\n",
-    ):
+    if args.verbose:
+        engine_iter = tqdm(
+            range(args.vllm_num_engines),
+            total=args.vllm_num_engines,
+            desc=f"Accumulating results from {args.vllm_num_engines} engines ({timeout=})",
+            bar_format="{l_bar}{bar}{r_bar}\n",
+        )
+    else:
```
mnoukhov: same here

mnoukhov (changes requested):
looks good except these two

open_instruct/grpo_fast.py
```diff
@@ -1936,11 +1930,15 @@ def generate_thread(vllm_engines, local_eval_every, num_training_steps, resume_t
...
                 desc="[Generate Thread] Waiting for vLLM engines to return",
                 bar_format="{l_bar}{bar}{r_bar}\n",
+                disable=not args.verbose,
             ) as pbar:
                 for future in futures.as_completed(engine_futures):
                     processed_results.append(future.result())
                     pbar.update(1)
             num_processed = sum(int(result) for result in processed_results)
+            # Suppress timing output if nothing was processed
+            if num_processed == 0:
```
finbarrtimbers, line 1940: What if we instead have _gen_timer check if the timing output is <1s and skips outputting then? That's almost every one of the issues where log spamming is a problem. 

Anyway, I leave it to your judgement. Whatever you prefer.

hamishivi: I sort of prefer checking for 0 processed, since there's some imaginary case where the generation just runs really fast?

finbarrtimbers: Very reasonable!

PR #884, opened 2025-08-12
Fix up finetune

hamishivi:
- Fix the attention implementation call in finetune.py so it actually works.
- Update datasets due to an issue with datasets from pre/after 4.0.0 not being compatible (https://github.com/huggingface/datasets/issues/7676). Open to instead downgrading the version instead...?

other changes just downstream of me updating the requirements for the new datasets version.


hamishivi:
@jacob-morrison I tested with tulu 3 (combined) + a subset of the tulu subsets in the dataset mixer and it seemed fine. I also changed up the stats computation code a little to speed it up, as it was hanging for > 10 min when trying to process large datasets (e.g. openthoughts2 1M).

I also tested the GRPO fast debug script and it seems to work fine there!

PR #885, opened 2025-08-13
Removes git from the Docker image.

finbarrtimbers:
Instead, pass the current git commit (and git branch!) as environment variables in `mason.py`. Has the nice side effect of making it easier to debug experiments from the Beaker page.

This leads to a minor decrease in image size.

Successful single GPU run: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01K2JAH7STMXGYWN20AR71DN03?taskId=01K2JAH7T2WA8QGREJGJT6XGWD&jobId=01K2JAH7XRWX5AXHEAB5JMW3GR), [Wandb](https://wandb.ai/ai2-llm/open_instruct_internal/runs/s6kftvuo/overview).


mason.py
```diff
@@ -266,6 +267,13 @@ def get_env_vars(pure_docker_mode: bool, cluster: List[str], beaker_secrets: Lis
                 secret=secret["value"],
             )
         )
+    
+    # Add Git commit and branch information
+    git_commit = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"], text=True).strip()
+    env_vars.append(beaker.EnvVar(name="GIT_COMMIT", value=git_commit))
+    
+    git_branch = subprocess.check_output(["git", "rev-parse", "--abbrev-ref", "HEAD"], text=True).strip()
+    env_vars.append(beaker.EnvVar(name="GIT_BRANCH", value=git_branch))
```
hamishivi: does this mean now the tag will be associated with the branch the mason job is run from? I guess usually this is fine, although sometimes im launching stuff from mason to an image from another branch. 

Maybe we could compute the git hash in the dockerfile creation...?

finbarrtimbers: Done!

Dockerfile
```diff
@@ -1,5 +1,13 @@
 FROM ghcr.io/allenai/cuda:12.8-dev-ubuntu22.04-torch2.7.0-v1.2.170
 
+# Add build arguments for git information
+ARG GIT_COMMIT=""
+ARG GIT_BRANCH=""
```
hamishivi, line 5: can we add to the docs to use this arg?

finbarrtimbers:
> Seems good, we should add the new build args to the main readme too!

Done!

Dockerfile
finbarrtimbers: What do you mean?

hamishivi: ah i meant add to the readme, which you have done!

hamishivi (approved):

PR #889, opened 2025-08-13
Adds the wandb url to the experiment description.

finbarrtimbers:
Successful experiment run: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01K2JG8X2CTKWPGRVW1GNSEQ5M?taskId=01K2JG8X2JQXNJM410A67R4PJA&jobId=01K2JG8X6EEGA3W3WFAREAKHF9).

And in the future, we can make it clickable:

https://github.com/allenai/beaker/issues/8835

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -122,12 +122,12 @@ from open_instruct.utils import (
     is_beaker_job,
     launch_ai2_evals_on_weka,
     maybe_get_beaker_config,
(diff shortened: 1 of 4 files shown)


mnoukhov (changes requested):
Looks good but can we add an `is_beaker_job` check to maintain compatibility with non-ai2 runs?

Could also switch to `wandb_url is not None` but that's minor

PR #892, opened 2025-08-13
Fix crash for debug script

hamishivi:
In an interactive session, we can't find the experiment (I think because it requires a diff search) and the run crashes. This prevents that. Broke in #889 since there the beaker client searches for the experiment without a try-except guard.

### DIFF
--- a/open_instruct/utils.py
+++ b/open_instruct/utils.py
@@ -977,7 +977,7 @@ def maybe_update_beaker_description_with_wandb_url(wandb_url: str) -> None:
         spec = client.experiment.get(os.environ["BEAKER_WORKLOAD_ID"])
     except beaker.exceptions.ExperimentNotFound:
         print(f"Failed to update Beaker experiment description with wandb URL: {wandb_url}")
-        print("This might be fine if you are e.g. running in an interactive job.")
+        logger.warning("This might be fine if you are e.g. running in an interactive job.")
         return
     current_description = spec.description or ""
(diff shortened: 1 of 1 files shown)


finbarrtimbers:
Can you use logging instead of print?

open_instruct/utils.py
```diff
@@ -973,7 +973,12 @@ def maybe_update_beaker_description_with_wandb_url(wandb_url: str) -> None:
         return
 
     client = beaker.Beaker.from_env()
-    spec = client.experiment.get(os.environ["BEAKER_WORKLOAD_ID"])
+    try:
+        spec = client.experiment.get(os.environ["BEAKER_WORKLOAD_ID"])
+    except beaker.exceptions.ExperimentNotFound:
+        print(f"Failed to update Beaker experiment description with wandb URL: {wandb_url}")
```
finbarrtimbers:
```suggestion
        logger.warning(f"Failed to update Beaker experiment description with wandb URL: {wandb_url}")
```

open_instruct/utils.py
```diff
@@ -973,7 +973,12 @@ def maybe_update_beaker_description_with_wandb_url(wandb_url: str) -> None:
...
+        print("This might be fine if you are e.g. running in an interactive job.")
```
finbarrtimbers:
```suggestion
        logger.warning("This might be fine if you are e.g. running in an interactive job.")
```

finbarrtimbers (changes requested):
Use logger.warning instead?

finbarrtimbers (approved):

PR #903, opened 2025-08-18
Fix old ifeval verifier.

### LINKED ISSUE #902: Suspicious code in open-instruct/open_instruct /ground_truth_utils.py
Please help to check whether I am wrong or not.

```
class IFEvalVerifierOld(VerifierFunction):
    """
    Verifier for ifeval tasks that delegates evaluation to a function
    specified in the constraint.

    The constraint may be a JSON string or a dictionary containing a key
    'func_name' used to lookup the evaluation function.
    """

    def __init__(self, verifier_config: Optional[VerifierConfig] = None) -> None:
        super().__init__("ifeval_old", verifier_config=verifier_config, weight=1.0)

    def __call__(
        self, tokenized_prediction: List[int], prediction: str, label: Union[str, Dict], query: Optional[str] = None
    ) -> VerificationResult:
        constraint = label
        answer = remove_thinking_section(prediction)
        if isinstance(constraint, str):
            constraint = json.loads(constraint)
        if "func_name" not in constraint:
            logger.warning("Constraint missing 'func_name': %s", constraint)
        func_name = constraint.pop("func_name")
        func = IF_FUNCTIONS_MAP[func_name]
        non_none_args = {k: v for k, v in constraint.items() if v is not None}
        if not constraint:
            return VerificationResult(score=float(func(prediction)))
        return VerificationResult(score=float(func(answer, **non_none_args)))
```
```
return VerificationResult(score=float(func(prediction)))
```
 should be 
```
return VerificationResult(score=float(func(answer)))
```

hamishivi:
Fixes #902

### DIFF
--- a/open_instruct/ground_truth_utils.py
+++ b/open_instruct/ground_truth_utils.py
@@ -328,7 +328,7 @@ class IFEvalVerifierOld(VerifierFunction):
         func = IF_FUNCTIONS_MAP[func_name]
         non_none_args = {k: v for k, v in constraint.items() if v is not None}
         if not constraint:
-            return VerificationResult(score=float(func(prediction)))
+            return VerificationResult(score=float(func(answer)))
         return VerificationResult(score=float(func(answer, **non_none_args)))
 
 


hamishivi:
tbh I want to remove this verifier but you're right that we should have a judge test suite...

PR #905, opened 2025-08-18
Adds unit tests which run on GPU for open-instruct.

finbarrtimbers:
Now, GPU tests run on Beaker. Per PR, we run a single concurrent instance at normal priority. 

Also makes it so that we caches the intermediate Docker layers to the Github container registry, both for local and CI builds.

### DIFF
--- a/scripts/train/debug/run_gpu_pytest.sh
+++ b/scripts/train/debug/run_gpu_pytest.sh
@@ -7,6 +7,8 @@ BEAKER_IMAGE="${1:-${BEAKER_USER}/open-instruct-integration-test}"
 echo "Using Beaker image: $BEAKER_IMAGE"
 
 uv run python mason.py \
+       --cluster ai2/jupiter \
+       --cluster ai2/ceres \
        --cluster ai2/saturn \
        --image "$BEAKER_IMAGE" \
        --description "GPU tests for test_*_gpu.py" \


open_instruct/test_grpo_fast_gpu.py
```diff
@@ -0,0 +1,83 @@
...
+            pretrain=tokenizer_name,
+            revision="main",
+            seed=42,
+            enable_prefix_caching=False,
+            max_model_len=512,
+            vllm_gpu_memory_utilization=0.5,
+            prompt_queue=param_prompt_Q,
+            results_queue=inference_results_Q,
+            actor_manager=actor_manager,
+        )
```
mnoukhov, line 211: not sure but feels like the tokenizer and vllm engine could be in the setup of the test

open_instruct/test_grpo_fast_gpu.py
```diff
@@ -0,0 +1,83 @@
...
+
+        generation_config = SamplingParams(temperature=0.0, top_p=1.0, max_tokens=5, n=1)
```
mnoukhov: do you think its possible to get a deterministic output for Pythia 14m and specifically test for the expected completion for this prompt? Not sure how important that is

mnoukhov (commented):
Is this going to be a GPU job for every single updated main? I can't tell but that feels like it would be a substantial number of jobs on the cluster :/

finbarrtimbers:
> Is this going to be a GPU job for every single updated main? I can't tell but that feels like it would be a substantial number of jobs on the cluster :/ 

My attitude is to enable it and see if we get complaints. 

I could also make the trigger a lot more restrictive and have it only trigger on changes to grpo_fast or vllm_utils3.py.

mnoukhov (approved):
triggering on changes to `grpo_fast` or `vllm3_utils` makes sense to me
still feels like it could be overkill but I'm inclined to say let's try and then rollback if necessary

mnoukhov:
I accidentally clicked `approve` instead of `request changes` but I'll leave it as is. Swap to changes on those files only and merge!

PR #915, opened 2025-08-19
`actually_fix_build_v2_final.txt`

finbarrtimbers:
So, unfortunately, the previous PR didn't actually fix the build? I'm not sure what happened. This one does. 

Commit at HEAD is 0d38c8b8. When run with that commit, [manually triggered build succeeds](https://github.com/allenai/open-instruct/actions/runs/17076967500).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,8 +69,6 @@ ENV UV_CACHE_DIR=/root/.cache/uv
 ENV HF_HUB_ENABLE_HF_TRANSFER=1
 ENV UV_COMPILE_BYTECODE=0
 
-WORKDIR /stage/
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -69,16 +69,14 @@ ENV UV_CACHE_DIR=/root/.cache/uv
 ENV HF_HUB_ENABLE_HF_TRANSFER=1
 ENV UV_COMPILE_BYTECODE=0
 
-WORKDIR /stage/
```
finbarrtimbers, line 72: Note: There are multiple of these in the file, so it's fine to delete it.

PR #916, opened 2025-08-19
Prevent duplicate runs in Github Actions.

finbarrtimbers:
Previously, we would see duplicate Build and Beaker Experiment runs because both the `merge_group` and `push` triggers were triggering. E.g. on any merged PR ([example](https://github.com/allenai/open-instruct/commit/a916940b87dd83a5ef526e774e9e548b51a12f23)).


mnoukhov (approved):
I don't fully understand this, but I trust you on the details of the merge queue vs push actions

PR #918, opened 2025-08-19
Fixes batching bug. 

finbarrtimbers:
Previously, if the number of prompts was less than the number of engines, we'd insert empty prompts, which would cause the actor to [fail](https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01K2ZKBDSHZRHC6K3N6AH0YWKN?taskId=01K2ZKBDSQNWWYEN0XZF6AE51Q&jobId=01K2ZKBDXVZ0EKZH4VJPZ6NGMB). 

Now, in that case, we log a warning and skip inserting the empty prompts. 

I have also added tests to cover that edge case to make sure we don't insert empty prompts. 

Runs:

1. Single GPU run, standard config: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01K31X5WN30WZ7D5JEEH8GTBRC?taskId=01K31X5WN7AAK0XRN90157B30D&jobId=01K31X5WZJ39P2B973JVAFM76B)
2. Multi-GPU run, using less eval queries than engines: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01K31X7DMKMG4FP0M13561ECW4?taskId=01K31X7DMV3M8EDTWT136SM19N&jobId=01K31X7DRQ7RY0WKQMD7600HEB)

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -1897,30 +1897,15 @@ def split_and_insert_batch(
     is_eval: bool = False,
 ) -> None:
     """Split a batch into multiple inference batches and insert individual prompts into queues and mapping."""
-    # Handle case where we have fewer queries than engines
-    if len(batch.queries) < vllm_num_engines:
-        logging.warning(
-            f"Number of queries ({len(batch.queries)}) is less than number of engines ({vllm_num_engines}). "
-            f"Will distribute single prompts to first {len(batch.queries)} engines."
-        )
-        inference_batch_size = 1
-    else:
-        # Normal case: distribute evenly
-        inference_batch_size = len(batch.queries) // vllm_num_engines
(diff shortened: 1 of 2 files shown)


open_instruct/grpo_fast.py
```diff
@@ -1891,11 +1896,17 @@ def split_and_insert_batch(
     is_eval: bool = False,
 ) -> None:
     """Split a batch into multiple inference batches and insert individual prompts into queues and mapping."""
-    # Split the batch over the VLLM engines.
-    inference_batch_size = len(batch.queries) // vllm_num_engines
+    import math
```
hamishivi, line 1899: should the import be here or at top of file?

hamishivi (approved):
LGTM, tests seem good.

PR #919, opened 2025-08-19
default resumable commnand to be resumable

saurabh111233212:


mason.py
```diff
@@ -206,6 +206,18 @@ def get_args():
...
+
+    def _commands_include_resumable_target(cmds: List[List[str]]) -> bool:
+        for cmd in cmds:
+            for target in OPEN_INSTRUCT_RESUMABLES:
+                if target in cmd:
+                    return True
+        return False
+
+    # can resume if the command is in OPEN_INSTRUCT_RESUMABLES and --non_resumable is not set
+    is_resumable = _commands_include_resumable_target(commands) and not mason_args.non_resumable
```
hamishivi, line 218: could you add a log line that resumable is set to false but `non_resumable` was not set?

hamishivi (approved):

mason.py
saurabh111233212: yep

PR #921, opened 2025-08-20
Updated workflows to prevent duplicates.

finbarrtimbers:
I'm not entirely sure what I'm doing here; this is mostly trial and error. The core issue I'm trying to solve is that all of our workflows are triggered twice when we merge a PR to main: 

Screenshot 2025-08-20 at 8 17 28 AM

I think this will fix it? The core issue is that we trigger on both `merge_group` and `push` actions, but the only time we _ever_ have a `push` action to the main branch is right after a `merge_group` event. So we can just remove that trigger.


mnoukhov (approved):
also not sure whether this will fix it but it makes sense

PR #923, opened 2025-08-20
logging oe eval results to wandb when using new oe-eval-interal

mnoukhov:
requires PR https://github.com/allenai/oe-eval-internal/pull/636 but only locally so you can update your local ./oe-eval-internal with it and it'll work fine

its put behind a flag `--oe_eval_log_to_wandb` because it requires adding your beaker secret to `user_WANDB_API_KEY`. Ideally, we should check that this beaker secret exists for the user if this flag is set and throw an error otherwise

results are logged to wandb.run.summary currently e.g. https://wandb.ai/ai2-llm/open_instruct_internal/runs/hw5ish22/overview

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -396,6 +396,8 @@ class Args:
     """the docker image for evaluation for oe-eval"""
     eval_priority: Literal["low", "normal", "high", "urgent"] = "normal"
     """the priority of auto-launched evaluation jobs"""
+    oe_eval_log_to_wandb: bool = False
+    """Whether to log oe-eval results to wandb as run.summary"""
 
     # Evaluation behavior
     eval_on_step_0: bool = False
@@ -1093,6 +1095,7 @@ class PolicyTrainerRayProcess(RayProcess):
                 args.gs_bucket_path,
                 args.eval_priority,
                 args.oe_eval_beaker_image,
+                args.oe_eval_log_to_wandb,
             )
 
 
--- a/open_instruct/utils.py
+++ b/open_instruct/utils.py
@@ -1050,6 +1050,7 @@ def launch_ai2_evals_on_weka(
     gs_bucket_path: Optional[str] = None,
     eval_priority: Optional[str] = "normal",
     beaker_image: Optional[str] = None,
+    log_eval_to_wandb: bool = False,
 ) -> None:
     weka_cluster = "ai2/saturn-cirrascale ai2/neptune-cirrascale"
     gcp_cluster = "ai2/augusta-google-1"
@@ -1091,6 +1092,9 @@ python scripts/submit_eval_jobs.py \
 --skip_oi_evals"""
     if wandb_url is not None:
         command += f" --run_id {wandb_url}"
+        if log_eval_to_wandb:
+            wandb_run_path = wandb_url_to_run_path(wandb_url)
+            command += f" --wandb_run_path {wandb_run_path}"
     if oe_eval_max_length is not None:
         command += f" --oe_eval_max_length {oe_eval_max_length}"
     if training_step is not None:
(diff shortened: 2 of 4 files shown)


open_instruct/grpo_fast.py
```diff
@@ -407,6 +407,8 @@ class Args:
     """the docker image for evaluation for oe-eval"""
     eval_priority: Literal["low", "normal", "high", "urgent"] = "normal"
     """the priority of auto-launched evaluation jobs"""
+    oe_eval_log_to_wandb: bool = False
```
finbarrtimbers: Why would we ever want this to be false? (unless wandb isn't enabled)

open_instruct/utils.py
```diff
@@ -1197,6 +1201,27 @@ def launch_ai2_evals_on_weka(
     print(f"Submit jobs after model training is finished - process return code: {process.returncode}")
 
 
+def wandb_url_to_run_path(url):
```
finbarrtimbers: Can you add type annotations? `def wandb_url_to_run_path(url: str) -> str:`

Can you also add 2-3 [doctest](https://docs.python.org/3/library/doctest.html) style examples and some quick tests in test_utils.py? i.e.

>>> wandb_url_to_run_path("https://wandb.ai/ai2-llm/open_instruct_internal/runs/5nigq0mz")
"ai2-llm/open_instruct_internal/5nigq0mz"

and then a test that's just

```
@parameterized.parameterized(
("wandb_url1", "expected_wandb_run_path1"),
("wandb_url2", "expected_wandb_run_path2"),
("wandb_url3", "expected_wandb_run_path3"),
)
def test_wandb_url_to_run_path(self, url, expected_run_path):
    self.assertEqual(wandb_url_to_run_path(url), expected_run_path)
```

but with actual examples.

finbarrtimbers (changes requested):

open_instruct/grpo_fast.py
finbarrtimbers: ah sorry just read this: 

> its put behind a flag --oe_eval_log_to_wandb because it requires adding your beaker secret to user_WANDB_API_KEY. Ideally, we should check that this beaker secret exists for the user if this flag is set and throw an error otherwise

Can you check that the secret exists instead?

scripts/eval/oe-eval.sh
```diff
@@ -122,12 +124,20 @@ if [[ -n "$STOP_SEQUENCES" ]]; then
...
+    WANDB_ARG="--wandb-run-path $WANDB_RUN_PATH --gantry-secret-wandb-api-key ${beaker_user}_WANDB_API_KEY"
+fi
+
 DATALAKE_ARGS=""
 if [[ -n "$RUN_ID" ]]; then
     DATALAKE_ARGS+="run_id=$RUN_ID"
 fi
 if [[ -n "$STEP" ]]; then
     DATALAKE_ARGS+=",step=$STEP"
+    WANDB_ARG+="--wandb-run-step $STEP"
```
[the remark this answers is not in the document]
mnoukhov: fixed

open_instruct/grpo_fast.py
mnoukhov: I set this to default true and also check for the secret in `oe-eval.sh` which will throw a warning if it doesn't exist and won't log to wandb

open_instruct/utils.py
mnoukhov: added tests and docstring, let me know if this looks good

open_instruct/grpo_fast.py
finbarrtimbers: Why not remove the flag entirely and just check for that in the env?

mnoukhov: Did it!

PR #928, opened 2025-08-21
Remove duplicate olmo-instruct templates

hamishivi:
We had two olmo_thinker / olmo_thinker_r1_templates in the codebase for some reason. This removes one set. They were identical minus one minor typo in thinker_r1, which I have fixed.

### DIFF
--- a/open_instruct/dataset_transformation.py
+++ b/open_instruct/dataset_transformation.py
@@ -448,7 +448,7 @@ CHAT_TEMPLATES = {
         "The user asks a question, and the Assistant solves it. "
         "The assistant first thinks about the reasoning process in "
         "the mind and then provides the user with the answer. "
-        "The reasoning process and answer is enclosed within <think> </think> "
+        "The reasoning process and answer are enclosed within <think> </think> "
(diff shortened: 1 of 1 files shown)


open_instruct/dataset_transformation.py
```diff
@@ -532,7 +448,7 @@ def visualize_token_role(tokens: list[int], masks: list[int], tokenizer: PreTrai
         "The user asks a question, and the Assistant solves it. "
         "The assistant first thinks about the reasoning process in "
         "the mind and then provides the user with the answer. "
-        "The reasoning process and answer are enclosed within <think> </think> "
```
finbarrtimbers: I think "are" is correct?

finbarrtimbers (approved):

open_instruct/dataset_transformation.py
finbarrtimbers: I'm fine with whatever you decide, feel free to merge whenever.

hamishivi: whoops, yeah

PR #931, opened 2025-08-21
add olmo3 chat_template compatibility for user query extraction

fabrahman:
Modified the user query extraction logic to work with all `tulu/olmo_thinker` and `tulu/olmo_thinker_r1_style`.


hamishivi:
Closing in favour of #933

PR #937, opened 2025-08-22
Don't block main thread or trainer on weight update to generator

mnoukhov:
Single GPU run: [Wandb](https://wandb.ai/ai2-llm/open_instruct_internal/runs/3zs04fyu?nw=nwuserfinbarrt) ([baseline](https://wandb.ai/ai2-llm/open_instruct_internal/runs/hvswj753?nw=nwuserfinbarrt))
Multi-node run: [Wandb](https://wandb.ai/ai2-llm/open_instruct_internal/runs/jd3au9w5?nw=nwuserfinbarrt) ([baseline](https://wandb.ai/ai2-llm/open_instruct_internal/runs/yo6bdbq9?nw=nwuserfinbarrt))

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -50,6 +50,7 @@ from open_instruct import utils
 import asyncio
 import json
 import math
+import random
 import shutil
 import socket
 import threading
@@ -435,15 +436,25 @@ class Args:
             raise ValueError("`checkpoint_state_dir` must be provided if `checkpoint_state_freq` is greater than 0!")
         if self.checkpoint_state_dir is not None and self.checkpoint_state_freq == -1:
             raise ValueError("`checkpoint_state_freq` must be greater than 0 if `checkpoint_state_dir` is provided!")
+
+        if self.gs_checkpoint_state_dir is not None and not self.gs_checkpoint_state_dir.startswith("gs://"):
+            raise ValueError(f"`gs_checkpoint_state_dir` must start with 'gs://', got: {self.gs_checkpoint_state_dir}")
+        if self.gs_bucket_path is not None and not self.gs_bucket_path.startswith("gs://"):
+            raise ValueError(f"`gs_bucket_path` must start with 'gs://', got: {self.gs_bucket_path}")
+
         if self.gs_bucket_path is not None and self.gs_checkpoint_state_dir is None:
+            if self.checkpoint_state_dir is None:
+                raise ValueError("`checkpoint_state_dir` must be provided when using `gs_bucket_path`!")
+            checkpoint_dir_name = self.checkpoint_state_dir.rstrip("/")
             beaker_users = get_beaker_whoami()
             if beaker_users is not None:
-                self.gs_checkpoint_state_dir = f"{self.gs_bucket_path}/{beaker_users}/{self.checkpoint_state_dir}"
+                self.gs_checkpoint_state_dir = f"{self.gs_bucket_path}/{beaker_users}/{checkpoint_dir_name}"
             else:
-                self.gs_checkpoint_state_dir = f"{self.gs_bucket_path}/{self.checkpoint_state_dir}"
-        if self.gs_checkpoint_state_dir is not None:
-            download_latest_checkpoint_from_gs(self.gs_checkpoint_state_dir, self.checkpoint_state_dir)
+                self.gs_checkpoint_state_dir = f"{self.gs_bucket_path}/{checkpoint_dir_name}"
+
         if self.checkpoint_state_dir is not None:
+            if self.gs_checkpoint_state_dir is not None:
+                download_latest_checkpoint_from_gs(self.gs_checkpoint_state_dir, self.checkpoint_state_dir)
             calibrate_checkpoint_state_dir(self.checkpoint_state_dir)
         if self.tools is not None and len(self.tools) > 0:
             for tool in self.tools:
@@ -533,6 +544,20 @@ class ShufflingIterator:
 
         return batch
 
+    def get_state(self) -> Dict[str, Any]:
+        """Get the current state of the iterator for checkpointing."""
+        return {
+            "index": self.index,
+            "data": self.data.copy(),  # Current shuffled order
+            "rng_state": self.rng.bit_generator.state,
+        }
+
+    def set_state(self, state: Dict[str, Any]) -> None:
+        """Restore the iterator state from a checkpoint."""
+        self.index = state["index"]
+        self.data = state["data"].copy()
+        self.rng.bit_generator.state = state["rng_state"]
+
 
 @ray.remote(num_gpus=1)
 class PolicyTrainerRayProcess(RayProcess):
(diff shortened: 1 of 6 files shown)


open_instruct/grpo_fast.py
```diff
@@ -1849,11 +1853,12 @@ def create_model_and_optimizer(
     if resume_training_step > 1:
         logger.info(f"Resuming training from step {resume_training_step}... Broadcasting weights to vLLM engines.")
         with Timer("[Main Thread] 🔄 Loading weights using shared memory"):
-            ray_get_with_progress(
+            weight_sync_future = ray_get_with_progress(
```
finbarrtimbers: I don't think this works? `ray_get_with_progress` returns the result of `ray.get`ting the reference, not a future (or a ray reference). If you keep the code as-is It should do what you want.

Also, do we even need this first call to broadcast the weights given that we have the weight sync thread? We can just trigger that immediately and get rid of this, right?

open_instruct/grpo_fast.py
```diff
@@ -1921,36 +1926,24 @@ def split_and_insert_batch(
...
 
-        ray_get_with_progress(
-            [m.broadcast_to_vllm.remote() for m in policy_group.models],
-            desc=f"[Main thread] Broadcasting weights to vLLM engines at training step {training_step}",
-            enable=args.verbose,
-        )
-
-        ray.get(actor_manager.set_should_stop.remote(False))
-        logger.debug(f"[Main Thread] Set should_stop to False after weight sync at step {training_step}")
+    # Trigger weight sync in the weight sync thread
```
finbarrtimbers: I think this comment is redundant because of your excellent variable naming.

open_instruct/grpo_fast.py
```diff
@@ -1921,36 +1926,24 @@ def split_and_insert_batch(
...
+    weight_sync_trigger_event._training_step = training_step  # Store training step for the thread
```
finbarrtimbers: This feels hacky. Could we do something else? Maybe have a state dict that's shared between the two threads? (setting/getting key values is atomic in Python).

open_instruct/grpo_fast.py
```diff
@@ -1921,36 +1926,24 @@ def split_and_insert_batch(
         )
 
 
-def sync_weights_and_prepare_prompts(
+def trigger_weight_sync_and_prepare_prompts(
```
finbarrtimbers: Should this just be "prepare prompts"? 

We can just call `weight_sync_trigger_event.set()` in the main thread at the appropriate time. If it's this simple, no point pushing it down into a function.

open_instruct/grpo_fast.py
```diff
@@ -1983,6 +1976,60 @@ def load_data_from_packing_thread(packed_sequences_Q: Queue, num_total_tokens: i
...
+        with Timer("[Weight Sync]") as timer:
+            logger.debug(f"[Weight Sync Thread] Starting weight sync for step {training_step}")
+
+            # Set actors to stop
+            ray.get(actor_manager.set_should_stop.remote(True))
+            logger.debug(f"[Weight Sync Thread] Set should_stop to True for weight sync at step {training_step}")
+
+            # Broadcast weights to vLLM engines
+            # First get the futures (which return lists of ObjectRefs)
+            weight_broadcast_futures = ray_get_with_progress(
```
finbarrtimbers: Same comment as above. I don't think this is correct. 

Could do

```
weight_broadcast_refs = [m.broadcast_to_vllm.remote() for m in policy_group.models] 
```

and then build a list of weight refs as you did earlier.

open_instruct/grpo_fast.py
```diff
@@ -1983,6 +1976,60 @@ def load_data_from_packing_thread(packed_sequences_Q: Queue, num_total_tokens: i
...
+                [m.broadcast_to_vllm.remote() for m in policy_group.models],
+                desc=f"[Weight Sync Thread] Starting weight broadcast to vLLM engines at training step {training_step}",
+                enable=args.verbose,
+            )
+
+            # Flatten the list of ObjectRef lists and wait for all weight updates to complete
+            all_weight_refs = []
+            for refs_list in weight_broadcast_futures:
```
finbarrtimbers: I'm a tad confused as to the type of refs_list; so it's a `list[list[ray.ObjectRef]]`? Can you add a type annotation for that? I find it hard to follow. I assumed it was a `list[ray.ObjectRef]`.

open_instruct/grpo_fast.py
```diff
@@ -1983,6 +1976,60 @@ def load_data_from_packing_thread(packed_sequences_Q: Queue, num_total_tokens: i
...
+                all_weight_refs.extend(refs_list)
+
+            if all_weight_refs:
```
finbarrtimbers: Don't bother with the conditional, as if it's empty, the call to `ray_get_with_progress` won't do anything (iterating over an empty list).

open_instruct/grpo_fast.py
```diff
@@ -1983,6 +1976,60 @@ def load_data_from_packing_thread(packed_sequences_Q: Queue, num_total_tokens: i
...
+                    desc=f"[Weight Sync Thread] Waiting for weight updates to complete at training step {training_step}",
+                    enable=args.verbose,
+                )
+
+            # Allow actors to resume
+            ray.get(actor_manager.set_should_stop.remote(False))
+            logger.debug(f"[Weight Sync Thread] Set should_stop to False after weight sync at step {training_step}")
+
+        try:
+            weight_sync_metrics_Q.put_nowait({"time/weight_sync": timer.duration})
```
finbarrtimbers, line 2122: Nice use of `put_nowait`.

open_instruct/grpo_fast.py
mnoukhov: Great points
1. got rid of this extra call and just made weight_sync_thread do a sync if resume_training_step > 1
2. you're right I don't need the outside `ray_get` so removed it

open_instruct/grpo_fast.py
mnoukhov: We're only really using training_step for saying the training step during logging, so I just removed it

open_instruct/grpo_fast.py
mnoukhov: you're right, simplified it and put the trigger in the main thread

open_instruct/grpo_fast.py
mnoukhov: that works better, done!

open_instruct/grpo_fast.py
mnoukhov: since we don't have a `ray_get` anymore, this ends up being just a List[ray.ObjectRef] and is simpler

open_instruct/grpo_fast.py
mnoukhov: done

mnoukhov:
Test runs are good!
- single gpu https://wandb.ai/ai2-llm/open_instruct_internal/runs/3zs04fyu
- multinode https://wandb.ai/ai2-llm/open_instruct_internal/runs/jd3au9w5 

Multinode is marked as finished on wandb but crashed on beaker. This seems unrelated to my change and something to do with ray cleanup https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01K3HG59VXFV401705MGCZ6GKQ

finbarrtimbers:
I updated the PR description with the experiment links (and baselines). I think they look good but I asked @hamishivi via Slack to take a look as well. If he gives the thumbs up I think we're good to merge.

open_instruct/grpo_fast.py
```diff
@@ -2477,13 +2526,13 @@ def run_training(
...
-        generate_metrics = {}
-        try:
-            generate_metrics = generate_metrics_Q.get_nowait()
-        except Empty:
-            logger.info("[Main Thread] didn't get generation metrics")
+        for metrics_Q in [generate_metrics_Q, weight_sync_metrics_Q]:
+            try:
+                metrics = metrics_Q.get_nowait()
+            except Empty:
+                logger.info("[Main Thread] didn't get generation metrics")
```
hamishivi: nit: now this log should say which metrics queue it didnt get.

mnoukhov: fixed!

PR #947, opened 2025-08-26
Removed explicit cache reset in favour of cache_salt.

finbarrtimbers:
This way, the cache is only reused by requests with the same `cache_salt` value, and as `cache_salt` is a function of the training step and the prompt index (within the batch), it won't get reused across training steps. 

Runs:

1. Single GPU run: [Beaker](https://beaker.org/ex/01K3KF2GDJ7MVRRSQNFZYYG4AX)
2. Multi-node run: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01K3KFSS8Q52NQ27J1056AZMQE?)

Ran for 670 steps: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01K3KSENQPKYPNZ02TTJ6SWS2K?). Manual testing: 

> Q: What is 2+2?
> A: '\n\n</think> <answer> 4 </answer>'
> Q: what is the capital of paris?
> A: '\nOkay, the user is asking for the capital of Paris. I know that'
> Q: what is the capital of france?
> A: '\n\n</think> <answer> The capital of France is Paris. Paris is the largest'
> Q: "how many letter Rs in the word strawberry?"
> A: '\n\n</think> <answer> The word "strawberry" has one \'r'

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -421,6 +421,8 @@ class Args:
     def __post_init__(self):
         if os.environ.get("VLLM_USE_V1") == "0":
             logger.warning("When using the v0 version of vLLM, caching is broken and will never be invalidated.")
+            if self.vllm_enable_prefix_caching:
+                raise ValueError("Prefix caching is currently not supported for v0.")
         assert self.num_samples_per_prompt_rollout > 0, "Number of samples per prompt must be greater than 0!"
         if self.num_samples_per_prompt_rollout == 1:
             logger.warning("num_samples_per_prompt_rollout is 1. This reduces GRPO to REINFORCE.")


open_instruct/vllm_utils3.py
```diff
@@ -436,12 +436,12 @@ def _add_initial_requests(self, prompts, sampling_params, n_samples, training_st
...
                 for j in range(n_samples):
                     request_id = f"{training_step}_{i}-{j}"
-                    tokens_prompt = vllm.TokensPrompt(prompt_token_ids=prompt)
+                    tokens_prompt = vllm.TokensPrompt(prompt_token_ids=prompt, cache_salt=f"{training_step}_{i}")
                     self.llm_engine.add_request(request_id, tokens_prompt, sampling_params)
             else:
                 # Standard request format for non-tool mode
                 request_id = f"batch_{training_step}_{i}"
-                tokens_prompt = vllm.TokensPrompt(prompt_token_ids=prompt)
+                tokens_prompt = vllm.TokensPrompt(prompt_token_ids=prompt, cache_salt=request_id)
```
hamishivi, line 465: one thing: apparently this is not supported in V0 vllm... we use V1 by default but I'm curious if this just breaks if we set V0 backend? Should we keep the explicit resets as a flag if the user specifies and wants to use V0?

finbarrtimbers: How does the user set the v0 backend?

hamishivi: `VLLM_USE_V1=0`: https://docs.vllm.ai/en/latest/usage/v1_guide.html
But I think its sort of fair to say it isnt supported.

finbarrtimbers: I think I'll add a check to see if that's set and then error. Wdyt?

hamishivi: Maybe just log it? e.g. for fp8 cache it requires v0 engine, so for now I'd rather it just warn the user that this isnt really a supported feature.

finbarrtimbers: Done.

open_instruct/vllm_utils3.py
```diff
@@ -452,17 +454,22 @@ def _process_request(self, request):
 
     def _add_initial_requests(self, prompts, sampling_params, n_samples, training_step):
         """Add initial requests to the engine."""
+        if os.environ.get("VLLM_USE_V1") == "0":
```
hamishivi: You could check this in the args post_init and then assert that `enable_prefix_caching` is false

scripts/train/debug/single_gpu_on_beaker.sh
hamishivi, line 1: remove these changes

open_instruct/vllm_utils3.py
finbarrtimbers: Nice! Done.

hamishivi (approved):

PR #948, opened 2025-08-26
Adds try/except to guard against beaker errors, and adds comment describing compile issue.

finbarrtimbers:
The try/except is to prevent Beaker service issues from taking down experiments. 

The comment is to prevent future us from enabling the cache if it's still broken.


hamishivi (approved):
LGTM! At this point most of maybe_update_beaker_description is wrapped in try/excepts, haha.

finbarrtimbers:
I feel like Claude!

PR #949, opened 2025-08-26
Fix ray bundles indices, max len config.

hamishivi:
Two changes:
1. Allow configurable max length for max len verifier. This makes running experiments easier.
2. Properly assign bundle indices for the vllm engines. This prevents unnecessary communication when TP>1. Turns out in the past we could be doing inter-node communication for this case!

Testing now (tested on a diff branch with some other changes before).

### DIFF
--- a/open_instruct/vllm_utils3.py
+++ b/open_instruct/vllm_utils3.py
@@ -47,6 +47,24 @@ from open_instruct.utils import ray_get_with_progress
 logger = logger_utils.setup_logger(__name__)
 
 
+# From: https://github.com/OpenRLHF/OpenRLHF/pull/971/files
+# Turns out Ray doesnt necessarily place bundles together,
+# so this function is used to get the bundle indices of a placement group
+# and ensure that the bundles placed on the same node are grouped together.
+# avoids unnecessary communication for TP>1 with vllm.
+def get_bundle_indices(placement_group, index, length):
+    import ray
+
+    pg_infos = ray.util.placement_group_table(placement_group)
+
+    node_id_to_bundles = {}
+    for bundle, node_id in pg_infos["bundles_to_node_id"].items():
+        node_id_to_bundles.setdefault(node_id, []).append(bundle)
+
+    sorted_bundle_indices = sum(node_id_to_bundles.values(), [])
+    return sorted_bundle_indices[index * length : (index + 1) * length]
+
+
 def _init_tool_tracking():
     """Initialize tracking variables for tool mode."""
     return {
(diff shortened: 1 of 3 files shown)


open_instruct/vllm_utils3.py
```diff
@@ -47,6 +47,24 @@
 logger = logger_utils.setup_logger(__name__)
 
 
+# From: https://github.com/OpenRLHF/OpenRLHF/pull/971/files
+# Turns out Ray doesnt necessarily place bundles together,
+# so this function is used to get the bundle indices of a placement group
+# and ensure that the bundles placed on the same node are grouped together.
+# avoids unnecessary communication for TP>1 with vllm.
+def get_bundle_indices(placement_group, index, length):
+    import ray
```
finbarrtimbers: Imports at top of the file?

open_instruct/vllm_utils3.py
```diff
@@ -47,6 +47,24 @@
 logger = logger_utils.setup_logger(__name__)
 
 
+# From: https://github.com/OpenRLHF/OpenRLHF/pull/971/files
+# Turns out Ray doesnt necessarily place bundles together,
+# so this function is used to get the bundle indices of a placement group
+# and ensure that the bundles placed on the same node are grouped together.
+# avoids unnecessary communication for TP>1 with vllm.
+def get_bundle_indices(placement_group, index, length):
```
finbarrtimbers: Can we get type annotations here?

open_instruct/vllm_utils3.py
```diff
@@ -47,6 +47,24 @@
...
+    import ray
+
+    pg_infos = ray.util.placement_group_table(placement_group)
+
+    node_id_to_bundles = {}
```
finbarrtimbers: make it a defaultdict?

open_instruct/vllm_utils3.py
```diff
@@ -47,6 +47,24 @@
...
+    for bundle, node_id in pg_infos["bundles_to_node_id"].items():
+        node_id_to_bundles.setdefault(node_id, []).append(bundle)
+
+    sorted_bundle_indices = sum(node_id_to_bundles.values(), [])
```
finbarrtimbers: Why is this sorted?

open_instruct/vllm_utils3.py
```diff
@@ -47,6 +47,24 @@
...
+    return sorted_bundle_indices[index * length : (index + 1) * length]
```
finbarrtimbers: Isn't this function super wasteful? Couldn't we just create the ones we need in the for loop above?

open_instruct/vllm_utils3.py
hamishivi: sorted in the sense that now all bundles from the same node are adjacent -- cleared this up in next commit!

open_instruct/vllm_utils3.py
hamishivi: fair enough, was being lazy and just copying openrlhf. Made some changes to the flow and hopefully now its a bit clearer!

open_instruct/vllm_utils3.py
```diff
@@ -47,6 +47,25 @@
...
+# and ensure that the bundles placed on the same node are grouped together.
+# avoids unnecessary communication for TP>1 with vllm.
+def get_bundle_indices_list(placement_group: ray.util.placement_group) -> List[int]:
+    pg_infos = ray.util.placement_group_table(placement_group)
+
+    node_id_to_bundles = defaultdict(list)
+    for bundle, node_id in pg_infos["bundles_to_node_id"].items():
+        node_id_to_bundles[node_id].append(bundle)
+
+    # create a flattened list of bundles we can use.
```
finbarrtimbers: Redundant comment.

PR #951, opened 2025-08-26
Update SFT Gantry command

tyler-romero:
Gantry v3 has built in uv support so we're updating the command


scripts/data/convert_sft_data_for_olmocore.py
```diff
@@ -1,31 +1,57 @@
...
-        --gpus 1 \
-        --priority high \
-        -- /root/.local/bin/uv run python scripts/data/convert_sft_data_for_olmocore.py \
-        --dataset_mixer_list allenai/tulu-3-sft-olmo-2-mixture 1.0 \
-        --tokenizer_name_or_path allenai/OLMo-2-1124-7B \
-        --output_dir /weka/oe-training-default/ai2-llm/tylerr/data/sft/tulu-3-sft-olmo-2-mixture-0225-olmocore \
-        --visualize True \
-        --chat_template_name olmo \
-        --max_seq_length 16384
+        --install "echo 'do nothing'" \
```
tyler-romero, line 41: kind of weird but its a hack to avoid installing all of the project dependencies - which is a bit painful for open-instruct since it includes flash-attn, vllm, etc

jacob-morrison (approved):

PR #952, opened 2025-08-27
fix context truncation tokenization and add tests

fabrahman:
This PR contains:

1.  fixes for using llm judge tokenizer for context truncation.
2. Additionally, the `test_truncate_simple.py` script covers some basic tests to verify the logic/functionality of the context length checker.


test_truncate_simple.py
```diff
@@ -0,0 +1,555 @@
...
+    test_truncate_long_messages()
+    test_empty_messages()
+    test_only_system_messages()
+    test_real_qwen3_tokenizer()
+    test_hosted_vllm_prefix_handling()
+    
+    # Test integration between functions
+    test_context_window_integration()
+    
+    print("\n✅ All tests completed!")
```
hamishivi: This isn't useful logging, could we remove this?

test_truncate_simple.py
```diff
@@ -0,0 +1,555 @@
...
+            model_name="gpt2",
+            max_context_length=1024,
+            safety_margin=50
+        )
+        
+        print(f"   After truncation fits: {fits_after}")
+        print(f"   Messages before: {len(long_messages)}, after: {len(truncated)}")
+        
+        # The truncated version should fit
+        assert fits_after, "Truncated messages should fit within context window"
```
hamishivi: This assert fails for me when I ran testing? does it work for you?

test_truncate_simple.py
hamishivi, line 1: For the tests, could you make it use unittest like s.t. the style is similar to https://github.com/allenai/open-instruct/blob/main/open_instruct/test_grpo_fast.py ?

hamishivi (commented):
I think the core funcationality is solid but the tests should better align with the other tests in the repo. Ideally we run them as part of the regular test CI.

PR #958, opened 2025-08-28
Fix UnboundLocalError where metrics weren't defined

finbarrtimbers:
I think that we weren't hitting this bug because we were never triggering the case where the queue was empty. 

Single GPU run: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01K3RYST28QVYMQMTCD8AGJHSA?)


mnoukhov (approved):
damn, good catch

PR #973, opened 2025-08-30
Caches `should_stop` so we're not hammering Ray.

finbarrtimbers:
As part of https://github.com/allenai/open-instruct/pull/859, I made a number of major changes. To be a better software engineer, I'm breaking them out into separate PRs (and also because there's a bug in https://github.com/allenai/open-instruct/pull/859 that I can't identify).

This caches the value of `should_stop` so we're not constantly hammering Ray. 

Single GPU run: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01K3WNP6JNWZ8ZKW0T4P2BXXR2?).

### DIFF
--- a/open_instruct/vllm_utils3.py
+++ b/open_instruct/vllm_utils3.py
@@ -18,6 +18,7 @@
 import copy
 import os
 import queue
+import time
 from collections import defaultdict
 from concurrent.futures import ThreadPoolExecutor
 from datetime import timedelta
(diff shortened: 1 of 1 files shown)


open_instruct/vllm_utils3.py
```diff
@@ -368,17 +368,30 @@ def __init__(
...
+
+    def _should_stop(self) -> bool:
+        if (time.perf_counter() - self._last_should_stop_update) > self._should_stop_timeout_s:
+            should_stop_ref = self.actor_manager.should_stop.remote()
+            ready_refs, _ = ray.wait([should_stop_ref], timeout=0.1)
+            if ready_refs:
+                self._should_stop_value = ray.get(ready_refs[0])
+                self._last_should_stop_update = time.perf_counter()
+            else:
+                ray.cancel(should_stop_ref)
```
hamishivi, line 384: Do you think its worth adding some logging here? It seems like if the actor manager doesn't return at all / times out for this call its a sign something is a bit off.

hamishivi (approved):
Looks good to me, gave it a quick small test. One comment about some logging on a weird case.

PR #974, opened 2025-08-30
Fix inference_batch_size calculation.

finbarrtimbers:
Asked Claude to move some code, didn't check it thoroughly enough. Whoops!


hamishivi (approved):
whoops, yeah ncie catch.

PR #975, opened 2025-08-31
is_local_main_process -> is_main_process in finetune.py

jacob-morrison:
This resolves a race condition when cleaning up checkpoints during multinode training; just let the main process handle it

### DIFF
--- a/open_instruct/finetune.py
+++ b/open_instruct/finetune.py
@@ -956,7 +956,7 @@ def main(args: FlatArguments, tc: TokenizerConfig):
                             os.path.join(get_last_checkpoint_path(args, incomplete=True), "COMPLETED"), "w"
                         ) as f:
                             f.write("COMPLETED")  # annoyingly, empty files arent uploaded by beaker.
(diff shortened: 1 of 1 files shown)


open_instruct/finetune.py
```diff
@@ -971,7 +971,7 @@ def main(args: FlatArguments, tc: TokenizerConfig):
             # use this to mark the checkpoint as completely saved, to avoid restoring from garbled checkpoints
             with open(os.path.join(get_last_checkpoint_path(args, incomplete=True), "COMPLETED"), "w") as f:
```
hamishivi, line 972: I agree with gemini-bot here, we should probably put this completed write under the main process too.

open_instruct/finetune.py
```diff
@@ -956,7 +956,7 @@ def main(args: FlatArguments, tc: TokenizerConfig):
                             os.path.join(get_last_checkpoint_path(args, incomplete=True), "COMPLETED"), "w"
```
hamishivi, line 956: I agree with gemini-bot here, we should probably put this completed write under the main process too.

hamishivi (commented):
Looks good, one minor comment. Also, it would be good to link a beaker job with a multi-node run (can be a small job) that works correctly.

PR #965, opened 2025-08-29
load pretokenized user query (v0)

fabrahman:
This is the early pr for the pre tokenized user query. It leads to some `2025-08-29 17:03:30 - WARNING - grpo_fast.py:2076 - [Main Thread] Timeout waiting for packed sequences. Retrying...
2025-08-29 17:04:00 - WARNING - grpo_fast.py:2076 - [Main Thread] Timeout waiting for packed sequences. Retrying...` error.

@hamishivi can you TAK?

PR #987, opened 2025-09-04
Fix finetune keys

hamishivi:
#965 added these to the wrong place. We should add a test for the finetune script.


open_instruct/dataset_transformation.py
```diff
@@ -878,6 +872,8 @@ def sft_tokenize_v1(
     row[ATTENTION_MASK_KEY] = [1] * len(row[INPUT_IDS_KEY])
     labels = copy.deepcopy(row[INPUT_IDS_KEY])
     row[LABELS_KEY] = labels
+    # drop the raw_prompt key if it exists
+    row.pop(RAW_PROMPT_KEY, None)
```
finbarrtimbers, line 875: I don't think you need the comment as it's basically equivalent to the code

PR #992, opened 2025-09-05
adding PuzzleMatcherVerifier

tengxiao1:
Verifier for puzzle tasks that require string matching. It checks whether the model output matches the ground-truth answer.

Verified using TTTXXX01/All_Puzzles.

### DIFF
--- a/open_instruct/ground_truth_utils.py
+++ b/open_instruct/ground_truth_utils.py
@@ -419,6 +419,7 @@ class StringMatcherVerifier(VerifierFunction):
         return VerificationResult(score=score)
 
 
+
 class F1Verifier(VerifierFunction):
     """
     Verifier that computes the string F1 score between the prediction and the label.
@@ -438,6 +439,26 @@ class F1Verifier(VerifierFunction):
         score = f1_score(prediction, label)["f1"]
         return VerificationResult(score=score)
 
+class PuzzleMatcherVerifier(VerifierFunction):
+    """
+    Verifier for Puzzle tasks that require string matching.
+
+    It checks if the model output matches the ground truth answer.
+    """
+
+    def __init__(self, verifier_config: Optional[VerifierConfig] = None) -> None:
+        super().__init__("Puzzle", verifier_config=verifier_config, weight=1.0)
+
+    def __call__(
+        self, tokenized_prediction: List[int], prediction: str, label: str, query: Optional[str] = None
+    ) -> VerificationResult:
+        prediction = prediction.split("</think>")[-1]
+        # remove answer tags from the prediction
+        prediction = prediction.replace("<answer>", "").replace("</answer>", "")
+        # normalize
+        score = float(normalize_answer(prediction) == normalize_answer(label))
+        return VerificationResult(score=score)
+    
 
 class ReSearchVerifierF1(VerifierFunction):
     """


open_instruct/ground_truth_utils.py
```diff
@@ -419,6 +419,7 @@ def __call__(
         return VerificationResult(score=score)
 
 
+
```
finbarrtimbers, line 421: Remove extraneous whitespace please!

open_instruct/ground_truth_utils.py
```diff
@@ -438,6 +439,26 @@ def __call__(
...
         return VerificationResult(score=score)
 
+class PuzzleMatcherVerifier(VerifierFunction):
+    """
+    Verifier for Puzzle tasks that require string matching.
+
+    It checks if the model output matches the ground truth answer.
+    """
+
+    def __init__(self, verifier_config: Optional[VerifierConfig] = None) -> None:
```
finbarrtimbers, line 448: Should verifier_config be none?

open_instruct/ground_truth_utils.py
```diff
@@ -438,6 +439,26 @@ def __call__(
...
+        super().__init__("Puzzle", verifier_config=verifier_config, weight=1.0)
+
+    def __call__(
+        self, tokenized_prediction: List[int], prediction: str, label: str, query: Optional[str] = None
```
finbarrtimbers, line 452: As query is unused, do you need to accept it as an argument? 

If you do, can you prefix it with unused: unused_query

open_instruct/ground_truth_utils.py
```diff
@@ -438,6 +439,26 @@ def __call__(
...
+    ) -> VerificationResult:
+        prediction = prediction.split("</think>")[-1]
+        # remove answer tags from the prediction
+        prediction = prediction.replace("<answer>", "").replace("</answer>", "")
+        # normalize
```
finbarrtimbers: I don't think the comment adds anything.

open_instruct/ground_truth_utils.py
```diff
@@ -438,6 +439,26 @@ def __call__(
...
+    It checks if the model output matches the ground truth answer.
+    """
+
+    def __init__(self, verifier_config: Optional[VerifierConfig] = None) -> None:
+        super().__init__("Puzzle", verifier_config=verifier_config, weight=1.0)
+
+    def __call__(
+        self, tokenized_prediction: List[int], prediction: str, label: str, query: Optional[str] = None
+    ) -> VerificationResult:
+        prediction = prediction.split("</think>")[-1]
```
finbarrtimbers, line 455: Can you comment why this works, maybe with an example? Something like:

# CoT format is <think> lots of thinking tokens </think><answer> the answer </answer>. Here, we remove the CoT to get the answer.

finbarrtimbers (changes requested):

open_instruct/ground_truth_utils.py
```diff
@@ -438,6 +439,26 @@ def __call__(
         score = f1_score(prediction, label)["f1"]
         return VerificationResult(score=score)
 
+class PuzzleMatcherVerifier(VerifierFunction):
```
finbarrtimbers, line 441: Can you also add a quick test? Just call this on a string and verify it has the right answer.

finbarrtimbers: Add the test in `test_ground_truth_utils.py`.

fabrahman, line 441: Why do we need a new class for this verifier if it's essentially doing string matching (string_f1) ?

tengxiao1: Yeah, this one is similar to F1 but uses exact matching. I think the F1 score is meaningless for puzzle questions, which require ordering?

tengxiao1: But we can use F1 if we don’t want to be too strict

open_instruct/ground_truth_utils.py
tengxiao1: Thanks! done

open_instruct/ground_truth_utils.py
tengxiao1: I think so? I was following what the other verifiers do.

open_instruct/ground_truth_utils.py
tengxiao1: Yeah. Removing it now.

open_instruct/ground_truth_utils.py
tengxiao1: Yeah. I agree.  I was following what the other verifiers do. Removing it now

open_instruct/ground_truth_utils.py
tengxiao1: Added `test_ground_truth_utils.py.` Test Summary: 24/24 passed (100.0%)

open_instruct/ground_truth_utils.py
hamishivi: yup, this lets us subclass verifiers with and without configs with the same init signature i think.

open_instruct/ground_truth_utils.py
```diff
@@ -438,6 +439,26 @@ def __call__(
...
+    """
+    Verifier for Puzzle tasks that require string matching.
+
+    It checks if the model output matches the ground truth answer.
+    """
+
+    def __init__(self, verifier_config: Optional[VerifierConfig] = None) -> None:
+        super().__init__("Puzzle", verifier_config=verifier_config, weight=1.0)
```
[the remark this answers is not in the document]
hamishivi: +1 on this -- could we do "puzzle" maybe?

open_instruct/ground_truth_utils.py
hamishivi: I think we still need it to subclass correctly from VerifierFunction? v open to making this interface nicer (maybe we should have some verificationInput class?)

open_instruct/ground_truth_utils.py
hamishivi: Could you use `remove_thinking_section` instead?

open_instruct/ground_truth_utils.py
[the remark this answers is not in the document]
hamishivi: Or actually, maybe we should just call this verifier "exact_match", since thats all its doing? I don't mind.

open_instruct/ground_truth_utils.py
finbarrtimbers, line 421: I don't think this was removed.

finbarrtimbers (approved):

open_instruct/ground_truth_utils.py
tengxiao1: Sure

hamishivi:
Closing in favour of #1003

PR #1007, opened 2025-09-12
Fix olmo crash

hamishivi:
#1006 set `log_stats` to false, but it should be `disable_log_stats` to true


finbarrtimbers:
ahhh nice catch.

PR #1009, opened 2025-09-15
Changes the `build_image_and_launch.sh` script so that it doesn't run with a dirty repo. 

finbarrtimbers:
If we're building a Beaker (Docker) image, we want to make sure that we can trace the code that was in it, so we shouldn't be able to do so with local modifications.


hamishivi (approved):
Good idea, thanks!

PR #1011, opened 2025-09-15
Small refactor to prepare for the `backfill-prompts` PR

finbarrtimbers:
Changelog: 

1. Passes `args.verbose` through to LLMRayActor to silence logs
2. Adds `make_request_id` and `_extract_base_request_id` functions to unify common patterns
3. Changes `_insert_result_to_queue` so that it blocks, and removes the upper bound on queue size for eval results (which was sometimes blocking).
4. Refactors `_process_outputs` to be cleaner, and combines `_process_outputs` with `_process_outputs_with_tools`.

### DIFF
--- a/open_instruct/test_vllm_utils3.py
+++ b/open_instruct/test_vllm_utils3.py
@@ -157,7 +157,11 @@ class TestVllmUtils3(unittest.TestCase):
         mock_final_output.prompt = []
         mock_final_output.prompt_token_ids = []
         mock_final_output.prompt_logprobs = None
-        mock_final_output.outputs = [mock_outputs[0].outputs[0], mock_outputs[1].outputs[0], mock_outputs[2].outputs[0]]
+        mock_final_output.outputs = [
+            mock_outputs[0].outputs[0],
+            mock_outputs[1].outputs[0],
+            mock_outputs[2].outputs[0],
+        ]
         mock_final_output.finished = True
 
         mock_sampling_params = MagicMock(spec=vllm.SamplingParams)


open_instruct/grpo_fast.py
```diff
@@ -2869,8 +2870,6 @@ def main(args: Args, tc: TokenizerConfig, model_config: ModelConfig):
             actor_manager,
             checkpoint_state,
         )
-    except Exception as e:
```
hamishivi, line 2872: Guessing we can remove this because the error bubbles up anyway?

open_instruct/vllm_utils3.py
```diff
@@ -82,6 +82,23 @@ def _init_tool_tracking():
     }
 
 
+def make_request_id(request: PromptRequest) -> str:
+    """Generate a unique tracking key for a request."""
+    prefix = "eval" if request.is_eval else "train"
+    return f"{prefix}_{request.training_step}_{request.dataset_index}"
```
hamishivi, line 88: one thing im wondering about: if we have a really small dataset and try prefilling, we might end up with multiple samples from the same training step in a single step? Maybe we should assert that the maximum batch size is 1 epoch over the dataset.

Although, then, if 1 batch == 1 epoch, and we prefill two steps of prompts, that would be two epochs and we would have clashing ids...

hamishivi (approved):
LGTM! From a quick test seems solid.

One higher-level question abt prompt ids.

open_instruct/grpo_fast.py
finbarrtimbers: yes, exactly. and by catching + logging it, we prevent it from actually causing an error in the main thread, causing the code to exit with status code 0.

PR #1012, opened 2025-09-15
Check we have enough data for bsz/prefill

hamishivi:


open_instruct/grpo_fast.py
```diff
@@ -2773,6 +2773,13 @@ def main(args: Args, tc: TokenizerConfig, model_config: ModelConfig):
     beaker_config, wandb_url = setup_experiment_tracking(args, tc, model_config)
 
     train_dataset, eval_dataset = setup_datasets(args, tc, tokenizer)
+
+    # ensure we have enough prompts for bsz and prefill
```
finbarrtimbers: Remove the comment, I think it's redundant.

PR #1016, opened 2025-09-16
Changes LLMRayActor to use vllm's AsyncLLMEngine

finbarrtimbers:


open_instruct/grpo_fast.py
```diff
@@ -847,7 +847,7 @@ def broadcast_to_vllm(self):
                         shape = param.shape if self.args.deepspeed_stage != 3 else param.ds_shape
                         refs = [
                             engine.update_weight.remote(
-                                name, dtype=param.dtype, shape=shape, empty_cache=count == num_params
+                                name, dtype=str(param.dtype), shape=shape, empty_cache=count == num_params
```
finbarrtimbers, line 876: This is needed to support the async engine's serialization, which for some unknown reason, goes through MessagePack, and not pickle.

open_instruct/vllm_utils_workerwrap.py
```diff
@@ -46,8 +46,10 @@ def init_process_group(
     def update_weight(self, name, dtype, shape, empty_cache=False):
         import torch
 
-        assert dtype == self.model_config.dtype, f"mismatch dtype: src {dtype}, dst {self.model_config.dtype}"
-        weight = torch.empty(shape, dtype=dtype, device="cuda")
+        assert dtype == str(self.model_config.dtype), (
```
finbarrtimbers: As discussed earlier, this is done because of weird serialization format in vLLM's async engine.

pyproject.toml
```diff
@@ -22,14 +22,15 @@ dependencies = [
...
+    "torch>=2.8.0,<2.9",
+    "transformers>=4.55.2", # required by vLLM 0.10.2
+    "vllm==0.10.2",
     "wandb==0.18.1",
     "langdetect==1.0.9",
     "immutabledict==1.2.0",
-    "flash-attn>=2.8.0.post2; platform_system != 'Darwin'",
+    "flash-attn>=2.8.3; platform_system != 'Darwin'",
     "liger-kernel>=0.5.4; platform_system != 'Darwin'",
+    "memray>=1.18.0",
```
hamishivi: Was this an intentional add?

open_instruct/vllm_utils3.py
```diff
@@ -54,6 +51,19 @@
...
+DRAIN_TIMEOUT_S = 300.0
+SHOULD_STOP_CACHE_TIMEOUT_S = 5
+PREFETCH_SLEEP_S = 1
+QUEUE_GET_TIMEOUT_S = 0.1
+COMPLETION_QUEUE_TIMEOUT_S = 1.0
+PROCESS_SLEEP_S = 0.1
+RAY_WAIT_TIMEOUT_S = 0.1
+KV_CACHE_RETRIES = 5
+KV_CACHE_RETRY_SLEEP_S = 0.2
 WEIGHT_UPDATE_SLEEP_INTERVAL_S = 0.1
```
hamishivi: Could we add short explanations of each constant? Some are obvious (weight_update_timeout), but others are a bit less clear (default_workers, kv_cache_retries)

open_instruct/vllm_utils3.py
```diff
@@ -54,6 +51,19 @@
 
 logger = logger_utils.setup_logger(__name__)
 
+DEFAULT_WORKERS = 2
+TOOL_WORKERS = 20
+INIT_TIMEOUT_S = 120
+WEIGHT_UPDATE_TIMEOUT_S = 120
```
hamishivi: How does update time scale as model size scales? How much buffer is left in this timeout when doing e.g. 32b weight sync?

open_instruct/vllm_utils3.py
```diff
@@ -500,454 +659,142 @@ def get_model_dims_dict(self):
...
-        completed_outputs = []
+    def process_from_queue(self) -> int:
+        total_processed = 0
 
-        for req_id, (future, last_o, last_output) in list(tracking["pending_tool_futures"].items()):
-            if not future.done():
-                continue
+        while True:
+            # Health check: ensure prefetch worker is alive. This will raise if it has crashed.
+            if self._prefetch_future.done():
```
hamishivi: do we need this check if `check_background_threads` also is checking the prefetch worker?

open_instruct/vllm_utils3.py
finbarrtimbers: I actually just removed the timeout-- I don't think we actually ever want it. This was really only useful for debugging (where the weight update would sometimes hang forever because of the dtype issue).

pyproject.toml
finbarrtimbers: Ah, no. Removed.

open_instruct/vllm_utils3.py
finbarrtimbers: No, that's a merge error. We should instead be calling `check_background_threads` here (note that it was never actually being called).

open_instruct/vllm_utils3.py
finbarrtimbers: I went over all of these. A bunch of them were redundant and could be removed; others, imo, should be handled by timeouts in `grpo_fast.py` on the actual `ray.get` calls, so I removed them. I think the remaining ones are clear from the names. What do you think?

open_instruct/vllm_utils3.py
```diff
@@ -83,6 +84,159 @@ def assert_threaded_actor(instance):
...
+            break
+
+        current_prompt = vllm.TokensPrompt(prompt_token_ids=prompt_and_tool_output, cache_salt=base_request_id)
+        current_prompt_token_ids = prompt_and_tool_output
+        final_prompt_token_ids = prompt_and_tool_output
+        current_sampling_params = sampling_params.clone()
+        current_sampling_params.max_tokens = new_sample_tokens
+
+    complete_output = vllm.CompletionOutput(
+        index=sub_request_id.split("_")[-1],
```
hamishivi: +1 on the cursor thing here - i think the typing on the `CompletionOutput` class says index should be an int?

open_instruct/vllm_utils3.py
```diff
@@ -500,491 +624,149 @@ def get_model_dims_dict(self):
...
-                    prompt=request_outputs[request_id].prompt,
-                    prompt_token_ids=request_outputs[request_id].prompt_token_ids,
-                    prompt_logprobs=request_outputs[request_id].prompt_logprobs,
-                    outputs=[outputs_by_index[j]],
-                    finished=True,
-                )
-            )
+    def _finalize_completed_request(self, base_request_id: str) -> None:
+        outputs = self.request_outputs[base_request_id]["outputs"]
+        ordered_outs = sorted(outputs, key=lambda x: _extract_base_request_id(x.request_id))
```
hamishivi: Should we sort by the full index (like how the bugbot is suggesting?)

hamishivi (approved):
Great to get this in! I added my fixes from the previous PR to fix tool use. Left two minor comments (from testing, everything works fine anyway).

open_instruct/vllm_utils3.py
finbarrtimbers: yes! Changed.

open_instruct/vllm_utils3.py
finbarrtimbers: ah, yes. nice.

PR #1024, opened 2025-09-19
Removes the `reduce_loss` option, as we don't need to support `reduce_loss="sum"` any longer.

finbarrtimbers:
Originally, we didn't scale by the number of tokens. We then added support to our mean loss implementation to track the number of tokens. Since our mean loss implementation now does track the number of tokens correctly, I have removed the (incorrect) sum loss implementation.

Basically we found [this](https://unsloth.ai/blog/gradient) during the prior summer and @hamishivi added `reduce_loss="sum"` as a cheap hack around it. But now it's all supported much nicer.


hamishivi (approved):
thanks!

PR #1025, opened 2025-09-19
Removes `nvcc` and the "delete big files" steps from our tests as they're slow and not needed now that we don't use `flashinfer`.

finbarrtimbers:
[Test runs](https://github.com/allenai/open-instruct/actions/workflows/tests.yml) would routinely take >10m. Now, they take 4m.

Also removes the `-n auto` argument (and `pytest-xdist` dep) as it was causing the code tests to fail sporadically.


hamishivi (approved):
Looks good thanks!

PR #1029, opened 2025-09-23
Update filtering script

hamishivi:
update the filtering script for the new verifier configs in prep for offline filtering.
Should probably refactor a little in a future to make the verifier args auto update.

### DIFF
--- a/scripts/data/rlvr/filter_existing_dataset_correctness.py
+++ b/scripts/data/rlvr/filter_existing_dataset_correctness.py
@@ -11,6 +11,7 @@ python scripts/data/rlvr/filter_existing_dataset_correctness.py \
 If you have code data, you might have to launch code server too before running:
 source configs/beaker_configs/code_api_setup.sh
 """
+import os
 import argparse
 import json
 from functools import partial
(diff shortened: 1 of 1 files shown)


scripts/data/rlvr/filter_existing_dataset_correctness.py
```diff
@@ -85,7 +86,7 @@ def main():
     )
     parser.add_argument(
         "--code_api_url",
-        default=None,
+        default="http://127.0.0.1:1234/test_program",
```
finbarrtimbers: Is this correct?

scripts/data/rlvr/filter_existing_dataset_correctness.py
```diff
@@ -107,6 +119,12 @@ def main():
         type=int,
         help="Give a max tokens for llm judge."
     )
+    parser.add_argument(
+        "--llm_judge_max_context_length",
+        default=128000,
```
finbarrtimbers: 128_000?

scripts/data/rlvr/filter_existing_dataset_correctness.py
hamishivi: changed to exactly match grpo_fast!

PR #1033, opened 2025-09-23
Adds timing for the individual weight sync events

finbarrtimbers:
Previously, we only timed the entire weight sync event, and not the individual ones. This adds timing for each one.

To enable this, we now record timing of each future in `ray_get_with_progress`.

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -1247,9 +1247,9 @@ class ModelGroup:
         ).remote(world_size, 0, 0, None, None)
 
         self.models.append(master_policy)
-        master_addr, master_port = ray_get_with_progress(
+        (master_addr, master_port), _ = ray_get_with_progress(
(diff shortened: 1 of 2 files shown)


open_instruct/grpo_fast.py
```diff
@@ -2184,8 +2186,17 @@ def weight_sync_thread(
...
 
+        # Calculate distribution statistics
+        sync_time_stats = {
+            "time/weight_sync": timer.duration,
+            "time/weight_sync_mean": np.mean(actor_sync_times),
+            "time/weight_sync_min": np.min(actor_sync_times),
+            "time/weight_sync_max": np.max(actor_sync_times),
+            "time/weight_sync_median": np.median(actor_sync_times),
+        }
+
```
mnoukhov, line 2197: you could consider keeping the actual list and logging a histogram

mnoukhov (approved):
lgtm, suggestion is optional

PR #1035, opened 2025-09-24
Default to tokenizer chat template, configurable random seed.

hamishivi:
1. Default to using the tokenizer chat template instead of the tulu chat template when not specified. I think this is just a more reasonable default in general.
2. Allow the dataset_config_seed to be configured a bit more than fully hardcoded.

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -1153,7 +1153,7 @@ class PolicyTrainerRayProcess(RayProcess):
 
     def save_model(self, output_dir: str, chat_template_name: str, tokenizer: PreTrainedTokenizer) -> None:
         model_to_save = self.model
-        if "olmo" in chat_template_name:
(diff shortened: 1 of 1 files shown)


open_instruct/dataset_transformation.py
```diff
@@ -682,6 +682,8 @@ def get_tokenizer_tulu_v2_2(tc: "TokenizerConfig"):
     if "olmo" in config.model_type:
         if "olmo" in tc.chat_template_name:
             assert not tc.add_bos, "For newer OLMo chat templates, you must *not* run with `--add_bos`."
+        elif tc.chat_template_name == "none":
```
finbarrtimbers: Can we use a different value here? I worry that there will be confusion as to whether it's the string 'none' or if it's the Python object None. Or use None.

hamishivi: Done! just using `None`.

PR #1038, opened 2025-09-25
Logs MFU for the learner to wandb

finbarrtimbers:
Previously we were only calculating MFU for the actors.

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -1372,7 +1372,7 @@ def calculate_utilization_metrics(
     response_lengths: list[int],
     total_generation_time: float,
     samples_per_prompt: int,
-    num_gpus: int,
+    num_inference_gpus: int,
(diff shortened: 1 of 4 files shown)


open_instruct/grpo_fast.py
```diff
@@ -1576,7 +1580,7 @@ def accumulate_inference_batches(
         raw_queries=all_raw_queries,
         indices=None,  # Not meaningful for combined results
     )
-    return combined_result, batch
+    return combined_result, batch, prompt_lengths, response_lengths
```
saurabh111233212, line 1583: i think this change is breaking some of the tests, probably just need to update the test with some more `_, _` in the call to `accumulate_inference_batches`

finbarrtimbers: Ah, you're right. fixed!

saurabh111233212 (approved):

PR #1041, opened 2025-09-29
Now, `LLMRayActor` returns logprobs, and we calculate some stats about them vs the trainer logprobs in `grpo_fast.py`.

finbarrtimbers:


finbarrtimbers:
Run with truncated importance sampling: [Wandb](https://wandb.ai/ai2-llm/open_instruct_internal/runs/ak8n44w0?nw=nwuserfinbarrt)

open_instruct/grpo_fast.py
```diff
@@ -1028,6 +1071,35 @@ def train(
...
+
+                        if valid_mask.any():
+                            # Calculate logprob difference only for valid positions
+                            logprob_diff_is = mb_old_logprobs - mb_vllm_logprobs
+                            # Clamp to prevent numerical overflow in exp
+                            logprob_diff_is = torch.where(
+                                valid_mask, logprob_diff_is.clamp(-10.0, 10.0), torch.zeros_like(logprob_diff_is)
+                            )
+                            # Compute importance ratio only for valid positions
+                            tis_imp_ratio = torch.where(valid_mask, torch.exp(logprob_diff_is), tis_imp_ratio)
```
mnoukhov, line 1117: seems like this valid mask is unnecessary since `torch.exp(0) =  1` for the invalid, but fine to keep

open_instruct/grpo_fast.py
```diff
@@ -1028,6 +1071,35 @@ def train(
                     pg_losses2 = -mb_advantages[:, 1:] * torch.clamp(
                         ratio, 1.0 - args.clip_lower, 1.0 + args.clip_higher
                     )
+
+                    # Apply truncated importance sampling if enabled
+                    if args.truncated_importance_sampling_ratio_cap > 0 and mb_vllm_logprobs is not None:
+                        # Only apply importance sampling where both logprobs are valid (not INVALID_LOGPROB)
+                        # This excludes query tokens and padding which have INVALID_LOGPROB=1.0
+                        valid_mask = (mb_old_logprobs != INVALID_LOGPROB) & (mb_vllm_logprobs != INVALID_LOGPROB)
+                        valid_mask = valid_mask & mb_response_masks_bool
```
mnoukhov: are there situations where the vllm and hf logprobs will have different invalid ones?

open_instruct/grpo_fast.py
```diff
@@ -275,6 +277,8 @@ class Args:
 
     record_entropy: bool = False
     """whether to record the entropy of the policy during training. Uses extra memory."""
+    use_vllm_logprobs: bool = False
+    """whether to use vLLM's logprobs for training instead of calculating them via forward pass"""
```
mnoukhov, line 281: can we do a check that makes sure the user isn't seeing both `use_vllm_logprobs` and `truncated importance sampling` since vllm logprobs will set old_logprobs to vllm logprobs

open_instruct/grpo_fast.py
```diff
@@ -958,7 +965,7 @@ def train(
                         attention_mask = collated_attention_masks[i]
                         position_id = collated_position_ids[i]
                         response_mask = collated_response_masks[i]
-                        old_logprob, _ = self.forward(
+                        local_old_logprob, _ = self.forward(
```
mnoukhov: We should be able to skip this whole forward pass if we're doing `use_vllm_logprobs`. I don't really care about the extra compute right now tbh

finbarrtimbers:
After discussing with Hamish, one of the main issues is that we had code which adds EOS when we use a stop string (for instance, in reasoning, we end when we generate `</answer>`, not `<eos>`). Because we're manually adding EOS instead of having vLLM generate it, we obviously don't have logprobs, which is why the masks are different. 

I changed this PR to remove the EOS addition. Runs seem unaffected:

https://wandb.ai/ai2-llm/open_instruct_internal?nw=ba89hjgyfp6

open_instruct/grpo_fast.py
finbarrtimbers: Added. Good point.

open_instruct/grpo_fast.py
finbarrtimbers: Yeah, good point. Done.

open_instruct/grpo_fast.py
finbarrtimbers: So, the answer to this was yes, because we had code that added EOS to the end of the sequence whenever we ended a generation without EOS, which meant we didn't have vllm logprobs for EOS. 

I spoke with Hamish and we agreed that we can remove that code. So I think we're good now and they should always match (by masking out the query tokens).

mnoukhov: Nice!

PR #1076, opened 2025-10-10
Constant denom for loss

hamishivi:
Allows using a constant divisor for loss/masked mean stats following Dr GRPO (https://arxiv.org/abs/2503.20783)
Using this + masked_mean_axis of 1 recovers the dr grpo loss.

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -449,7 +449,9 @@ class Args:
                 "use_vllm_logprobs sets old_logprobs to vLLM logprobs, making importance sampling pointless."
             )
         if self.masked_mean_denominator is not None:
-            assert self.masked_mean_denominator > 0, "`masked_mean_denominator` must be greater than 0!"
+            assert self.masked_mean_denominator > 0, (
+                f"masked_mean_denominator (={self.masked_mean_denominator}) must be greater than 0!"
+            )
         assert self.num_samples_per_prompt_rollout > 0, "Number of samples per prompt must be greater than 0!"
         if self.num_samples_per_prompt_rollout == 1:
             logger.warning("num_samples_per_prompt_rollout is 1. This reduces GRPO to REINFORCE.")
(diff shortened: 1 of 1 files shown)


open_instruct/grpo_fast.py
```diff
@@ -446,6 +448,8 @@ def __post_init__(self):
                 "Cannot use both `use_vllm_logprobs` and `truncated_importance_sampling_ratio_cap`. "
                 "use_vllm_logprobs sets old_logprobs to vLLM logprobs, making importance sampling pointless."
             )
+        if self.masked_mean_denominator is not None:
+            assert self.masked_mean_denominator > 0, "`masked_mean_denominator` must be greater than 0!"
```
finbarrtimbers: nit: f"`masked_mean_denominator` (={self.masked_mean_denominator}) must be greater than 0!"

open_instruct/grpo_fast.py
```diff
@@ -511,12 +513,17 @@ def next_batch(dataset_indices: List[int], dataset: datasets.Dataset) -> Batch:
...
 
-def masked_mean(values: torch.Tensor, mask: torch.Tensor, axis: Optional[int] = None) -> torch.Tensor:
+def masked_mean(
+    values: torch.Tensor, mask: torch.Tensor, axis: Optional[int] = None, denominator: Optional[float] = None
+) -> torch.Tensor:
     """Compute mean of tensor with a masked values."""
     if axis is not None:
-        return ((values * mask).sum(axis=axis) / mask.sum(axis=axis)).mean()
+        numerator = (values * mask).sum(axis=axis)
+        denom = denominator if denominator is not None else mask.sum(axis=axis)
```
finbarrtimbers: I find positive conditionals more readable:

```
denom = mask.sum(axis.axis) if denominator is None else denominator
```

open_instruct/grpo_fast.py
```diff
@@ -511,12 +513,17 @@ def next_batch(dataset_indices: List[int], dataset: datasets.Dataset) -> Batch:
     )
 
 
-def masked_mean(values: torch.Tensor, mask: torch.Tensor, axis: Optional[int] = None) -> torch.Tensor:
+def masked_mean(
+    values: torch.Tensor, mask: torch.Tensor, axis: Optional[int] = None, denominator: Optional[float] = None
+) -> torch.Tensor:
     """Compute mean of tensor with a masked values."""
     if axis is not None:
```
finbarrtimbers: I think that `Tensor.sum(axis=None)` is the same as `Tensor.sum()`, right?

In which case we can remove the conditional and do:

```
numerator = (values * mask).sum(axis=axis)
denom = max.sum(axis=axis) if denominator is None else denominator
return (numerator/denom).mean()
```

(because in the case where axis is None, these are a single element, so `.mean()` does nothing, right?)

hamishivi: Good point! I tested to make sure this would the same and it is, so changed to this new implementation.

PR #1077, opened 2025-10-11
Remove duplicated `allenai/tulu-2-7b` config

emmanuel-ferdman:
## PR Summary
This minor maintenance PR removes the duplicated `allenai/tulu-2-7b` configuration, which could potentially cause issues in the future if one copy were updated independently.


finbarrtimbers:
Thanks for the PR! LGTM!

PR #1082, opened 2025-10-14
fix idx issue

saurabh111233212:
If you don't have this, you'll get a idx error during this line (2006 of grpo_fast.py):
`collated_vllm_logprobs.append(
                        collate_fn([per_device_packed_vllm_logprobs[idx] for idx in micro_range], 0)
                    )`

this only can occur when you set `allow_world_padding True`

example job of the error [here](https://beaker.allen.ai/orgs/ai2/workspaces/olmo-instruct/work/01K7G2BWQBFGGAQGQAQ5QKWD8S?taskId=01K7G2BWQE1B54R18YP95T2PP7&jobId=01K7G2BWTXREWTV0K2F8DSVV1B)
after the fix: job is [here](https://beaker.allen.ai/orgs/ai2/workspaces/olmo-instruct/work/01K7J65E59QEH5G6RG694D8H24?taskId=01K7J65E5CP6TQH4VAV8S4M6RA&jobId=01K7J65E8Z16GY61JHRD5E6GRZ)

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -1051,11 +1051,18 @@ class PolicyTrainerRayProcess(RayProcess):
                     # Compare vLLM logprobs with local logprobs
                     with torch.no_grad():
                         valid_mask = mb_response_masks_bool & ~torch.isnan(mb_vllm_logprobs)
+                        valid_count = valid_mask.sum()
+                        valid_count_value = int(valid_count.item())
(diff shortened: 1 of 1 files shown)


finbarrtimbers:
@gemini-code-assist is there a way we can make this code unit testable that would let us reproduce the bug in a unit test?

finbarrtimbers (approved):

PR #1085, opened 2025-10-16
Fixing tools on main

hamishivi:
Tool use was broken on main due to vllm logprobs not being padded appropriately.

In most cases, we mask out tool outputs, so padding out the logprobs is fine. It's an interesting research question to see if we can train on the outputs directly ;)

Also fixes a hang I found: if the process_queue thread dies while the requests are draining, you end up hanging because the draining requests function has no health checks + infinite-ish loop.

Successful tool call run: https://beaker.allen.ai/orgs/ai2/workspaces/open-instruct-dev/work/01K7Q14QQ6QW4XF6BKP8CAJ0BY?taskId=01K7Q14QQC26C0C5RYN141FP9C&jobId=01K7Q14QXFPY3Y1TQBJMJT16N1

Successful 1-gpu run: https://beaker.allen.ai/orgs/ai2/workspaces/open-instruct-dev/work/01K7Q1ZBY1KM0ZTKD0T31NVN7Y?taskId=01K7Q1ZBY50MMRVVV2H16P6TFV&jobId=01K7Q1ZC1ZBJAMG7B51KV83V0C

successful looking multi node run: https://beaker.allen.ai/orgs/ai2/workspaces/open-instruct-dev/work/01K7Q2SZQ4TA1B7YHG8RR47GFY?taskId=01K7Q2SZQ89V5R4Y4BNHX0X3NG&jobId=01K7Q2SZTSP9WVCMT61N5Z084P

### DIFF
--- a/open_instruct/vllm_utils3.py
+++ b/open_instruct/vllm_utils3.py
@@ -176,21 +176,9 @@ def _handle_output(output, tools, tracking, sampling_params, max_tool_calls, exe
 
     if output.request_id in tracking["concat_outputs"]:
         stored = tracking["concat_outputs"][output.request_id].outputs[0]
-        assert getattr(stored, "logprobs", None) is not None, f"{stored.logprobs=}"
-        new_logprobs = getattr(o, "logprobs", None)
-        # e.g. tools may not return logprobs, so we provide placeholders.
-        # these will be masked out during training.
-        if new_logprobs is None:
-            new_logprobs = [{tid: types.SimpleNamespace(logprob=0.0)} for tid in o.token_ids]
-
         stored.token_ids.extend(o.token_ids)
-        stored.logprobs.extend(new_logprobs)
+        stored.logprobs.extend(o.logprobs)
     else:
-        # First time seeing this request, store it and ensure logprobs are present.
-        # in reality, the first time seeing should involve model generation and always appear,
-        # but we'll handle there being some initial tool call here just in case.
-        if getattr(o, "logprobs", None) is None:
-            o.logprobs = [{tid: types.SimpleNamespace(logprob=0.0)} for tid in o.token_ids]
         tracking["concat_outputs"][output.request_id] = output
 
     tracking["masks"][output.request_id].extend([1] * len(o.token_ids))


open_instruct/vllm_utils3.py
```diff
@@ -173,10 +174,23 @@ def _handle_output(output, tools, tracking, sampling_params, max_tool_calls, exe
     assert len(output.outputs) <= 1, f"{len(output.outputs)=}"  # In tool mode, sampling_params.n == 1
     o = output.outputs[0]
 
-    # Update concatenated outputs
     if output.request_id in tracking["concat_outputs"]:
-        tracking["concat_outputs"][output.request_id].outputs[0].token_ids.extend(o.token_ids)
+        stored = tracking["concat_outputs"][output.request_id].outputs[0]
+        assert getattr(stored, "logprobs", None) is not None, f"{stored.logprobs=}"
```
finbarrtimbers: I really don't like assert getattr. Will this ever be None? I think we should always have logprobs, right?

open_instruct/grpo_fast.py
```diff
@@ -503,6 +503,10 @@ def __post_init__(self):
                 if tool not in ["search", "code"]:
                     raise ValueError(f"Tool {tool} is not supported. Supported tools are: search, code")
             assert len(self.tools) == len(set(self.tools)), "Duplicate tools are not allowed"
+            if self.use_vllm_logprobs or self.truncated_importance_sampling_ratio_cap > 0.0:
```
finbarrtimbers, line 506: Can we set mask_tool_use programmatically? Does it need to be a configurable setting?

finbarrtimbers (approved):

open_instruct/vllm_utils3.py
hamishivi: This is just being overly cautious haha - happy to change.

open_instruct/grpo_fast.py
hamishivi: Way back people experimented with not masking the outputs, but it usually causes divergence. I sorta wanna experiment with ways to learn from tool ouputs during RL training, but that is future research work. For now the default is to mask so this should only fire when someone is explicitly turning it off...

open_instruct/vllm_utils3.py
hamishivi: Simplified this logic!

PR #1092, opened 2025-10-17
Compiles models used for training

finbarrtimbers:
Performance is marginally better in `large_test_script.sh` ([Wandb](https://wandb.ai/ai2-llm/open_instruct_internal?nw=bg1bj8j5xiv)):

Screenshot 2025-10-20 at 10 12 17 AM

(orange line is with `torch.compile` using default settings, green line is without)

We also see extremely slow performance when we add 

`compile_kwargs={"mode": "max-autotune"}`

Screenshot 2025-10-20 at 10 13 49 AM

I suspect this is because our batch size keeps changing due to filtering. Using @mnoukhov's active refilling should help here. 

I also suspect that we'll see better performance if we use fsdp2, like #1095 suggests.

Runs: 

1. Single GPU: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/open-instruct-dev/work/01K7T6E00C5GQRC3J7KCCMXDWT?)
2. Multi-node: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/open-instruct-dev/work/01K7T7C76RCT4KPGX4BWZTBAQ8?)

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -355,6 +355,8 @@ class Args:
     """vLLM top p for nucleus sampling"""
     deepspeed_stage: int = 0
     """the deepspeed stage"""
+    compile_mode: str = "default"
+    """torch.compile mode: 'default', 'reduce-overhead', 'max-autotune', or 'max-autotune-no-cudagraphs'"""
     gather_whole_model: bool = True
     """whether to gather the whole model to boardcast (not doable for 70B but can be faster for 8B)"""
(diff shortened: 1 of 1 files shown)


hamishivi (approved):
If it doesn't break anything happy to merge, but if it doesn't change much else then not sure if worth including... up to you!

finbarrtimbers:
We don't actually see a performance improvement, so I'm putting this on hold.

PR #1093, opened 2025-10-19
Remove old rl-rag code

hamishivi:
Removing old code from the RL-rag project. This is not needed on main and no longer even used.

Part of me breaking up the big tool refactor PR (#1086)


finbarrtimbers (approved):
BAM! Might be a new record for `open-instruct` deletion!

PR #1097, opened 2025-10-20
Allow manual system prompt + streamline tokenize

hamishivi:
Another part of breaking up my big tool refactor PR. This one adds the option to manually provide a system prompt (in the form of a text file), which will then add the given system prompt to all training examples, or override the system prompt in the data (if present).

I also streamlined the tokenize function a little to remove cruft that wasn't needed.
Also removes the max_token_length parameter, which was weird (it filtered out examples that were too long when including the ground truth).

### DIFF
--- a/scripts/train/debug/local_tool_grpo_fast.sh
+++ b/scripts/train/debug/local_tool_grpo_fast.sh
@@ -4,7 +4,6 @@ uv run open_instruct/grpo_fast.py \
     --dataset_mixer_list_splits train \
     --dataset_mixer_eval_list hamishivi/tulu_3_rewritten_100k_with_tool_prompt 16 \
     --dataset_mixer_eval_list_splits train \
-    --max_token_length 512 \
     --max_prompt_token_length 512 \
     --response_length 512 \
     --pack_length 1024 \


open_instruct/dataset_transformation.py
```diff
@@ -1236,6 +1236,47 @@ def rlvr_tokenize_v2(
...
+    # concatenate all the previous messages as <role>: <content>\n <role>: <content>\n ...
+    row[RAW_PROMPT_KEY] = "\n".join(f"{msg['role']}: {msg['content']}" for msg in prompt)
+    # some basic transformations:
+    # if ground truths is a string, make it a list
+    if isinstance(row[ground_truths_key], str):
+        row[ground_truths_key] = [row[ground_truths_key]]
+    # if dataset source is a string, make it a list
+    if isinstance(row[verifier_source_key], str):
+        row[verifier_source_key] = [row[verifier_source_key]]
+    return row
```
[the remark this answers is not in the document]
hamishivi: This comment is fine, since we don't require that every tokenize func works with every filter func.

open_instruct/dataset_transformation.py
```diff
@@ -1236,6 +1236,47 @@ def rlvr_tokenize_v2(
...
+    ground_truths_key: str = GROUND_TRUTHS_KEY,
+    verifier_source_key: str = VERIFIER_SOURCE_KEY,
+    system_prompt_override: Optional[str] = None,
+):
+    prompt = row.pop(sft_messages_key)
+    assert len(prompt) > 0, "Empty prompt in dataset"
+    # if the prompt has multiple messages, make sure we don't end in an assistant message.
+    if len(prompt) > 1:
+        if prompt[-1]["role"] == "assistant":
+            prompt = prompt[:-1]
```
finbarrtimbers: Let's combine these two conditions?

open_instruct/dataset_transformation.py
```diff
@@ -1236,6 +1236,47 @@ def rlvr_tokenize_v2(
...
+    # override the system prompt if we have a new one provided.
+    if system_prompt_override:
+        # if the first message is a system message, override it.
```
finbarrtimbers: What if you do: 

```
if prompt[0]["role"] == "system":
    del prompt[0]
prompt = [{"role": "system", "content": system_prompt_override}] + prompt
```

I think that's equivalent and we remove an else?

open_instruct/dataset_transformation.py
```diff
@@ -1255,6 +1296,12 @@ def rlvr_filter_v1(
     return max_prompt_token_length_ok and max_token_length_ok and (contain_some_labels or not need_contain_labels)
 
 
+def rlvr_filter_v2(row: Dict[str, Any], tokenizer: PreTrainedTokenizer, max_prompt_token_length: Optional[int] = None):
```
finbarrtimbers: Can we call this "max_length_filter"?

open_instruct/dataset_transformation.py
```diff
@@ -1265,8 +1312,8 @@ def rlvr_filter_v1(
     "preference_filter_v1": (preference_filter_v1, "filter"),
     "preference_tulu_tokenize_and_truncate_v1": (preference_tulu_tokenize_and_truncate_v1_2, "map"),
     "preference_tulu_filter_v1": (preference_tulu_filter_v1, "filter"),
-    "rlvr_tokenize_v1": (rlvr_tokenize_v2, "map"),
-    "rlvr_filter_v1": (rlvr_filter_v1, "filter"),
+    "rlvr_tokenize_v1": (rlvr_tokenize_v3, "map"),
+    "rlvr_filter_v1": (rlvr_filter_v2, "filter"),
```
finbarrtimbers: Can we also call this max_length_filter?

open_instruct/dataset_transformation.py
hamishivi: I think the tricky thing is that SFT/DPO use a different filter, since they take into account full sample length (max_token_length) instead of just prompt (max_prompt_token_length)?

open_instruct/dataset_transformation.py
hamishivi: Good point, done!

open_instruct/dataset_transformation.py
```diff
@@ -1265,8 +1311,8 @@ def rlvr_filter_v1(
     "preference_filter_v1": (preference_filter_v1, "filter"),
     "preference_tulu_tokenize_and_truncate_v1": (preference_tulu_tokenize_and_truncate_v1_2, "map"),
     "preference_tulu_filter_v1": (preference_tulu_filter_v1, "filter"),
-    "rlvr_tokenize_v1": (rlvr_tokenize_v2, "map"),
-    "rlvr_filter_v1": (rlvr_filter_v1, "filter"),
+    "rlvr_tokenize_v1": (rlvr_tokenize_v3, "map"),
```
finbarrtimbers, line 1314: Why are these different versions? (why not both v1 or v3?)

finbarrtimbers (approved):

open_instruct/dataset_transformation.py
hamishivi: This is an old design Costa felt strongly about --that maybe you could have multiple tokenize and filter functions and give these diff names, and so swap prepro easily at script calling time... but then the default options in the script use v1, so we ended up just updating this mapping so that people could get new prepro without updating scripts.

hamishivi: The _v3 name for the function itself helps avoid the HF cache invalidation issue: now its a new function with a new name, so it won't try to reuse the old caches.

PR #1111, opened 2025-10-27
active sampling or whatever we call it

mnoukhov:
> [!NOTE]
> Introduce active sampling to refill batches after filtering, refactor data preparation into aggregated chunks with safer iterators and enriched metrics, and adjust save/eval flow.
> 
> - **GRPO pipeline**:
>   - Add `args.active_sampling` and `args.active_sampling_max_attempts`; remove `args.fill_completions`.
>   - Refactor `data_preparation_thread` to iteratively accumulate, score, filter (zero-std), optionally refill with new prompts, aggregate chunks, and pack sequences.
>   - Mask truncated completions by zeroing scores/advantages and response masks when enabled.
>   - Compute and log new metrics: per-batch filtered group counts, zero/solved reward group ratios, packed ratio, and guarded tokens/sec; propagate `prompt_lengths`/`response_lengths`.
> - **Batch/inference handling**:
>   - `accumulate_inference_batches` now returns `(GenerationResult, Batch, prompt_lengths, response_lengths)` and preserves `indices` field.
>   - Add `combine_reward_metrics`, `FilteredChunk`, and `AggregationState` to merge multi-iteration results and stats.
> - **Iterator robustness**:
>   - `ShufflingIterator` now guards empty data, centralizes effective-size computation, and updates it on state restore.
> - **Training loop**:
>   - Track `val/total_filtered_groups` to submit extra prompt batches when filtering shrinks data; avoid step collisions via fractional `training_step` offsets.
>   - Save condition updated to allow save at step 1 when `eval_on_step_0` is true.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f8181e51ac1c6f83396e7171066de216fb7ccee3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -279,9 +279,9 @@ class Args:
     use_vllm_logprobs: bool = False
     """whether to use vLLM's logprobs for training instead of calculating them via forward pass"""
 
-    active_fill_completions: bool = False
+    active_sampling: bool = False
     """Whether to refill the batch with *new prompts/completions* after filtering."""
-    active_fill_max_attempts: int = 3
+    active_sampling_max_attempts: int = 3
     """How many times to attempt to fill"""
 
     # Reward
@@ -1889,7 +1889,7 @@ def data_preparation_thread(
 
             prompts_kept_this_iter = len(scores) // per_prompt_rollout if per_prompt_rollout > 0 else 0
 
-            if not args.active_fill_completions:
+            if not args.active_sampling:
                 break
 
             collected_samples = sum(len(chunk.scores) for chunk in aggregation.chunks)
@@ -1898,11 +1898,11 @@ def data_preparation_thread(
             if prompts_remaining <= 0:
                 break
 
-            if prompts_kept_this_iter == 0 and fill_iteration > args.active_fill_max_attempts:
+            if prompts_kept_this_iter == 0 and fill_iteration > args.active_sampling_max_attempts:
                 total_responses = sum(len(chunk.kept_indices) for chunk in aggregation.chunks)
                 logger.warning(
                     "[Active fill completions] Unable to collect non-zero advantage prompts in iteration %d; "
-                    "set as max by args.active_fill_max_attempts, proceeding with existing batch of size %d.",
+                    "set as max by args.active_sampling_max_attempts, proceeding with existing batch of size %d.",
                     fill_iteration,
                     total_responses,
                 )


open_instruct/grpo_fast.py
```diff
@@ -574,16 +578,21 @@ def __init__(self, data: np.ndarray, batch_size: int, seed: Optional[int] = None
         self.epoch_number = 0
         self.rng = np.random.default_rng(seed)
         self.rng.shuffle(self.data)
+        self.exclude_list = []
 
         # Ensure the effective dataset size is divisible by batch_size
-        self.effective_size = len(self.data) - (len(self.data) % batch_size)
+        self._update_effective_size()
```
finbarrtimbers, line 581: Why don't we just remove this and always have batches of size `batch_size`?

Either by throwing away the extra data or refilling across epoch numbers.

open_instruct/grpo_fast.py
```diff
@@ -593,6 +602,10 @@ def __next__(self) -> List[int]:
 
         return batch
 
+    def exclude_indices(self, exclude_list: List) -> None:
```
finbarrtimbers: Why would we use this?

open_instruct/grpo_fast.py
```diff
@@ -1709,181 +1736,346 @@ def data_preparation_thread(
...
-                result.responses[i].append(tokenizer.eos_token_id)
-                result.masks[i].append(1)
-                result.logprobs[i].append(float("nan"))
-        with Timer("🔥 [Data Preparation Thread] Decoding responses", noop=True):
-            decoded_responses = tokenizer.batch_decode(result.responses, skip_special_tokens=True)
-            decoded_queries = batch.raw_queries
-            stop_rate = sum(int(finish_reason == "stop") for finish_reason in result.finish_reasons) / len(
-                result.finish_reasons
-            )
+    def combine_reward_metrics(metric_records: list[tuple[Dict[str, Any], int]]) -> Dict[str, Any]:
```
finbarrtimbers: Can you add a docstring?

open_instruct/grpo_fast.py
```diff
@@ -1709,181 +1736,346 @@ def data_preparation_thread(
...
+        if not metric_records:
```
finbarrtimbers: I don't think we need this, everything else should be empty and not iterate if metric_records is empty

open_instruct/grpo_fast.py
```diff
@@ -1709,181 +1736,346 @@ def data_preparation_thread(
...
+            return {}
+
+        buckets: Dict[str, list[tuple[Any, int]]] = {}
+        for metrics, weight in metric_records:
+            if not metrics:
```
finbarrtimbers: Similarly, this is not needed as the for loop will not execute if metrics is empty

open_instruct/grpo_fast.py
```diff
@@ -1709,181 +1736,346 @@ def data_preparation_thread(
...
+                continue
+            for key, value in metrics.items():
+                buckets.setdefault(key, []).append((value, weight))
```
finbarrtimbers: Let's make `buckets` a `collections.defaultdict(list)` so you don't need to do this?

open_instruct/grpo_fast.py
```diff
@@ -1709,181 +1736,346 @@ def data_preparation_thread(
...
+
+        combined: Dict[str, Any] = {}
+        for key, records in buckets.items():
+            sample_value = records[0][0]
+            if isinstance(sample_value, np.ndarray):
+                combined[key] = np.concatenate([np.asarray(value) for value, _ in records])
+            elif isinstance(sample_value, (list, tuple)):
+                concatenated: list[Any] = []
```
finbarrtimbers: what about: 

```
combined[key] = [x for value, _ in records for x in value]
```

open_instruct/grpo_fast.py
```diff
@@ -1709,181 +1736,346 @@ def data_preparation_thread(
...
+                for value, _ in records:
+                    concatenated.extend(list(value))
+                combined[key] = concatenated
+            elif isinstance(sample_value, (int, float, bool, np.integer, np.floating)):
+                total_weight = sum(weight for _, weight in records)
+                if total_weight == 0:
+                    combined[key] = float(sample_value)
+                else:
+                    combined[key] = sum(float(value) * weight for value, weight in records) / total_weight
```
finbarrtimbers: what about: 

```
weighted_sum = 0
total_weight = 0
for value, weight in records:
    weighted_sum += float(value) * weight
    total_weight += weight
combined[key] = weighted_sum / total_weight if total_weight < 1e-6 else sample_value
```

this way we 1) don't compare a sum of floats directly to 0 and 2) we avoid iterating over records twice.

open_instruct/grpo_fast.py
```diff
@@ -411,6 +413,8 @@ class Args:
     """the max generation length for evaluation for oe-eval"""
     oe_eval_beaker_image: Optional[str] = None
     """the docker image for evaluation for oe-eval"""
+    oe_eval_gpu_multiplier: Optional[int] = 1
+    """gpu mulitplier for eval jobs"""
```
finbarrtimbers: multiplier

open_instruct/grpo_fast.py
```diff
@@ -411,6 +413,8 @@ class Args:
     """the max generation length for evaluation for oe-eval"""
     oe_eval_beaker_image: Optional[str] = None
     """the docker image for evaluation for oe-eval"""
+    oe_eval_gpu_multiplier: Optional[int] = 1
```
finbarrtimbers: What does this do? Why would I want to multiply GPUs for eval jobs?

open_instruct/grpo_fast.py
```diff
@@ -274,14 +274,16 @@ class Args:
...
-    """Whether to refill the batchsize with after filtering."""
-
     record_entropy: bool = False
     """whether to record the entropy of the policy during training. Uses extra memory."""
     use_vllm_logprobs: bool = False
     """whether to use vLLM's logprobs for training instead of calculating them via forward pass"""
 
+    active_sampling: bool = False
+    """Whether to refill the batch with *new prompts/completions* after filtering."""
+    active_sampling_max_attempts: int = 3
```
finbarrtimbers: Do we actually want a max number of attempts? Let's have it wait indefinitely.

open_instruct/grpo_fast.py
mnoukhov: my bad, this is holdovers from my next PR which will remove resampling of batches with 90% solve rate

open_instruct/grpo_fast.py
mnoukhov: also holdover from another PR

PR #1102, opened 2025-10-21
Add `VLLM_ALLOW_INSECURE_SERIALIZATION=1` in `grpo_fast.py`

gau-nernst:
In the latest main, I encounter this error

```
2025-10-21 00:04:23 - INFO - grpo_fast.py:2272 - ======== ✅ all models and vLLM engines initialized =========
(LLMRayActor pid=1516796) (EngineCore_DP0 pid=1517585) Exception in thread Thread-5 (process_output_sockets):
(LLMRayActor pid=1516796) (EngineCore_DP0 pid=1517585) Traceback (most recent call last):
(LLMRayActor pid=1516796) (EngineCore_DP0 pid=1517585)   File "/home/ubuntu/.local/share/uv/python/cpython-3.12.11-linux-x86_64-gnu/lib/python3.12/threading.py", line 1075, in _bootstrap_inner
(LLMRayActor pid=1516796) (EngineCore_DP0 pid=1517585)     self.run()
(LLMRayActor pid=1516796) (EngineCore_DP0 pid=1517585)   File "/home/ubuntu/.local/share/uv/python/cpython-3.12.11-linux-x86_64-gnu/lib/python3.12/threading.py", line 1012, in run
(LLMRayActor pid=1516796) (EngineCore_DP0 pid=1517585)     self._target(*self._args, **self._kwargs)
(LLMRayActor pid=1516796) (EngineCore_DP0 pid=1517585)   File "/home/ubuntu/thien/venv/open-instruct/lib/python3.12/site-packages/vllm/v1/engine/core.py", line 932, in process_output_sockets
(LLMRayActor pid=1516796) (EngineCore_DP0 pid=1517585)     buffers = encoder.encode_into(outputs, buffer)
(LLMRayActor pid=1516796) (EngineCore_DP0 pid=1517585)               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(LLMRayActor pid=1516796) (EngineCore_DP0 pid=1517585)   File "/home/ubuntu/thien/venv/open-instruct/lib/python3.12/site-packages/vllm/v1/serial_utils.py", line 136, in encode_into
(LLMRayActor pid=1516796) (EngineCore_DP0 pid=1517585)     self.encoder.encode_into(obj, buf)
[...]

PR #1114, opened 2025-10-28
Sets default env vars, copying mason.py

finbarrtimbers:
There has to be a better way to keep these in sync...

See also #1102, which ran into this.

---

> [!NOTE]
> Adds VLLM environment defaults in `scripts/train/debug/grpo_fast.sh` (enable V1, disable compile cache, allow insecure serialization).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d84783d5689702375451726931337bf3a85541b8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/scripts/train/debug/grpo_fast.sh
+++ b/scripts/train/debug/grpo_fast.sh
@@ -1,3 +1,6 @@
+export VLLM_ALLOW_INSECURE_SERIALIZATION=1
+export VLLM_DISABLE_COMPILE_CACHE=1
+export VLLM_USE_V1=1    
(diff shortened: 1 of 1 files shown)


mnoukhov:
maybe we should create a file called `env_vars` that sets all of these and source it for respective scripts

finbarrtimbers:
> maybe we should create a file called `env_vars` that sets all of these and source it for respective scripts

That's a great idea.

PR #1159, opened 2025-11-10
Fix comment clarity in `maybe_evaluate` function in `grpo_fast.py`

MohdElgaar:
The comment describing the timeout logic in `maybe_evaluate` was confusing because it didn't clearly state when the short timeout (0.01) is used. It could be read as saying we use the short timeout when it IS the last step, and otherwise should use a long timeout.

Fixed by adding "not" to clarify: we use the short timeout when this is NOT the last training step or we're not evaluating. The long timeout (100) is only used at the last step. 

---

> [!NOTE]
> Clarifies the timeout comment in `maybe_evaluate` to state the short timeout is used when not on the final step or when eval is disabled.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit de781b219a1bf3dd2585ce9255dace3fd4602712. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/open_instruct/benchmark_generators.py
+++ b/open_instruct/benchmark_generators.py
@@ -29,7 +29,7 @@ from ray.util import queue as ray_queue
 
 from open_instruct import dataset_transformation, grpo_fast, logger_utils, model_utils, utils, vllm_utils
 from open_instruct.actor_manager import ActorManager
-from open_instruct.queue_types import PromptRequest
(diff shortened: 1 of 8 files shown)


finbarrtimbers:
Closing as I have made the fix in a separate PR (I can't figure out how to merge this). Thank you!

PR #1183, opened 2025-11-12
pass in `model_name_or_path` that is only augusta and it works

mnoukhov:
copies down the tokenizer, tokenizer_config, and config to weka in order to do dataset preprocessing

---

> [!NOTE]
> Enable dataset preprocessing with gs:// models by auto-downloading tokenizer/config from GCS and add multi-file GCS download utility with tests.
> 
> - **Open Instruct integration**:
>   - Add `maybe_download_tokenizer_from_gs_bucket()` in `mason.py` to detect `--model_name_or_path gs://...` and download `tokenizer.json`, `tokenizer_config.json`, and `config.json` to a local cache, rewriting the arg for dataset caching.
>   - Invoke the helper before running caching commands for `open_instruct/*` scripts.
> - **Utilities**:
>   - Update `open_instruct/utils.py` `download_from_gs_bucket()` to accept multiple `src_paths` and ensure destination directory exists; construct `gsutil cp -r` with all sources.
> - **Tests**:
>   - Add `TestDownloadFromGsBucket` in `open_instruct/test_utils.py` to verify constructed `gsutil` command and destination creation.
> - **Misc**:
>   - Minor comment tweak in GCP model handling (short commit hash note).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ea22a7cb3197d398cb06e020036f44c5dbcf5701. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


mason.py
```diff
@@ -467,6 +472,31 @@ def make_internal_command(command: list[str], args: argparse.Namespace, whoami:
...
+                                    f"{model_name_or_path}/tokenizer.json",
+                                    f"{model_name_or_path}/tokenizer_config.json",
+                                    f"{model_name_or_path}/config.json",
+                                ],
+                                local_cache_folder,
+                            )
+
+                        filtered_command[model_name_idx] = local_cache_folder
+                except ValueError:
+                    pass
```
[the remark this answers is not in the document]
mnoukhov: if `model_name_or_path` is the last argument then we would want to crash as that's a weird input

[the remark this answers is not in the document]
mnoukhov: same as gemini above

mason.py
```diff
@@ -467,6 +472,31 @@ def make_internal_command(command: list[str], args: argparse.Namespace, whoami:
...
+
+                        if not os.path.exists(local_cache_folder):
+                            download_from_gs_bucket(
+                                [
+                                    f"{model_name_or_path}/tokenizer.json",
+                                    f"{model_name_or_path}/tokenizer_config.json",
+                                    f"{model_name_or_path}/config.json",
+                                ],
+                                local_cache_folder,
+                            )
```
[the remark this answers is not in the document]
mnoukhov: ehh, I think mine is fine as is

open_instruct/utils.py
```diff
@@ -1074,9 +1075,11 @@ def download_from_gs_bucket(src_path: str, dest_path: str) -> None:
         "-m",
         "cp",
         "-r",
-        src_path,
-        dest_path,
     ]
+    if not isinstance(src_paths, list):
```
finbarrtimbers: Let's change this to only take a list and force the callers to pass one in?

open_instruct/utils.py
```diff
@@ -1074,9 +1075,11 @@ def download_from_gs_bucket(src_path: str, dest_path: str) -> None:
...
+        src_paths = [src_paths]
+    cmd.extend(src_paths)
```
finbarrtimbers, line 1162: Can you add some tests here? Let's mock `live_subprocess_output` so we capture the cmd it's called with and verify it against some known correct values. Should be a one prompt change with Codex

mason.py
finbarrtimbers: Can you rewrite this to make these changes: 

1. add a comment saying when we get a ValueError
2. Can you change this to be a function, and return early when a condition is not true? then we can have less nesting, which will make it easier to follow.

finbarrtimbers (changes requested):

open_instruct/utils.py
mnoukhov: done

open_instruct/utils.py
mnoukhov: multiple prompts but I think its good

mason.py
mnoukhov: 1. we actually only get it if there's no `--model_name_or_path` in the command, so I made that an explicit `if` statement

2. made it a function which is cleaner

mason.py
```diff
@@ -464,6 +464,9 @@ def make_internal_command(command: list[str], args: argparse.Namespace, whoami:
                     continue
 
                 filtered_command = build_command_without_args(command[idx:], CACHE_EXCLUDED_ARGS)
+                filtered_command = maybe_download_tokenizer_from_gs_bucket(
+                    filtered_command, args.auto_output_dir, whoami
```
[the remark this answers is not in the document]
mnoukhov: fixed

mason.py
```diff
@@ -784,6 +787,35 @@ def make_task_spec(args, full_command: str, i: int, beaker_secrets: list[str], w
...
+    """if model is only on gs, download tokenizer from gs to local cache folder for dataset preprocessing"""
+
+    if "--model_name_or_path" not in filtered_command:
+        return filtered_command
+
+    model_arg_idx = filtered_command.index("--model_name_or_path")
+    model_name_idx = model_arg_idx + 1
+    model_name_or_path = filtered_command[model_name_idx].rstrip("/")
+
+    if model_name_or_path.startswith("gs://"):
```
finbarrtimbers: what if you do: 

```
if not model_name_or_path.starswith("gs://"):
    return filtered_command
...
```

then you can remove one level of nesting?

mason.py
```diff
@@ -784,6 +787,35 @@ def make_task_spec(args, full_command: str, i: int, beaker_secrets: list[str], w
...
+        model_name_hash = hashlib.md5(model_name_or_path.encode("utf-8")).hexdigest()[:8]
```
finbarrtimbers: Why 8? Please add a comment!

finbarrtimbers (approved):

mason.py
mnoukhov: good idea, done!

mason.py
mnoukhov: the short commit hash, adding comment!

PR #1174, opened 2025-11-11
replenish prompts directly after accumulating

mnoukhov:
successful [single GPU on beaker](https://beaker.allen.ai/orgs/ai2/workspaces/open-instruct-dev/work/01KA21TQT0YGQFJ5EPM8CF0JX9?taskId=01KA21TQT642Q8D9AWQZBPZCJW&jobId=01KA21TQY6SYPK15FAJ0QXDJ1C) 

moves prompt loading into `accumulate_inference_batches`:
- when `replenish_prompts=True` we add a prompt on the prompt_Q after consuming a finished result

simplifies a lot of our batch and dataset stuff
- now pulls single examples instead of `batch_size` 
- adds each prompt, one by one to the generator's queue
- removes extra code and functions

---

> [!NOTE]
> Moves to single-example prompt flow with immediate replenishment during accumulation, removes batch-splitting helpers, and updates training/eval and tests accordingly.
> 
> - **Core sampling/accumulation**:
>   - Add `replenish_prompts` option to `accumulate_inference_batches`; when enabled, pulls next index from `ShufflingIterator` and enqueues a new prompt via `add_prompt_to_generator` after each completed result.
>   - Introduce `add_prompt_to_generator(example, idx, ...)` to insert prompts and track them in `PendingQueriesMap`.
>   - Remove `next_batch` and `split_and_insert_batch`; flow now operates on single examples.
>   - `ShufflingIterator.__next__` returns a single index when `batch_size==1`.
> - **Training loop**:
>   - Seed initial work by enqueuing `async_steps * num_unique_prompts_rollout` single prompts.
>   - Data prep thread now passes `replenish_prompts=True` and receives/train-packs batches continuously.
>   - Replace `eval_batch` with iterating over `eval_dataset`; enqueue each eval example individually.
>   - Build iterator with `batch_size=1`; simplify refill logic and remove manual prompt refill counters.
> - **API/args tweaks**:
>   - Extend `accumulate_inference_batches` signature (dataset, queues, iterator, training_step); assert presence when `replenish_prompts` or `no_resampling_pass_rate` used.
>   - Minor variable renames (`dataset` -> `dataset_name`), add `Dataset` import.
> - **Tests**:
>   - Replace batch-splitting usages with `setup_and_add_prompts_to_generator` and direct `add_prompt_to_generator` calls.
>   - Remove `model_utils` dependency in tests; update expectations for per-prompt requests and multiple-sample reference counting.
> 
[...]

PR #1193, opened 2025-11-14
There were bugs

finbarrtimbers:
Fixes two bugs from active resampling. 

Bug 1:

When all elements in the batch were filtered out and active sampling was not enabled, we'd get this bug:

```
2025-11-14T18:34:12.534Z   File "/stage/open_instruct/grpo_fast.py", line 1869, in accumulate_inference_batches
2025-11-14T18:34:12.535Z     epoch_number=results[0].epoch_number,
2025-11-14T18:34:12.535Z                  ~~~~~~~^^^
2025-11-14T18:34:12.535Z IndexError: list index out of range
```

Fixed this by returning early from `accumulate_inference_batches` so we go down the old `if B == 0` code path.

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -2990,8 +2990,12 @@ def run_training(
             enable=False,
         )
 
-    # Send initial data to ensure we have a N-step offset.
-    for _ in range(args.async_steps):
+    total_training_steps_remaining = max(0, args.num_training_steps - resume_training_step + 1)
+    prefill_steps = min(args.async_steps, total_training_steps_remaining)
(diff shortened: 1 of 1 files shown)


mnoukhov:
One thing to watch out for is that `generation_batches` != `training_batches`. 

Currently, we're assuming they are the same but because we're filtering, that isn't true. We should either
1. refill generation continuously until we're done then just kill the training_queue 
or
2.  do everything in terms of generation batch and keep track of those

Feels like 1 is the simpler solution

finbarrtimbers:
Closing as this was fixed by #1174.

PR #1198, opened 2025-11-16
add filestore fix for multi-learn node augsta checkpointing

saurabh111233212:
> [!NOTE]
> Ensure `checkpoint_state_dir` is rewritten under `/filestore` when `gs_bucket_path` is set and the path isn’t already in filestore.
> 
> - **Checkpointing**:
>   - When `gs_bucket_path` is provided (and `gs_checkpoint_state_dir` unset), rewrite `checkpoint_state_dir` to `/filestore/...` if it doesn't already start with `/filestore`, before GCS download/calibration.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7942a4a67980eeeba494c4872102436988858879. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -420,6 +420,8 @@ class Args:
     """multiply the gpus used for each oe-eval task"""
     eval_priority: Literal["low", "normal", "high", "urgent"] = "normal"
     """the priority of auto-launched evaluation jobs"""
+    eval_workspace: str = "ai2/tulu-3-results"
+    """the workspace to launch evaluation jobs on"""
     send_slack_alerts: bool = False
     """Whether to send Slack alerts on training failures"""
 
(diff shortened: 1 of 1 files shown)


open_instruct/grpo_fast.py
```diff
@@ -501,6 +501,8 @@ def __post_init__(self):
                 self.gs_checkpoint_state_dir = f"{self.gs_bucket_path}/{beaker_users}/{checkpoint_dir_name}"
             else:
                 self.gs_checkpoint_state_dir = f"{self.gs_bucket_path}/{checkpoint_dir_name}"
+            if not checkpoint_dir_name.startswith("/filestore"):
+                self.checkpoint_state_dir = f"/filestore{self.checkpoint_state_dir}"
```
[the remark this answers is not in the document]
saurabh111233212: sure this is reasonable, commiting

open_instruct/grpo_fast.py
```diff
@@ -501,6 +501,8 @@ def __post_init__(self):
                 self.gs_checkpoint_state_dir = f"{self.gs_bucket_path}/{beaker_users}/{checkpoint_dir_name}"
             else:
                 self.gs_checkpoint_state_dir = f"{self.gs_bucket_path}/{checkpoint_dir_name}"
+            if not checkpoint_dir_name.startswith("/filestore"):
```
finbarrtimbers, line 506: Wdyt about making this logic live in mason?

finbarrtimbers (changes requested):

open_instruct/grpo_fast.py
saurabh111233212: Ooh, that's a good idea. 

I'll make the change

saurabh111233212: @finbarrtimbers hmm, actually, this is the logic for auto-setting checkpoint state dir. 

should we move the whole block to mason? Basically, can't move just the filestore bit because the checkpoint state dir might not exist before grpo_fast is actually called

but we could move the entire logic of auto-setting checkpoint_state_dir to mason (the entire if statement in which the `if not checkpoint_dir_name.startswith("/filestore")` check lives under)

PR #1244, opened 2025-12-02
Adds an end to end integration test verifying that our generations match previous ones.

finbarrtimbers:
In this PR, I add a new file, `test_grpo_fast_gpu.py`, and move all the tests from `test_grpo_fast.py` which require a GPU into it. Then, I add a new test which runs generation and checks the results against a reference generation. The goal here is to make it very easy for us to look at generations and 1) verify they haven't changed and 2) verify the format is reasonable.  

I have done my best to make this as reproducible as possible, but obviously it's not going to match perfectly. The goal here is to have something we can run when we make changes to the generation loop to verify that it is correct. Small differences are fine; in that case, we can manually override it. The test will have to be run manually, in any case. It's quite fast, running in 3 minutes. Example: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/open-instruct-dev/work/01KBNV5ANCJA0G05VAFX0D5N7V?taskId=01KBNV5ANF725E24RGH2PS5VPC&jobId=01KBNV5ASJJNVW8ZJEBQH4R2VT).

The test itself, at least on the same type of machine (A100/H100), is consistent. I ran it 100 times in a loop and it passed every time. 

---

> [!NOTE]
> Adds GPU integration tests validating deterministic generation (with/without tools) against golden outputs, moves GPU tests out of CPU suite, and adds a Beaker run script.
> 
> - **Tests (GPU)**:
>   - Add `open_instruct/test_grpo_fast_gpu.py` with vLLM queue-based integration tests, deterministic generation checks (tools and no-tools), tool server bootstrap, and placement group setup.
> - **Golden test data**:
>   - Add `open_instruct/test_data/generation_with_tools_expected.json` and `generation_without_tools_expected.json` for exact token/id matching.
> - **Scripts**:
>   - Add `scripts/train/debug/run_gpu_pytest.sh` to launch GPU pytest on Beaker and persist `test_data` outputs.
> - **Refactor (CPU tests)**:
>   - Trim GPU-dependent and mock vLLM engine tests from `open_instruct/test_grpo_fast.py`; remove unused imports/env handling; keep queue/streaming and data prep tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ad3bd036ca6af9dd7725f7ec803c04e38416f4db. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/open_instruct/test_grpo_fast.py
+++ b/open_instruct/test_grpo_fast.py
@@ -70,8 +70,7 @@ class TestGrpoFastBase(unittest.TestCase):
         utils.check_runtime_leaks()
 
         # Initialize Ray for this test
-        if not ray.is_initialized():
-            ray.init(include_dashboard=False)
+        ray.init(include_dashboard=False)
 
     def _cleanup_ray_queues(self):
         """Clean up all Ray queues created during the test."""


hamishivi (approved):
oh super nice!

PR #1307, opened 2025-12-23
Don't crash on reference policy ckpt issue.

hamishivi:
Sometimes, our reference policy saving gets corrupted somehow and it stuffs up our runs. For the most part, unless we are periodically updating the reference model, this is fine, since we can just reload the original reference model (which doesn't change over training).

This PR allows this, except when we are periodically updating the reference.

Successful GPU tests [here](https://beaker.allen.ai/orgs/ai2/workspaces/open-instruct-dev/work/01KD6H18BQXJBNMMZRTRNQJGZ4?taskId=01KD6H18C1TNXS1E43XG6S84WA&jobId=01KD6H18FW6ZQE6DWNSGCRQTDJ).
Successful 1-gpu run [here](https://beaker.allen.ai/orgs/ai2/workspaces/open-instruct-dev/work/01KD6J4RSDZV0M69MADHGAJP78?taskId=01KD6J4RSNECYFGWD793SFBD9W&jobId=01KD6J4RXENXMW35PR58HSAMF8).

### DIFF
--- a/open_instruct/model_utils.py
+++ b/open_instruct/model_utils.py
@@ -150,6 +150,40 @@ def disable_dropout_in_model(model: torch.nn.Module) -> None:
             module.p = 0
 
 
+def maybe_load_checkpoint(
+    model: torch.nn.Module, checkpoint_path: str, device: torch.device, rank: int, throw_on_error: bool = True
+) -> None:
+    """Load a checkpoint into a model, with optional error handling.
+
+    Args:
+        model: The model to load the checkpoint into.
+        checkpoint_path: Path to the checkpoint file.
+        device: Device to load the checkpoint onto.
+        rank: Global process rank for logging.
+        throw_on_error: if true, throw an error if the checkpoint fails to load.
+    """
+
+    def _load_checkpoint(path: str, dev: torch.device):
+        state_dict = torch.load(path, map_location=dev)
+        if hasattr(model, "module"):
+            # Needed if wrapped by DeepSpeed.
+            model.module.load_state_dict(state_dict)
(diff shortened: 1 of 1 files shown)


open_instruct/model_utils.py
```diff
@@ -188,14 +192,28 @@ def load_ref_policy(
     ref_policy.eval()
 
     if checkpoint_path:
```
finbarrtimbers, line 228: Can we just pull this whole segment out into a free function (maybe_load_checkpoint) which lives at the module level?

open_instruct/model_utils.py
```diff
@@ -188,14 +192,28 @@ def load_ref_policy(
...
+            logger.info(f"{rank=}: Loaded reference policy checkpoint from {path}")
+
+        # Only catch errors gracefully if ref policy won't be updated,
+        # since then we can just re-use the original reference model ckpt.
+        # if we are updating the reference, then we need to load the checkpoint.
+        if ref_policy_update_freq is None or alpha == 0:
+            try:
+                _load_checkpoint(checkpoint_path, device)
+            except Exception as e:
+                logger.error(f"{rank=}: Failed to load reference policy checkpoint from {checkpoint_path}: {e}")
```
finbarrtimbers: Combine these into one statement?

hamishivi: done!

open_instruct/model_utils.py
hamishivi: done!

finbarrtimbers (approved):

PR #1308, opened 2025-12-23
World-aware packing

### LINKED ISSUE #1295: `B = 0` when packed_sequences.query_responses < args.world_size
Because `B = len(packed_sequences.query_responses) // args.world_size`

If we pack our sequences too well, we accidentally skip training.

We should instead throw a warning that there are fewer sequences than world size and set `B=1`

hamishivi:
Previously, our packing tried to pack everything into a minimal number of sequences. This led to issues when we had lots of really short outputs, because then some ranks wouldn't receive samples. Additionally, this could also lead to uneven processing between ranks: one rank might just receive a tiny number of tokens.

This fixes that by instead enforcing that every rank must receive at least one sample. We now enforce that the number of samples per batch == the number of ranks to ensure this is always possible (note that previously this was also true but not enforced).

Since world padding was a different solution to the same problem (but did it by just adding padding, not packing less tightly), I've removed the world padding argument. Should also fix #1295!

Successful GPU tests [here](https://beaker.org/ex/01KD6P1DKFVEHA7BMSC5S7FQN5).
Successful 1-GPU [here](https://beaker.org/ex/01KD6NDBZBMT2EB9DR27R18DN8).
Successful multi-GPU [here](https://beaker.org/ex/01KD6NKC0E583FYZ6BVDQNEBD5).

### DIFF
--- a/open_instruct/rl_utils.py
+++ b/open_instruct/rl_utils.py
@@ -179,16 +179,15 @@ def pack_sequences(
             f"This can happen if vLLM returns N-1 logprobs for N tokens (missing first token logprob)."
         )
         combined_logprobs = query_logprobs + response_logprobs
-        if len(query_response) + len(cur_data) > effective_pack_length:
-            # Only flush if there's data (handles case where single sequence exceeds pack_length)
-            if len(cur_data) > 0:
-                query_responses.append(cur_data)
-                response_masks.append(cur_response_mask)
-                attention_masks.append(cur_attention_mask)
-                num_actions.append(cur_num_actions)
-                packed_seq_lens.append(cur_packed_seq_lens)
-                dones.append(cur_dones)
-                packed_vllm_logprobs.append(cur_vllm_logprobs)
+        # only flush if we have data and we exceed the pack length.
+        if len(query_response) + len(cur_data) > effective_pack_length and len(cur_data) > 0:
+            query_responses.append(cur_data)
+            response_masks.append(cur_response_mask)
+            attention_masks.append(cur_attention_mask)
+            num_actions.append(cur_num_actions)
+            packed_seq_lens.append(cur_packed_seq_lens)
+            dones.append(cur_dones)
+            packed_vllm_logprobs.append(cur_vllm_logprobs)
             cur_data = []
             cur_response_mask = []
             cur_attention_mask = []


open_instruct/grpo_fast.py
```diff
@@ -205,15 +205,13 @@ class Args:
...
     world_size: int | None = None
-    """RUNTIME VALUE: The number of processes (GPUs) to use"""
+    """RUNTIME VALUE: The number of processes (GPUs) to use for training ONLY"""
     num_training_steps: int | None = None
     """RUNTIME VALUE: The number of training_steps to train"""
     local_eval_every: int = 100
     """Run evaluation after this many training steps. This controls in-loop evals, which reuse the generation/reward verifier setup. Set to -1 to disable."""
     save_freq: int = 200
     """How many train steps to save the model"""
-    allow_world_padding: bool = False
```
finbarrtimbers, line 215: I love getting rid of flags! Less options!!

open_instruct/rl_utils.py
```diff
@@ -86,9 +86,43 @@ def pack_sequences(
     pack_length: int,
     pad_token_id: int,
     vllm_logprobs: List[List[float]],
+    min_num_batches: int = 1,
```
finbarrtimbers, line 89: Would this ever be not equal to world_size? If so let's just call it world_size?

open_instruct/rl_utils.py
```diff
@@ -145,14 +179,16 @@ def pack_sequences(
...
-            query_responses.append(cur_data)
-            response_masks.append(cur_response_mask)
-            attention_masks.append(cur_attention_mask)
-            num_actions.append(cur_num_actions)
-            packed_seq_lens.append(cur_packed_seq_lens)
-            dones.append(cur_dones)
-            packed_vllm_logprobs.append(cur_vllm_logprobs)
+        if len(query_response) + len(cur_data) > effective_pack_length:
+            # Only flush if there's data (handles case where single sequence exceeds pack_length)
+            if len(cur_data) > 0:
```
finbarrtimbers: Do we need the conditional? If len(cur_data) == 0, it's a no-op right?

finbarrtimbers (approved):

open_instruct/rl_utils.py
hamishivi: No, because len(query_response) on its own could be larger than effective_pack_length. Simplified the logic here to make it a bit clearer.

open_instruct/rl_utils.py
hamishivi: I lightly prefer calling it this, since its clear what it does to the output. Additionally, in the seq-parallel pr, since we add the sequence parallel dimension, really we only want to pack across the data-parallel dimension, so this would be 'dp_world_size' instead of 'world_size'.

PR #1309, opened 2025-12-23
Add sequence parallel via DeepSpeed Ulysses

### LINKED ISSUE #675: Context parallelism
Hello,

Are there plans to include context parallelism/ring attention to support larger contexts any time soon?

Thanks!

hamishivi:
Adds support for sequence parallel training via deepspeed ulysses, following this great guide: https://www.deepspeed.ai/tutorials/ulysses-alst-sequence-parallelism/

Basically, this just leverages features in HF transformers + deepspeed, and then edits the dataloader accordingly. This allows us to train on much longer sequences, or even train on fewer GPUs (trading off on training time). Also edits some of the e.g. metrics code to account for the sequence ranks.

Some notes:
- Relies on #1308 for the packing change, since long sequences excerbate the packing issues.
- The sequence parallel size is limited by the number of kv heads, which can be small with GQA (like 4-9). The code will throw an error and crash if it can't support the seq parallel you ask for. In future, we would have things like ring-attention via olmo-core for extending further.
- For some reason, Nvidia limits sequence lengths to 64k in the batch dim. If you implicitly ask for sequences longer than this, flash-attn will throw a misconfiguration error. See https://github.com/Dao-AILab/flash-attention/issues/1786 for details.
- Seq parallel only works with deepspeed stage 3, this is checked in the post_init now.

Fixes #675

Test multi-node rlzero math job here: https://beaker.allen.ai/orgs/ai2/workspaces/tulu-thinker/work/01KDVRYM483C00N8TGT416SRKR?taskId=01KDVRYM4HDYZ40424Q90A1RGX&jobId=01KDVRYM87K7SENFRZD8ZHTMQY

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -371,7 +371,6 @@ class PolicyTrainerRayProcess(RayProcess):
         self.world_size = world_size
         self.local_rank = local_rank
         self.dp_world_size = world_size // args.sequence_parallel_size
-        self._streaming_dataloader = None
         self._data_prep_actor_name = data_prep_actor_name
 
     def get_dataloader_state(self) -> dict[str, Any]:
--- a/scripts/train/debug/grpo_fast_3_gpu.sh
+++ b/scripts/train/debug/grpo_fast_3_gpu.sh
@@ -1,3 +1,6 @@
+# debug script using 3 gpus to specifically test sequence parallelism
+# with minimal gpus (2 for training for seq parallel, 1 for inference)
+# useful for debugging multi-gpu training without too many gpus
 export VLLM_ALLOW_INSECURE_SERIALIZATION=1
 export VLLM_DISABLE_COMPILE_CACHE=1
 export VLLM_USE_V1=1


open_instruct/utils.py
```diff
@@ -2578,6 +2588,59 @@ def get_beaker_experiment_url() -> str | None:
...
+        self.sp_world_size = sp_world_size
+        self.device = device
+        self.pad_token_id = pad_token_id
+
+    def split_collated_batch(self, data: CollatedBatchData) -> CollatedBatchData:
+        """Get this rank's shard of a CollatedBatchData for sequence parallelism."""
+        fields = [f.name for f in dataclasses.fields(data)]
+
+        # Find max sequence length across all ranks to ensure consistent padding
+        local_max = max(t.shape[-1] for t in getattr(data, fields[0]))
```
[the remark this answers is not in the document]
hamishivi: If a batch is empty, we should crash...

open_instruct/grpo_fast.py
```diff
@@ -740,13 +770,19 @@ def load(self, path: str, map_location=None):
...
+                    self.model.mpu = None
                 path, states = self.model.load_checkpoint(
                     args.checkpoint_state_dir,
                     load_module_strict=True,
                     load_optimizer_states=True,
                     load_lr_scheduler_states=True,
                     load_module_only=False,
                 )
+                if old_mpu is not None:
+                    self.model.mpu = old_mpu
```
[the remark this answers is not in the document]
hamishivi: If the loading fails, we should crash anyway.

open_instruct/grpo_fast.py
```diff
@@ -1276,6 +1368,11 @@ def save_checkpoint_state(self, checkpoint_state_dir: str, client_state: dict[st
             client_state["ref_policy_saved"] = True
 
         # Save the main model checkpoint with enhanced client state
+        # mpu is just used for sequence parallel, so we remove it for saving, and then re-add it after.
+        old_mpu = None
+        if self.model.mpu is not None:
+            old_mpu = self.mpu
+            self.model.mpu = None
         self.model.save_checkpoint(checkpoint_state_dir, client_state=client_state)
```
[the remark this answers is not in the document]
hamishivi: If we have an error, we should crash anyway.

scripts/train/debug/grpo_fast_tool.sh
finbarrtimbers, line 1: Should we merge/replace `tool_grpo_fast.sh` with this?

open_instruct/utils.py
```diff
@@ -2578,6 +2588,59 @@ def get_beaker_experiment_url() -> str | None:
         return None
 
 
+class UlyssesSPSplitter:
```
finbarrtimbers, line 2584: Can you add a test for this?

open_instruct/utils.py
```diff
@@ -2578,6 +2588,59 @@ def get_beaker_experiment_url() -> str | None:
...
+    """Splits CollatedBatchData for Ulysses sequence parallelism.
+
+    Adapted from the UlyssesSPDataLoaderAdapter
+    (https://github.com/deepspeedai/DeepSpeed/blob/master/deepspeed/runtime/sequence_parallel/ulysses_sp.py#L427)
+    Rather than wrapping a dataloader, we just want to split a given batch into the shards across sp group.
+    """
+
+    def __init__(self, sp_rank: int, sp_group, sp_world_size, device, pad_token_id):
```
finbarrtimbers: Type annotations please!

finbarrtimbers: Also wdyt about making it a dataclass?

open_instruct/utils.py
finbarrtimbers: What's `fields[0]`?

open_instruct/utils.py
```diff
@@ -2578,6 +2588,59 @@ def get_beaker_experiment_url() -> str | None:
...
+        chunk_len = max_seqlen // self.sp_world_size
+
+        # Compute start and end indices for this rank's chunk
+        start_idx = chunk_len * self.sp_rank
+        end_idx = chunk_len * (self.sp_rank + 1)
+
+        # slice and pad tensors for this sp rank
+        kwargs = {}
+        for field in fields:
+            tensors = getattr(data, field)
```
finbarrtimbers: Inline this, as it's only used once?

open_instruct/utils.py
```diff
@@ -2578,6 +2588,59 @@ def get_beaker_experiment_url() -> str | None:
...
+            if field == "query_responses":
+                pad_value = self.pad_token_id
+            elif field == "vllm_logprobs":
+                pad_value = INVALID_LOGPROB
+            else:
+                pad_value = 0
+            sharded = []
+            for t in tensors:
+                padded_sliced = F.pad(t, (0, max_seqlen - t.shape[-1]), value=pad_value)[:, start_idx:end_idx]
```
finbarrtimbers, line 2627: Can you comment what this is doing?

open_instruct/utils.py
```diff
@@ -2578,6 +2588,59 @@ def get_beaker_experiment_url() -> str | None:
...
+    def __init__(self, sp_rank: int, sp_group, sp_world_size, device, pad_token_id):
+        self.sp_rank = sp_rank
+        self.sp_group = sp_group
+        self.sp_world_size = sp_world_size
+        self.device = device
+        self.pad_token_id = pad_token_id
+
+    def split_collated_batch(self, data: CollatedBatchData) -> CollatedBatchData:
+        """Get this rank's shard of a CollatedBatchData for sequence parallelism."""
+        fields = [f.name for f in dataclasses.fields(data)]
```
finbarrtimbers: As it's only used twice, wdyt about inline it? I think you only use `fields[0]` to get the sequence length, right? Can we directly access that instead?

open_instruct/utils.py
```diff
@@ -1438,6 +1443,11 @@ def get_train_ds_config(
         "data_types": {"grad_accum_dtype": grad_accum_dtype if grad_accum_dtype else "fp32"},
     }
 
+    if sequence_parallel_size > 1:
```
finbarrtimbers: What happens if sequence_parallel_size = 1? Is this all a no-op? If so, it'd be great to include it to have less code paths.

open_instruct/utils.py
```diff
@@ -68,6 +70,7 @@
 from transformers.integrations import HfDeepSpeedConfig
 
 from open_instruct import logger_utils
+from open_instruct.data_types import CollatedBatchData
```
finbarrtimbers: Can we do `from open_instruct import data_types` as we don't use it a lot?

open_instruct/rl_utils.py
```diff
@@ -284,7 +284,11 @@ def calculate_advantages_packed(
 def masked_mean(
     values: torch.Tensor, mask: torch.Tensor, axis: int | None = None, denominator: float | None = None
 ) -> torch.Tensor:
-    """Compute mean of tensor with masked values."""
+    """Compute mean of tensor with masked values.
+
+    Returns 0 if mask is empty (no valid elements) to avoid division by zero.
```
finbarrtimbers, line 289: Can you add a test covering this case?

open_instruct/grpo_fast.py
```diff
@@ -801,10 +847,26 @@ def load(self, path: str, map_location=None):
...
                 if hasattr(self, "ref_policy_checkpoint_path")
                 else None,
+                mpu=self.mpu,
                 ref_policy_update_freq=args.ref_policy_update_freq,
                 alpha=args.alpha,
             )
         self.local_metrics = utils.MetricsTracker(device=self.device)
+
+        if self.mpu is not None:
+            self.sp_group = groups._get_sequence_parallel_group()
```
finbarrtimbers: Are these ever referenced again? Do they need to be members?

open_instruct/grpo_fast.py
```diff
@@ -1221,25 +1299,47 @@ def train(self, data_BT: CollatedBatchData, pad_token_id: int) -> dict[str, floa
...
-                        self.local_metrics[f"objective/kl{j}_avg"] = loss_stats_B["kl"][j].mean()
-                    self.local_metrics["loss/kl_avg"] = loss_stats_B["kl_loss"].mean()
-                self.local_metrics["loss/policy_avg"] = loss_stats_B["pg_loss"].mean()
-                self.local_metrics["loss/total_avg"] = loss_stats_B["loss"].mean()
-                self.local_metrics["policy/clipfrac_avg"] = loss_stats_B["pg_clipfrac"].mean()
-                self.local_metrics["val/ratio"] = loss_stats_B["ratio"].mean()
-                self.local_metrics["val/ratio_var"] = loss_stats_B["ratio"].var()
-                if self.args.record_entropy:
-                    self.local_metrics["policy/entropy_avg"] = loss_stats_B["entropy"].mean()
+                # Compute weighted averages using token counts (important for sequence parallel)
```
finbarrtimbers: Can we pull this whole bit out into a function?

open_instruct/grpo_fast.py
```diff
@@ -1221,25 +1299,47 @@ def train(self, data_BT: CollatedBatchData, pad_token_id: int) -> dict[str, floa
...
+                    if self.args.record_entropy:
+                        self.local_metrics["policy/entropy_avg"] = 0.0
                 self.local_metrics["lr"] = self.scheduler.get_last_lr()[0]
+                # Store total token count for this rank (for weighted averaging across ranks)
+                self.local_metrics["_token_count"] = total_valid_tokens
                 return self.local_metrics.get_metrics_list()
 
     def save_checkpoint_state(self, checkpoint_state_dir: str, client_state: dict[str, Any]) -> None:
         args = self.args
-
```
finbarrtimbers, line 999: Undo this please!

open_instruct/grpo_fast.py
```diff
@@ -2376,7 +2489,40 @@ def one_training_step(
 
     ray.get(actor_manager.report_training_step_time.remote(train_timer.duration))
 
-    average_metrics = {k: sum(m[k] for m in metrics_list) / len(metrics_list) for k in metrics_list[0]}
+    # Compute token-weighted average for metrics (important for sequence parallel)
```
finbarrtimbers: Can we make this a function too?

finbarrtimbers (changes requested):

open_instruct/utils.py
hamishivi: Ah, fair point, that makes it a bit cleaner I think!

open_instruct/utils.py
hamishivi: should be query_responses. Let me make that explicit.

open_instruct/utils.py
hamishivi: I wanted to automatically run this over everything in the collated batch, instead of hardcoding in the attributes, since in future if you add things you should have to think carefully about what the SP behaviour should be. 

Fixed fields[0] to explicitly get the attribute I wanted, and inlined the loop to the latter point!

open_instruct/utils.py
hamishivi: Good point, deepspeed defaults to this as value 1 anyway. Also removed an unnecessary conditional around the mpu setup.

open_instruct/grpo_fast.py
hamishivi: good point... I think this was from back when I was doing loss reduction per-sequence group, in which case you need this (so you can only reduce across the sp group). But now the loss is normalised across the whole training world, so we don't need to care about sp groups.

scripts/train/debug/grpo_fast_tool.sh
hamishivi: Ah, I'll delete this in favour of the other tool scripts. I'll also edit the large test script to use sequence parallel.

open_instruct/utils.py
hamishivi: done!

open_instruct/utils.py
```diff
@@ -1428,16 +1432,19 @@ def get_train_ds_config(
         zero_opt_dict["stage3_max_live_parameters"] = 0
         zero_opt_dict["stage3_max_reuse_distance"] = 0
 
-    return {
+    config = {
```
finbarrtimbers: Keep the inline structure to minimize the diff? (i.e. return { ...})

open_instruct/grpo_fast.py
```diff
@@ -623,16 +648,17 @@ def prepare_collated_data_for_workers(
...
+        batch_data = CollatedBatchData(
+            query_responses=collated_query_responses,
+            attention_masks=collated_attention_masks,
+            position_ids=collated_position_ids,
+            advantages=collated_advantages,
+            response_masks=collated_response_masks,
+            vllm_logprobs=collated_vllm_logprobs,
         )
+        # Assign the same batch data to all SP ranks within this DP group
+        for sp_j in range(sequence_parallel_size):
```
finbarrtimbers: I think this will now be redundant with the dataloader abstraction, as we'll just have each DP group load a batch?

open_instruct/grpo_fast.py
```diff
@@ -749,13 +788,19 @@ def load(self, path: str, map_location=None):
                     f"Skipping loading checkpoint state from {args.checkpoint_state_dir} because it does not exist!"
                 )
             else:
+                old_mpu = None
```
finbarrtimbers: Does this need a conditional? If `self.mpu is None`, this is all a no-op right?

open_instruct/grpo_fast.py
```diff
@@ -749,13 +788,19 @@ def load(self, path: str, map_location=None):
...
+                    old_mpu = self.mpu
+                    self.model.mpu = None
                 path, states = self.model.load_checkpoint(
                     args.checkpoint_state_dir,
                     load_module_strict=True,
                     load_optimizer_states=True,
                     load_lr_scheduler_states=True,
                     load_module_only=False,
                 )
+                if old_mpu is not None:
```
finbarrtimbers: Same Q here as above.

open_instruct/grpo_fast.py
```diff
@@ -810,10 +855,26 @@ def load(self, path: str, map_location=None):
                 checkpoint_path=self.ref_policy_checkpoint_path
                 if hasattr(self, "ref_policy_checkpoint_path")
                 else None,
+                mpu=self.mpu,
                 ref_policy_update_freq=args.ref_policy_update_freq,
                 alpha=args.alpha,
             )
         self.local_metrics = utils.MetricsTracker(device=self.device)
+
+        if self.mpu is not None:
```
finbarrtimbers, line 571: What happens. here if `self.mpu is None`?

open_instruct/grpo_fast.py
```diff
@@ -1438,6 +1536,18 @@ def get_bundle_index(rank, num_gpus_per_node):
             self.models.append(worker_policy)
 
 
+def compute_token_weights(metrics_list: list[dict[str, float]]) -> list[float]:
+    """Compute token-weighted weights for averaging metrics across ranks.
+
+    Important for sequence parallel where different ranks may have different token counts.
+    """
+    token_counts = [m.get("_token_count", 1.0) for m in metrics_list]
```
finbarrtimbers: Make this a for-loop to avoid iterating over the list twice? We can calculate token_counts and total_tokens simultaneously.

open_instruct/grpo_fast.py
```diff
@@ -1003,6 +1090,11 @@ def train(self, data_BT: CollatedBatchData, pad_token_id: int) -> dict[str, floa
         Returns:
             Dictionary of training metrics (loss, KL, entropy, etc.).
         """
+        # do splitting before GPU transfer so we only move data once
+        if self.splitter is not None:
+            with Timer("✂️ Splitting batch for SP", noop=self.rank != 0):
```
finbarrtimbers, line 819: Use the decorator version instead? Then we don't have as much nesting.

open_instruct/grpo_fast.py
hamishivi: good point!

open_instruct/grpo_fast.py
hamishivi: Then we don't have a splitter, and then later on during dataloading we don't do the SP splitting (`if self.splitter is not None:` later on). `mpu` is `None` when `sequence_parallel=1` in the DeepSpeed API, so I think this is the correct check? I inlined the `sp_*` variables to make it clearer they are only used to init the splitter.

open_instruct/grpo_fast.py
hamishivi: I think because we need to check the rank to avoid printing 8 times on every node, we can't do this as a decorator?

open_instruct/grpo_fast.py
hamishivi: done! TBH I would hope this never really hits lengths where python would struggle here.

open_instruct/grpo_fast.py
hamishivi: Yes, now this is fine!

open_instruct/grpo_fast.py
```diff
@@ -331,9 +339,6 @@ def __post_init__(self):
                 "When load_ref_policy=False, beta must be 0.0. "
                 f"Got beta={self.beta}. Set --beta 0.0 or --load_ref_policy to use KL penalty."
             )
-        assert self.num_samples_per_prompt_rollout * self.num_unique_prompts_rollout >= sum(
```
hamishivi, line 334: I think main was broken cuz this was there

scripts/train/debug/grpo_fast_3_gpu.sh
finbarrtimbers, line 1: Can you add a comment explaining why this file exists (I think we should eventually do this for all of the debug scripts tbh)?

open_instruct/grpo_fast.py
```diff
@@ -363,16 +368,11 @@ def __init__(
...
-            fs_local_rank=local_rank,
-            num_training_steps=args.num_training_steps,
-            work_dir=args.output_dir,
-            dp_world_size=world_size,
-        )
-        self.dataloader = iter(self._streaming_dataloader)
+        self.world_size = world_size
+        self.local_rank = local_rank
+        self.dp_world_size = world_size // args.sequence_parallel_size
+        self._streaming_dataloader = None
```
finbarrtimbers: Can we just not define _streaming_dataloader until it's needed?

PR #1311, opened 2025-12-26
Upgrade deepspeed

hamishivi:
Previously we were pinned to < 0.17.3 due to an unknown issue... well I (and claude) worked this out! The core issue was that [this PR in deepspeed](https://github.com/deepspeedai/DeepSpeed/pull/7266) added device_id to the init_process_group call, which caused NCCL hangs on our weight sync, potentially due to our funky having multiple process groups. This does a simple fix: just do the torch-level init ourself without the device ids. Note that changing to the gloo backend also worked, but this causes really, really slow weight syncs, so it doesn't scale.

successful gpu tests [here](https://beaker.allen.ai/orgs/ai2/workspaces/open-instruct-dev/work/01KDD74TKZ0ZKVQY1TSH9AKTD7?taskId=01KDD74TM8SW7Q98CM8PYF43QJ&jobId=01KDD74TR1RBBS7E8FS55C1FS0).
successful multi-node run [here](https://beaker.allen.ai/orgs/ai2/workspaces/open-instruct-dev/work/01KDD6PR565NP5XFV7VMXWCTN1?taskId=01KDD6PR5EXTCDNVSE91C5AP0T&jobId=01KDD6PRFXBF57B5MG4KEWN67E).

Claude's summary:
# DeepSpeed 0.17.3+ Weight Sync Hang Fix

## Problem
Training hangs after one training step at the weight sync stage when using DeepSpeed versions >= 0.17.3 (including 0.18.x).

## Root Cause
DeepSpeed 0.17.3 introduced a change ([PR #7266](https://github.com/deepspeedai/DeepSpeed/pull/7266)) that sets `device_id` in `torch.distributed.init_process_group()`. This change:

1. Binds the NCCL communicator to a specific GPU device
2. Conflicts with custom process groups created for weight synchronization between DeepSpeed learners and vLLM engines
3. Results in NCCL hangs during broadcast operations when multiple process groups exist with inconsistent device bindings

Related PyTorch issue: [#153960](https://github.com/pytorch/pytorch/issues/153960) - "Passing `device_id` to `torch.distributed.init_process_group()` results in NCCL randomly hanging"

## Solution
Pre-initialize `torch.distributed` ourselves WITHOUT `device_id` before DeepSpeed does. DeepSpeed will detect it's already initialized and wrap it instead of re-initializing with `device_id`.

```python
# Pre-initialize torch.distributed WITHOUT device_id to avoid NCCL hangs.
# DeepSpeed 0.17.3+ sets device_id in init_process_group which can cause hangs
# when multiple process groups exist (e.g., for weight sync to vLLM).
# By initializing first, DeepSpeed will detect it and wrap it instead of re-initializing.
if not torch.distributed.is_initialized():
    torch.distributed.init_process_group(
        backend="nccl",
        timeout=timedelta(minutes=args.backend_timeout),
    )
deepspeed.init_distributed(timeout=timedelta(minutes=args.backend_timeout))
```

This allows us to keep using NCCL for both DeepSpeed and weight sync, avoiding the need for slower gloo backend.

## Files Modified
- `open_instruct/grpo_fast.py`: Pre-initialize torch.distributed before DeepSpeed
- `open_instruct/vllm_utils.py`: Added `device_id` parameter support (for future use)
- `pyproject.toml`: Removed DeepSpeed version restriction

## Performance Comparison
| Backend | Weight Sync Time |
|---------|------------------|
| gloo (workaround) | ~0.8-1.0 seconds |
| nccl (this fix) | ~0.12-0.16 seconds |

The NCCL solution is **6x faster** for weight synchronization.

## Testing
Verified fix works with:
- DeepSpeed 0.18.3
- PyTorch 2.8.0
- NCCL 2.27.3
- Training completed 6 steps with successful weight syncs (~0.12s each)

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -2028,7 +2028,7 @@ requires-dist = [
     { name = "bitsandbytes", marker = "sys_platform != 'darwin'", specifier = ">=0.44.1" },
     { name = "datasets", specifier = ">=4.0.0" },
     { name = "debugpy", specifier = ">=1.8.13" },
-    { name = "deepspeed" },
+    { name = "deepspeed", specifier = ">=0.18.3" },
     { name = "fastapi", marker = "extra == 'code'", specifier = ">=0.100.0" },
     { name = "flash-attn", marker = "sys_platform != 'darwin'", specifier = ">=2.8.3" },
     { name = "hf-transfer", specifier = ">=0.1.8" },


finbarrtimbers:
Nice find!!

PR #1312, opened 2025-12-27
Tool refactor [for real this time]

hamishivi:
# Tool System Refactor

**TL;DR**: Brand new tool system that allows for closer integration with vllm, multiple tool calls in-loop, etc. We separate tool parsing from tool execution, and into separate classes.

This is a big change, apologies! Unfortunately, I wanted to implement the core things, which required ripping out a lot of code and adding a bunch more. I've tried to keep things simple for adding tools in future.

For now, olmo 3 native support is implemented on branch `tool-refactor-olmo3-tool-parser`, which builds off this branch. This is because it requires upgrading vllm, which I'll do in a future PR.

Future TODOs (will add in future PRs, just tracking here):
- Verifiers / OpenEnv compatibility (?)
- Some sort of caching system (reduce API costs / speed up tool calls)
- Differentiate different tool errors (API error, model call error)
- (potentially) better session/client handling
- (potentially) full DR Tulu support (adding adaptive rubrics, more complex rubric setups)
- (potentially) dynamically adding tools at runtime...? (e.g. connecting to different mcp servers that change what they expose)

In the process of merging this! what to change:
- [x] basic initial refactor
- [x] better debug scripts
- [x] better logging of per-tool metrics
- [x] vllm parser
- [x] olmo3 parser
- [x] crawl4ai proxy tool
- [x] per-sample tools
- [x] dr tulu parser + tool + integration
- [x] mcp tools + basic weather sample
- [x] fix for ground truth utils (string f1 can optionally take list, better debugging)
- [x] better retry / rate limit handling (using https://github.com/mjpieters/aiolimiter?)
- [ ] better env suport (verifiers envs, bash tools, etc... needs a bit more research)
- [x] max calls exceeded behaviour
- [ ] Add a key to the python server to make it not exposed.
- [ ] clean up all the rlvr tokenization files
- [ ] anything else ive forgotten in other prs after the above is done

### New `open_instruct/tools/` Module

```
tools/
├── __init__.py       # Clean exports for the package
├── utils.py          # Tool, ToolCall, ToolOutput, BaseToolConfig base classes
├── config.py         # ToolArgs (CLI), ToolConfig (internal), TOOL_REGISTRY, create_tool_parser
├── parsers.py        # ToolParser ABC + implementations (Vllm, Legacy, DRTulu) + VLLM_PARSERS registry
├── proxy.py          # ToolProxy + ToolActor (Ray wrapper pattern)
├── tools.py          # Actual tool implementations (PythonCodeTool, SerperSearchTool, etc.)
└── code_server/      # Python execution server
```

### The Tool Pattern

Every tool follows the same shape:

```python
@dataclass
class MyToolConfig(BaseToolConfig):
    tool_class: ClassVar[type[Tool]] = MyTool
    some_option: int = 10
    # tag_name inherited from BaseToolConfig

class MyTool(Tool):
    _default_tool_function_name = "mytool"
    _default_tool_description = "Does something useful"
    _default_tool_parameters = {  # JSON Schema for your args
        "type": "object",
        "properties": {
            "query": {"type": "string", "description": "The search query"},
            "limit": {"type": "integer", "description": "Max results"},
        },
        "required": ["query"],
    }

    def __init__(self, some_option: int = 10, tag_name: str | None = None):
        self._tag_name = tag_name
        # ...

    def __call__(self, query: str, limit: int = 10) -> ToolOutput:
        # Tools can accept any arguments - define what you need
        # The parser extracts args from model output and passes them as kwargs
        return ToolOutput(output=..., called=True, ...)
```

Then add to registry:
```python
TOOL_REGISTRY = {
    "mytool": MyToolConfig,  # Just the config - tool_class is a ClassVar on the config
}
```

CLI args are auto-generated. All tools are wrapped in Ray actors via `ToolProxy` for safe cross-process communication.

Access individual tool configs via `config.get_tool_config("tool_name")`.

### CLI Usage

Tools are specified with `--tools` and configured via `--tool_configs` (a 1:1 list of JSON objects):

```bash
--tools mcp python --tool_configs '{"tool_names": "snippet_search"}' '{"api_endpoint": "http://..."}'
```

Use `{}` for defaults:
```bash
--tools s2_search serper_search --tool_configs '{}' '{"num_results": 10}'
```

### Parser System

We support three parser families:

- **`legacy`**: Original `<tool_name>content</tool_name>` XML style
- **`dr_tulu`**: For MCP tools with their own routing (from DR Tulu project)
- **`vllm_*`**: Native vLLM tool parsers (`vllm_hermes`, `vllm_llama3_json`, `vllm_qwen3_coder`)

Parsers handle:
1. Extracting tool calls from model output
2. Formatting tool responses back
3. Providing stop sequences

vLLM parsers are wrappers around their vllm classes. In the future, we can hopefully remove more and more of the scaffolding around the vLLM parsers, and completely remove the legacy and dr_tulu parsers. For now, I would like to keep them around, since a tight integration isn't really possible (and, there are some higher-level concerns).

### Per-Tool Metrics

Now tracking per-tool call rates and counts:

```
val/tool_search_call_rate      # % samples that called search
val/tool_search_total_calls    # Total calls across all samples
val/tool_search_calls_per_sample
```

We do this for all tools in the training job.

### Tag Name Override

We can override the names tools use (for simpler names or to match model training format):

```bash
--tools s2_search --tool_tag_names search
```

Now `S2SearchTool` responds to `<search>` instead of `<s2_search>`, and its OpenAI function definition uses `search` as the function name.

Multiple tools:
```bash
--tools s2_search serper_search --tool_tag_names papers web
```

## Available Tools

| Name | Description |
|------|-------------|
| `python` | Code execution via FastAPI server |
| `serper_search` | Google search via Serper API |
| `massive_ds_search` | Wikipedia/doc retrieval via massive_ds |
| `s2_search` | Semantic Scholar papers |
| `mcp` | DR Agent MCP tools wrapper (snippet_search, google_search, massive_serve, browse_webpage) |

## Adding a New Tool

1. Create `YourTool` class in `tools.py` extending `Tool`:
   - Set `_default_tool_function_name`, `_default_tool_description`, `_default_tool_parameters`
   - Implement `__call__(self, **kwargs) -> ToolOutput` with whatever arguments your tool needs

Note that the names, descriptions, and parameters are used to auto generate things like CLI args and system prompts, so make sure they are correct and reasonably well-written!

2. Create `YourToolConfig` dataclass in `tools.py`:
   - Inherit from `BaseToolConfig`
   - Set `tool_class: ClassVar[type[Tool]] = YourTool`
   - Add configuration fields (with defaults)

3. Register in `config.py`:
   ```python
   TOOL_REGISTRY = {
       ...,
       "yourtool": YourToolConfig,
   }
   ```

### DIFF
--- a/open_instruct/data_loader.py
+++ b/open_instruct/data_loader.py
@@ -530,7 +530,7 @@ def accumulate_inference_batches(
                 result.masks[i].append(1)
                 result.logprobs[i].append(float("nan"))
 
-        decoded_responses = tokenizer.batch_decode(result.responses, skip_special_tokens=True)
+        decoded_responses = tokenizer.batch_decode(result.responses, skip_special_tokens=False)
 
         k_queries = repeat_each([query], generation_config.n)
         k_ground_truths = repeat_each([ground_truth], generation_config.n)


open_instruct/tools/utils.py
```diff
@@ -0,0 +1,119 @@
...
+
+        Returns:
+            Dict in OpenAI tool format with type, function name, description, and parameters.
+        """
+        return {
+            "type": "function",
+            "function": {
+                "name": self.tool_function_name,
+                "description": self.tool_description,
+                "parameters": self.tool_parameters,
```
pdasigi: It's nice that you have a way to standardize the function specification format. Can we standardize the way the parameters are specified too?

open_instruct/tools/utils.py
```diff
@@ -0,0 +1,119 @@
...
+
+        config = MyToolConfig(param1="custom")
+        tool = config.build()  # Equivalent to MyTool(param1="custom", param2=10)
+    """
+
+    if TYPE_CHECKING:
+        # ClassVars are not included in asdict(), so they won't be passed to build()
+        tool_class: ClassVar[type[Tool]]
+
+    tag_name: str | None = None
```
pdasigi: Minor: Why not just call this "overriden_name" or something similar to make it more explicit?

open_instruct/tools/tools.py
```diff
@@ -0,0 +1,699 @@
...
+    num_results: int = 5
+    """Number of results to return from Serper."""
+
+
+# =============================================================================
+# DrAgentMCPTool + Config
+# =============================================================================
+
+
+class DrAgentMCPTool(Tool):
```
pdasigi, line 439: It would be helpful to also have a more general MCP tool wrapper that can take any MCP server. Maybe we can just make this class more general.

open_instruct/tools/utils.py
hamishivi: Adding pydantic-based annotations to infer this!

open_instruct/tools/tools.py
hamishivi: Done via a GenericMCPTool! Now we can use the mcp tool to grab tool definitions and then the general mcp tool will mediate any calls made to tools that it manages.

hamishivi:
Closing since this is now entirely merged <3

PR #1316, opened 2025-12-30
Fix warnings from GRPO training logs

finbarrtimbers:
## Summary

This PR fixes several warnings that appeared in GRPO training experiment logs and fixes a bug that caused tool-use training to fail.

### Bug Fix
- **max_tool_calls type conversion**: Fixed a TypeError where `max_tool_calls` values were compared as strings instead of integers, causing tool-use GRPO training to fail with `< not supported between instances of int and str`

### Warning Fixes
- **PyTorch device_id warning**: Added `device_ids=[self.local_rank]` to `torch.distributed.barrier()` calls to prevent potential hangs
- **Environment variables**: Added to `ray_node_setup.sh` to suppress various warnings:
  - `RAY_ACCEL_ENV_VAR_OVERRIDE_ON_ZERO=0` - Suppresses Ray FutureWarning
  - `VLLM_WORKER_MULTIPROC_METHOD=spawn` - Suppresses vLLM multiprocessing warning
  - `GLOO_LOG_LEVEL=ERROR` - Reduces Gloo verbosity

## Warnings Analysis

| Warning | Description | Single GPU | Multi-node | Tools | Fixed |
|---------|-------------|:----------:|:----------:|:-----:|:-----:|
| Ray FutureWarning | Accelerator visible devices override | ✓ | ✓ | ✓ | ✅ |
| ProcessGroupNCCL device_id | GPU device unknown warning | ✓ | | | ✅ |
| DeepSpeed CPU accelerator | Setting accelerator to CPU | ✓ | ✓ | ✓ | ✅ |
| vLLM multiprocessing | Must use spawn method | ✓ | ✓ | ✓ | ✅ |
| Checkpoint state | Directory not found (expected) | ✓ | ✓ | ✓ | N/A |
| vLLM insecure serialization | Pickle serialization (intentional) | ✓ | ✓ | ✓ | N/A |
| FlashInfer | Not available, using PyTorch fallback | ✓ | ✓ | ✓ | N/A |
| vLLM sampling params | HF generation config override | ✓ | ✓ | ✓ | N/A |
| torch_dtype deprecation | vLLM internal (cannot fix) | ✓ | ✓ | ✓ | N/A |
| Gloo peer messages | Informational (vLLM internal) | ✓ | ✓ | ✓ | N/A |
| CUDA graph dispatching | Expected with enforce_eager | ✓ | | | N/A |

## Verification Experiments

All three test scripts pass with these changes:

| Experiment | URL |
|------------|-----|
[...]

Issue #1296, opened 2025-12-18
Post-training loss curves and training time estimates for OLMo-2 1B


BrownianNotion:
Hi team,
Would you be able to share the post-training loss curves and training time estimates for each post-training stage of the OLMo-2 1B model on an 8xH100 cluster? I cannot seem to find these in the OLMo-2 paper.

I've reviewed #485 and the OLMo-3 paper (which mentions a 9-day estimate), but both reference either larger models or more GPUs, so I'm hoping to get figures specific to this setup.

Thanks so much for any guidance!

hamishivi:
Hi, 

SFT: 9 hours on one 1 H100 node (8 gpus).
DPO: 2 hours on one 1 H100 node (8 gpus).
RL: 2000000 episodes took roughly 43 hours on 16 H100s (2 nodes). We did two stages of RL training with same cost (so ~86 hours on 16 gpus all up).

Sharing the curves is a bit harder... I am not sure on the precise way to move out our wandb curves. Ping me again in the new year and I can see what I can do!

BrownianNotion:
Thanks @hamishivi, will reach out in the new year about Wandb! DPO estimate lines up with that I've seen (perhaps closer to 1.5 hours with deepspeed stage 0). Haven't done any full SFT runs but here are some of my estimates on one 8xH100 node (based on tqdm and running ~200 steps):

- **default:** (`docs/olmo2.md`) 8+ hours (lines up with your estimate)
- **`per_device_batch_size=4`, `gradient_accumulation_steps=4`:** (same effective batch size) 5.5-6 hours
- **deepspeed stage 0:** 3.5-4 hours (`per_device_batch_size=2`)

It seems using deepspeed stage 0 instead of stage 2 gives huge time savings given that the model already fits in memory.

BrownianNotion:
Hey @hamishivi, did you have any luck on sharing the wandb curves?

hamishivi:
SFT: https://wandb.ai/ai2-llm/open_instruct_public/runs/mihc7mab
DPO: https://wandb.ai/ai2-llm/open_instruct_public/runs/k34ppe9q
The RL curves didnt transfer for some reason... working on it.

BrownianNotion:
Hey @hamishivi , thanks so much! 
Any chance you know why the loss magnitudes are so different? Green is olmo, blue is my repro. Was the loss function, reduction method etc. changed? It's still a sum loss on `finetune.py` as far as I can see though.

[...]

Issue #751, opened 2025-07-02
KPI of allenai/OLMo-2-0425-1B-SFT is not reproducible


Yi-Yu-Yvonne:
I am attempting to reproduce the evaluation scores reported on Hugging Face for the model [`allenai/OLMo-2-0425-1B-SFT`](https://huggingface.co/allenai/OLMo-2-0425-1B-SFT), but I am seeing inconsistent results.

This seems similar to [Issue #688](https://github.com/allenai/open-instruct/issues/688), but that issue is still unresolved.

I used the following commands to evaluate the model:
```bash
olmes --model allenai/OLMo-2-0425-1B-SFT --task ifeval::tulu --output-dir "output/ifeval"
olmes --model allenai/OLMo-2-0425-1B-SFT --task gsm8k::tulu --output-dir "output/gsm8k"
```

## My results
- `ifeval::tulu`: **0.480591**
- `gsm8k::tulu`: **0.518575**

## Expected results (from [Hugging Face model card](https://huggingface.co/allenai/OLMo-2-0425-1B-SFT))
- `ifeval::tulu`: **50.5**
- `gsm8k::tulu`: **52.1**

Thanks in advance!

hamishivi:
Hi, thanks for your interest! I think the point or so difference is likely due to differences in package versions (e.g. vllm). For example, changing vllm version can yield a 1-2 point change like you see here. The exact commands used to run eval were:

```bash
# ifeval
[...]

PR #1318, opened 2026-01-05
Fix OLMo-2 1B training configs

BrownianNotion:
Hi team,
I've been working on reproducing the OLMo-2 1B post-training results. From the GitHub issues, it seems there's still interest in the OLMo-2 1B model, so I thought it would be good to update the configs. 

Would appreciate any feedback, particularly on the RLVR side - I did my best to reconcile the flags with the new script, but I may have missed some things.

### Summary of changes

1. Updated flags for RLVR's new script `grpo_fast.py` from `grpo_vllm_thread_ray_gtrl.py`, as the latter has now been removed.
2. Fixed flags such as `tokenizer path`, `exp_name` and `chat template_name`. 
(I know the tokenizer is the same across the models, but I thought it'd be less confusing with the 1B path)
3. Changed flags for [efficiency](https://github.com/allenai/open-instruct/issues/1296#issuecomment-3698915804) on 80GB GPUs, given that the 1B model fits on a single GPU (eg. using deepspeed stage 0, not using gradient checkpointing, not using tensor parallel for RLVR). 
This is all in one commit 617dd1d and can be removed if you prefer to be more faithful to the original run, or if you want to accommodate reproductions on lower memory GPUs by default. 

### Other Notes
1. Time estimates in doc were given by @hamishivi https://github.com/allenai/open-instruct/issues/1296#issuecomment-3691946819
2. Added `remap_verifier ifeval=ifeval_old` for RLVR, the [dataset](https://huggingface.co/datasets/allenai/RLVR-GSM-MATH-IF-Mixed-Constraints/viewer/default/train?f[dataset][value]=%27ifeval%27) seems to be in the old format for IF with keys like `func_name`, this flag is needed to avoid an python literal evaluation error that would hang the training script https://github.com/allenai/open-instruct/blob/main/open_instruct/ground_truth_utils.py#L296. 
```
ValueError: malformed node or string on line 1: <ast.Name object at 0x7a0d704f2450>
```
3. Added `export VLLM_ALLOW_INSECURE_SERIALIZATION=1` to avoid vllm warning crashing the script. The warning seems to be intended https://github.com/allenai/open-instruct/pull/1316
4. (Apologies for the alphabetical flag sorting in RLVR, found it easier to reconcile that way)


docs/olmo2.md
```diff
@@ -231,101 +235,92 @@ done
...
-    --actor_num_gpus_per_node 4 8 \
+    --non_stop_penalty_value 0.0 \
     --num_epochs 1 \
-    --vllm_tensor_parallel_size 4 \
-    --lr_scheduler_type constant \
-    --apply_verifiable_reward true \
-    --seed 1 \
-    --num_evals 100 \
+    --num_learners_per_node 4 8 \
+    --num_mini_batches 1 \
```
hamishivi: To exactly match the old args, this should be 2. This is because we get (48*16)/12 = 64 samples per rollout batch per gpu, but set mini_batch_size to 32 in the original script, so we did 2 train batches per rollout batch.

The fact that this took some math to work out is why we changed how the args work now :)

docs/olmo2.md
```diff
@@ -231,101 +235,92 @@ done
...
+    --local_eval_every 5 \
+    --lr_scheduler_type constant \
+    --max_prompt_token_length 2048 \
+    --model_name_or_path allenai/OLMo-2-0425-1B-RLVR1 \
+    --model_revision main \
+    --non_stop_penalty \
+    --non_stop_penalty_value 0.0 \
+    --num_epochs 1 \
+    --num_learners_per_node 4 8 \
+    --num_mini_batches 1 \
```
hamishivi: same comment as above.

hamishivi (commented):
Looks good apart from the num_mini_batches argument. I would also just note that our 1B training was done with an old (deprecated) script that DIDN'T use packing. This made training slower and slightly different loss aggregation. Packing should perform at least similarly with much faster speeds.

[this](https://github.com/allenai/open-instruct/blob/ac827196a549a015980a6545e782bb0853a18c9f/open_instruct/grpo_vllm_thread_ray_gtrl.py) is the commit that the doc was added at, so the grpo_thread_... script at that commit should also work for training with the original command.

docs/olmo2.md
BrownianNotion: Cheers for the pointer, fixed!

BrownianNotion:
@hamishivi I've reverted the commit for more efficient training settings - while attempting to run sft on deepspeed stage 0, the loss trajectory diverged from your run and actually ended up higher than the its starting value. The grad norms also increased significantly. This doesn't happen with deepspeed stage 3 (original setting) and I've attached the loss curves/grad norms below (orange is stage 3, pink is stage 0) 

I'm not sure if this is a bug in deepspeed or some interaction with how `finetune.py` is structured, but I thought it'd be better to leave it out. 
image

hamishivi:
Oh, this is interesting. We are pretty close to replacing our DPO and SFT trainers with olmo-core wrappers, which hopefully shouldn't have such issues.

docs/olmo2.md
```diff
@@ -231,101 +238,93 @@ done
...
---num_evals 100 \
---save_freq 200 \
---reward_model_multiplier 0.0 \
---no_try_launch_beaker_eval_jobs \
---try_launch_beaker_eval_jobs_on_weka \
---gradient_checkpointing \
---with_tracking \
---tokenizer_name_or_path allenai/OLMo-2-1124-7B-DPO \
---tokenizer_revision main
+export VLLM_ALLOW_INSECURE_SERIALIZATION=1
```
hamishivi, line 336: Rather than replacing, could you put in something like:
"A modern reproduction with commit XXX (current commit) would be:"
and then keep the old command and say "The original command, on an older version of open-instruct, was:"

hamishivi (changes requested):

baon6052:
Hi @hamishivi !

Thanks for helping us out on this!

After running SFT via the command in this [PR](https://github.com/allenai/open-instruct/pull/1318/changes#diff-ca54e2e4873933e644ae2ef850f5affed538136451a815ec32a9dafef312db6eR61-R91), which results in [this](https://github.com/allenai/open-instruct/issues/1296#issuecomment-3728293985) loss curve. We went on to do evals. We ran GSM8K and IFEval which resulted in 46% for both, a 5% drop as reported in the OLMo 2 paper (51%).

I am not sure why, but wondering if this could be due to removing `reduce_loss="sum"` [here](https://github.com/allenai/open-instruct/pull/1024)? From the code removed, I can see a [comment](https://github.com/allenai/open-instruct/pull/1024/changes#diff-408d69efc79bb0c4dd513149521ed724e4f8fb12e8db91362600f258a3e22348L793-L799):

>  reduce loss is sum
>  this ensures that we weight all tokens in the dataset equally,
>  rather than weighting each overall example equally when
>  using high amounts of gradient accumulation.
>  this can result in > 5 point improvements in AlpacaEval
>  see https://github.com/huggingface/transformers/issues/24725 for
>  more discussion and details.

As it suggests a 5 point improvement when `reduce_loss="sum"` is used and our SFT run is with the current mean loss, with a 5% drop, do you think perhaps this might be the reason why?

baon6052:
> Hi @hamishivi !
> 
> Thanks for helping us out on this!
> 
> After running SFT via the command in this [PR](https://github.com/allenai/open-instruct/pull/1318/changes#diff-ca54e2e4873933e644ae2ef850f5affed538136451a815ec32a9dafef312db6eR61-R91), which results in [this](https://github.com/allenai/open-instruct/issues/1296#issuecomment-3728293985) loss curve. We went on to do evals. We ran GSM8K and IFEval which resulted in 46% for both, a 5% drop as reported in the OLMo 2 paper (51%).
> 
> I am not sure why, but wondering if this could be due to removing `reduce_loss="sum"` [here](https://github.com/allenai/open-instruct/pull/1024)? From the code removed, I can see a [comment](https://github.com/allenai/open-instruct/pull/1024/changes#diff-408d69efc79bb0c4dd513149521ed724e4f8fb12e8db91362600f258a3e22348L793-L799):
> 
> > reduce loss is sum
> > this ensures that we weight all tokens in the dataset equally,
> > rather than weighting each overall example equally when
> > using high amounts of gradient accumulation.
> > this can result in > 5 point improvements in AlpacaEval
> > see [huggingface/transformers#24725](https://github.com/huggingface/transformers/issues/24725) for
> > more discussion and details.
> 
> As it suggests a 5 point improvement when `reduce_loss="sum"` is used and our SFT run is with the current mean loss, with a 5% drop, do you think perhaps this might be the reason why?

Just a follow up on this:

I believe I am running into the same issue as https://github.com/allenai/open-instruct/issues/751

I also ran `git checkout cd0a3812 -- open_instruct/finetune.py` which has the `reduce_loss = "sum"` option and ran SFT again. This resulted in a similar loss curve (with the same scale) you posted [here](https://wandb.ai/ai2-llm/open_instruct_public/runs/mihc7mab)

What we got:
image


What you posted:
image

However, we still get `~46%` on GSM8K and IFEval. So I don't think it's due to `reduce_loss = "sum"`.

I can see in the wandb config you posted [here](https://wandb.ai/ai2-llm/open_instruct_public/runs/mihc7mab), SFT was trained with the base model `/gs/ai2-llm/post-training/deletable_cache_models/allenai/OLMo-2-0425-1B-v2/f967027b0baf304de96a73065faddeb81c078ec4`. Could you confirm, is this model the same as the public [allenai/OLMo-2-0425-1B](https://huggingface.co/allenai/OLMo-2-0425-1B) from hugging face? Or, would you be able to provide this base model so we can determine it is not an issue upstream?

hamishivi (approved):

PR #1337, opened 2026-01-12
New tools + utils

hamishivi:
Add the new tool base class and a single instantiation of it (pythoncodetool) to make it clear how it gets used. Also add related tests.

### DIFF
--- a/open_instruct/tools/test_new_tools.py
+++ b/open_instruct/tools/test_new_tools.py
@@ -1,6 +1,7 @@
 """Tests for new tools (PythonCodeTool from new_tools.py)."""
 
 import asyncio
+import dataclasses
 import unittest
 from unittest.mock import AsyncMock, MagicMock, patch
 
@@ -97,7 +98,6 @@ class TestPythonCodeToolExecution(unittest.IsolatedAsyncioTestCase):
 
     async def test_none_code_returns_error(self):
         """Test that None code returns an error."""
-        # This shouldn't happen in practice due to type hints, but test defensively
         result = await self.tool(None)
 
         self.assertIsInstance(result, ToolOutput)
@@ -106,7 +106,6 @@ class TestPythonCodeToolExecution(unittest.IsolatedAsyncioTestCase):
     @patch("open_instruct.tools.new_tools.aiohttp.ClientSession")
     async def test_successful_execution(self, mock_session_class):
         """Test successful code execution."""
-        # Set up mock response
         mock_response = AsyncMock()
         mock_response.json.return_value = {"output": "Hello, World!\n", "error": None}
         mock_response.raise_for_status = MagicMock()
@@ -128,7 +127,6 @@ class TestPythonCodeToolExecution(unittest.IsolatedAsyncioTestCase):
         self.assertFalse(result.timeout)
         self.assertGreater(result.runtime, 0)
 
-        # Verify API was called with correct parameters
         mock_session.post.assert_called_once_with(
             "http://localhost:1212/execute", json={"code": "print('Hello, World!')", "timeout": 3}
         )
@@ -237,13 +235,8 @@ class TestPythonCodeToolConfig(unittest.TestCase):
 
     def test_config_requires_api_endpoint(self):
         """Test config requires api_endpoint."""
-        # api_endpoint is now required, so we can't create a config without it
-        # This test verifies the field is required by the dataclass
-        import dataclasses
-
         fields = {f.name: f for f in dataclasses.fields(PythonCodeToolConfig)}
         self.assertIn("api_endpoint", fields)
-        # Check it has no default (required field)
         self.assertEqual(fields["api_endpoint"].default, dataclasses.MISSING)
         self.assertEqual(fields["api_endpoint"].default_factory, dataclasses.MISSING)
 


open_instruct/tools/utils.py
```diff
@@ -1,5 +1,9 @@
-from dataclasses import dataclass
-from typing import Any
+import inspect
+from abc import ABC, abstractmethod
+from dataclasses import asdict, dataclass
+from typing import Any, ClassVar, get_type_hints
+
+from pydantic import create_model
```
finbarrtimbers: can we do `import pydantic` as it's only used once?

finbarrtimbers: Also should we add `pydantic` into the main deps in `pyproject.toml`?

open_instruct/tools/utils.py
```diff
@@ -19,3 +23,151 @@ class ToolOutput:
...
+
+class Tool(ABC):
+    default_tool_name: str
+    """Default name used when calling this tool."""
+    default_description: str
+    """Default description used for e.g. system prompts."""
+    override_name: str | None = None
+    """Override name from default tool name. Useful if you want to swap different tools but use the same call name."""
+    _default_tool_parameters: dict[str, Any] | None = None
+    """If not set, auto-inferred from __call__ signature using pydantic."""
```
finbarrtimbers: How often are we going to be adding tools? Can we just require that people set this when adding a new tool?

open_instruct/tools/utils.py
```diff
@@ -19,3 +23,151 @@ class ToolOutput:
...
+    return schema
+
+
+class Tool(ABC):
+    default_tool_name: str
+    """Default name used when calling this tool."""
+    default_description: str
+    """Default description used for e.g. system prompts."""
+    override_name: str | None = None
+    """Override name from default tool name. Useful if you want to swap different tools but use the same call name."""
```
finbarrtimbers: Can you give me an example of why we would want to do this?

open_instruct/tools/utils.py
```diff
@@ -19,3 +23,151 @@ class ToolOutput:
...
+    # some cleaning up
+    schema.pop("title", None)
+    if "$defs" in schema and not schema["$defs"]:
+        schema.pop("$defs", None)
+
+    return schema
+
+
+class Tool(ABC):
+    default_tool_name: str
```
finbarrtimbers: Let's just call this (and `default_description`) `name`/`description`?

open_instruct/tools/utils.py
```diff
@@ -19,3 +23,151 @@ class ToolOutput:
...
+    default_description: str
+    """Default description used for e.g. system prompts."""
+    override_name: str | None = None
+    """Override name from default tool name. Useful if you want to swap different tools but use the same call name."""
+    _default_tool_parameters: dict[str, Any] | None = None
+    """If not set, auto-inferred from __call__ signature using pydantic."""
+
+    def __init_subclass__(cls, **kwargs: Any) -> None:
+        super().__init_subclass__(**kwargs)
+        if "default_tool_name" not in cls.__dict__:
```
finbarrtimbers: I think these are redundant, as it's impossible to initialize the class without passing in values, right?

open_instruct/tools/utils.py
```diff
@@ -19,3 +23,151 @@ class ToolOutput:
...
+            raise TypeError(f"{cls.__name__} must define 'default_description' class attribute")
+
+    @property
+    def tool_function_name(self) -> str:
+        if self.override_name is not None:
+            return self.override_name
+        return self.default_tool_name
+
+    @property
+    def tool_description(self) -> str:
```
finbarrtimbers: Why does this exist?

open_instruct/tools/utils.py
```diff
@@ -19,3 +23,151 @@ class ToolOutput:
...
+
+    def __init_subclass__(cls, **kwargs: Any) -> None:
+        super().__init_subclass__(**kwargs)
+        if "default_tool_name" not in cls.__dict__:
+            raise TypeError(f"{cls.__name__} must define 'default_tool_name' class attribute")
+        if "default_description" not in cls.__dict__:
+            raise TypeError(f"{cls.__name__} must define 'default_description' class attribute")
+
+    @property
+    def tool_function_name(self) -> str:
```
finbarrtimbers: Why does this exist? Let's just override default_tool_name?

open_instruct/tools/utils.py
```diff
@@ -19,3 +23,151 @@ class ToolOutput:
...
+        Otherwise, infers the schema from the `__call__` method's type hints and caches it.
+        """
+        cls = type(self)
+        if cls._default_tool_parameters is not None:
+            return cls._default_tool_parameters
+
+        cls._default_tool_parameters = infer_tool_parameters(self.__call__)
+        return cls._default_tool_parameters
+
+    def get_openai_tool_definitions(self) -> list[dict[str, Any]]:
```
finbarrtimbers: Let's make this a free function?

open_instruct/tools/utils.py
```diff
@@ -19,3 +23,151 @@ class ToolOutput:
...
+
+        Returns:
+            List of tool names.
+        """
+        return [self.tool_function_name]
+
+    @abstractmethod
+    async def __call__(self, *args: Any, **kwargs: Any) -> ToolOutput:
+        """Execute the tool, must be implemented by subclasses."""
+        pass
```
finbarrtimbers: raise NotImplementedError?

open_instruct/tools/utils.py
```diff
@@ -19,3 +23,151 @@ class ToolOutput:
...
+@dataclass
+class BaseToolConfig:
+    """Base configuration class for individual tools.
+
+    Subclasses must also define a config, which is used also used to instantiate the tool itself.
+    """
+
+    tool_class: ClassVar[type[Tool]]
+    """Related tool class for this config."""
+    override_name: str | None = None
```
finbarrtimbers: what about calling this "name"?

open_instruct/tools/new_tools.py
```diff
@@ -0,0 +1,114 @@
...
+    """Configuration for the Python code execution tool."""
+
+    tool_class: ClassVar[type[Tool]] = PythonCodeTool
+
+    api_endpoint: str | None = None
+    """The API endpoint for the code execution server."""
+    timeout: int = 3
+    """Timeout in seconds for code execution."""
+
+    def build(self) -> PythonCodeTool:
```
finbarrtimbers: What if you move the api_endpoint validation into PythonCodeTool.__init__, and then we can use the default `build` method?

open_instruct/tools/new_tools.py
```diff
@@ -0,0 +1,114 @@
...
+        return result
+
+
+@dataclass
+class PythonCodeToolConfig(BaseToolConfig):
+    """Configuration for the Python code execution tool."""
+
+    tool_class: ClassVar[type[Tool]] = PythonCodeTool
+
+    api_endpoint: str | None = None
```
finbarrtimbers: Why let this be None if that's invalid?

open_instruct/tools/new_tools.py
```diff
@@ -0,0 +1,114 @@
...
+                error = res.get("error") or ""
+
+                all_outputs.append(output)
+                if error:
+                    all_outputs.append("\n" + error)
+        except httpx.TimeoutException:
+            error = f"Timeout after {self.timeout} seconds"
+            all_outputs.append(error)
+            timed_out = True
+        except Exception as e:
```
finbarrtimbers: What does this do?

open_instruct/tools/new_tools.py
```diff
@@ -0,0 +1,114 @@
+"""
+Basic tools that are built-in to open-instruct.
+"""
+
+import time
+import traceback
+from dataclasses import asdict, dataclass
+from typing import Annotated, ClassVar
+
+import httpx
```
finbarrtimbers: We already use aiohttp in other places in `open-instruct`; can we standardize on that? (Or switch the uses of `aiohttp` to `httpx`?)

open_instruct/tools/new_tools.py
```diff
@@ -0,0 +1,114 @@
...
+from pydantic import Field
+
+from open_instruct.logger_utils import setup_logger
```
finbarrtimbers: As it's only called once, let's do 

from open_instruct import logger_utils?

open_instruct/tools/utils.py
hamishivi: It's decoupling the call name from the name of the tool itself (where the tool name is used in config for specifying what tools you want to run with, and the call name is the interface to the LM). I've called it override name to give the idea that by default we just reuse the config name as the call name, but sometimes you might want to change that.

So we might have different backends for the same functionality, and depending on cost/project/etc want to try different backends without changing the interface to the LM. For example, we could implement `search` with serper API (requires account, spending money) or with massiveDS (i.e., local datastore). We might want to train on massiveDS for a bit, and then swap later to serper for higher-quality but slower results. However, at other times you might want both serper and massiveds together.

open_instruct/tools/utils.py
hamishivi: I sort of wanted to make things standardised to json schema, and autobuilding from pydantic hints is a way to do this without asking the user to manually validate the parameters... but you're right, I can change this to make it specified, but then validate at init.

open_instruct/tools/utils.py
hamishivi: done!

open_instruct/tools/new_tools.py
hamishivi: Just a generic catchall for errors.... swapping to more specific error handling.

open_instruct/tools/utils.py
hamishivi: made the parameter annotation simpler, removing need for pydantic.

open_instruct/tools/utils.py
hamishivi: renamed to config_name / call_name

open_instruct/tools/utils.py
hamishivi: whoops removed. I think this was from some old stuff with proxying patterns (to be introduced in a future pr...)

open_instruct/tools/new_tools.py
```diff
@@ -0,0 +1,116 @@
...
+            async with aiohttp.ClientSession(timeout=timeout) as session:  # noqa: SIM117
+                async with session.post(self.api_endpoint, json={"code": code, "timeout": self.timeout}) as response:
+                    response.raise_for_status()
+                    res = await response.json()
+                    output = res.get("output", "")
+                    error = res.get("error") or ""
+
+                    all_outputs.append(output)
+                    if error:
+                        all_outputs.append("\n" + error)
```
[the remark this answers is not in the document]
hamishivi: this seems ugly to me...

PR #1340, opened 2026-01-12
Wire new tools into grpo_fast

hamishivi:
Wires the new tool handling into grpo-fast, along with a new way to specify the tools used. The setup is intentionally a bit generic to avoid hardcoding in new tools (e.g., asking the user to provide the tool config directly).

Also a small change to mason.py to allow the config dict to be passed through without issues.

Also edits the debug scripts to end-to-end test the new tool system!

### DIFF
--- a/open_instruct/tools/test_new_tools.py
+++ b/open_instruct/tools/test_new_tools.py
@@ -87,7 +87,7 @@ class TestPythonCodeToolInit:
 
 
 class TestPythonCodeToolExecution:
-    """Tests for PythonCodeTool execution (async __call__)."""
+    """Tests for PythonCodeTool execution (async execute)."""
 
     @pytest.fixture
     def tool(self):
--- a/open_instruct/tools/utils.py
+++ b/open_instruct/tools/utils.py
@@ -139,9 +139,13 @@ class Tool(ABC):
         return get_openai_tool_definitions(self)
 
     @abstractmethod
-    async def __call__(self, *args: Any, **kwargs: Any) -> ToolOutput:
+    async def execute(self, *args: Any, **kwargs: Any) -> ToolOutput:
         """Execute the tool, must be implemented by subclasses."""
-        raise NotImplementedError("__call__ must be implemented by subclasses.")
+        raise NotImplementedError("execute must be implemented by subclasses.")
+
+    async def __call__(self, *args: Any, **kwargs: Any) -> ToolOutput:
+        """Alias for execute, useful for inference scripts."""
+        return await self.execute(*args, **kwargs)
 
 
 @dataclass
--- a/open_instruct/vllm_utils.py
+++ b/open_instruct/vllm_utils.py
@@ -883,7 +883,7 @@ async def process_request(actor: LLMRayActor, sub_request_id: str, sampling_para
         # Execute tool calls
         outputs: list[str] = []
         for tool_call in tool_calls:
-            tool_result: ToolOutput = await actor.tool_actor_map[tool_call.name].__call__.remote(**tool_call.args)
+            tool_result: ToolOutput = await actor.tool_actor_map[tool_call.name].execute.remote(**tool_call.args)
 
             tool_called = True
             num_calls += 1
--- a/open_instruct/tools/new_tools.py
+++ b/open_instruct/tools/new_tools.py
@@ -52,7 +52,7 @@ class PythonCodeTool(Tool):
         self.api_endpoint = api_endpoint
         self.timeout = timeout
 
-    async def __call__(self, code: str) -> ToolOutput:
+    async def execute(self, code: str) -> ToolOutput:
         """Execute Python code via the API."""
         if not code or not code.strip():
             result = ToolOutput(


mason.py
```diff
@@ -470,7 +471,9 @@ def make_internal_command(command: list[str], args: argparse.Namespace, whoami:
                 filtered_command = maybe_download_tokenizer_from_gs_bucket(
                     filtered_command, args.auto_output_dir_path, whoami
                 )
-                caching_command = "python " + " ".join(filtered_command) + " --cache_dataset_only"
+                caching_command = (
+                    "python " + " ".join(shlex.quote(arg) for arg in filtered_command) + " --cache_dataset_only"
```
finbarrtimbers, line 477: Can you comment why we need shlex?

open_instruct/tools/test_new_tools.py
finbarrtimbers, line 1: Combine all of these into a parameterized test?

open_instruct/tools/utils.py
```diff
@@ -108,24 +108,40 @@ def _get_init_kwargs(self) -> dict[str, Any]:
...
+        Args:
+            call_name: Name used to identify in function calls. If not provided, uses tool's config name.
+
         Returns:
             A Tool instance.
         """
-        return self.tool_class(**self._get_init_kwargs())
-
-    def build_remote(self, max_concurrency: int = 512) -> ray.actor.ActorHandle:
+        args = self._get_init_kwargs()
```
finbarrtimbers: what about args = self._get_init_kwargs or {"call_name": call_name or self.tool_class.config_name}?

open_instruct/tools/utils.py
```diff
@@ -108,24 +108,40 @@ def _get_init_kwargs(self) -> dict[str, Any]:
...
 
         Args:
+            call_name: Name used to identify in function calls. If not provided, uses tool's config name.
             max_concurrency: Maximum number of concurrent calls the actor can handle.
 
         Returns:
             A Ray actor handle for the Tool.
         """
-        return ray.remote(self.tool_class).options(max_concurrency=max_concurrency).remote(**self._get_init_kwargs())
+        args = self._get_init_kwargs()
```
finbarrtimbers: Can we go through `build()` here to avoid duplicating the codepath?

open_instruct/data_loader.py
```diff
@@ -270,16 +270,15 @@ class StreamingDataLoaderConfig:
 
     # Tools
     tools: list[str] | None = None
-    max_tool_calls: tuple[int, ...] = (5,)
+    tool_call_names: list[str] | None = None
```
finbarrtimbers: Wdyt about pulling this stuff out into a `ToolsConfig`?

open_instruct/data_loader.py
```diff
@@ -316,13 +315,27 @@ def __post_init__(self):
         if self.stop_strings is None:
             self.stop_strings = []
 
-        self.max_tool_calls = tuple(int(x) for x in self.max_tool_calls)
+        self.max_tool_calls = int(self.max_tool_calls)
+
+        # Validate and set defaults for tool configuration
+        if self.tools:
```
finbarrtimbers: Can you pull this stuff out into a function?

open_instruct/data_loader.py
```diff
@@ -316,13 +315,27 @@ def __post_init__(self):
         if self.stop_strings is None:
             self.stop_strings = []
 
-        self.max_tool_calls = tuple(int(x) for x in self.max_tool_calls)
+        self.max_tool_calls = int(self.max_tool_calls)
```
finbarrtimbers: Is this still needed? We only needed it bc we were passing a tuple in earlier

open_instruct/grpo_fast.py
```diff
@@ -1500,42 +1502,88 @@ def setup_datasets(
...
+
+    Returns:
+        A list of Ray actor handles for the requested tools.
+
+    Raises:
+        ValueError: If an unknown tool is requested, configs are invalid, or required fields are missing.
+    """
+    tool_actors = []
+
+    if tools:
```
finbarrtimbers: Do we need this if? If tools is [], then the rest should be too, right? and then the loop won't execute?

open_instruct/grpo_fast.py
```diff
@@ -1500,42 +1502,88 @@ def setup_datasets(
...
+        for tool, call_name, config_str in zip(tools, tool_call_names, tool_configs):
+            tool_lower = tool.lower()
```
finbarrtimbers: Let's not do this, and then people have to use lower case in their scripts?

open_instruct/grpo_fast.py
```diff
@@ -1500,42 +1502,88 @@ def setup_datasets(
...
+
+            # Look up tool in registry
```
finbarrtimbers: Redundant!

open_instruct/grpo_fast.py
```diff
@@ -1500,42 +1502,88 @@ def setup_datasets(
...
+            if tool_lower not in TOOL_REGISTRY:
+                available_tools = ", ".join(TOOL_REGISTRY.keys())
+                raise ValueError(f"Unknown tool: {tool}. Available tools: {available_tools}")
+
+            tool_config_class = TOOL_REGISTRY[tool_lower]
+
+            # Parse config JSON
+            try:
+                config_dict = json.loads(config_str)
+            except json.JSONDecodeError as e:
```
finbarrtimbers: I think this is redundant? Don't you validate this string elsewhere?

open_instruct/grpo_fast.py
```diff
@@ -1500,42 +1502,88 @@ def setup_datasets(
...
+            if unknown_fields:
+                raise ValueError(
+                    f"Unknown config fields for tool '{tool}': {unknown_fields}. Expected fields: {expected_fields}"
+                )
+
+            # Build config from dictionary
+            try:
+                config = tool_config_class(**config_dict)
+            except TypeError as e:
+                raise ValueError(f"Invalid config for tool '{tool}': {e}. Expected fields: {expected_fields}") from e
```
finbarrtimbers: This entire section of code feels unnecessarily defensive to me. Do you really need all these try/excepts and the validation? What if we just let them fail like normal? How often do we expect these errors to happen?

open_instruct/grpo_fast.py
```diff
@@ -1500,42 +1502,88 @@ def setup_datasets(
...
+
+            # Build remote actor with custom call_name
+            tool_actor = config.build_remote(call_name=call_name, max_concurrency=512)
```
finbarrtimbers: Inline this, as it's only used once? And remove the comment, it's redundant?

open_instruct/grpo_fast.py
```diff
@@ -1555,41 +1603,17 @@ def create_model_and_optimizer(
     reward_config: RewardConfig,
     generation_config,
     data_prep_actor_state: dict | None = None,
-) -> tuple[ModelGroup, list[vllm_utils.LLMRayActor], dict[str, tools.Tool], int, int]:
+    tool_actors: list[ray.actor.ActorHandle] | None = None,
```
finbarrtimbers, line 1563: Let's make tool_actors always required, but an empty list is fine?

open_instruct/grpo_fast.py
```diff
@@ -2384,6 +2400,24 @@ def main(
         additive_format_reward=streaming_config.additive_format_reward,
         verifier_functions=build_all_verifiers(args, streaming_config),
     )
+
+    # Set up tool actors before creating generation configs
```
finbarrtimbers: Redundant!

open_instruct/grpo_fast.py
```diff
@@ -2384,6 +2400,24 @@ def main(
...
+    tool_actors = create_tools(
+        tools=streaming_config.tools,
+        tool_call_names=streaming_config.tool_call_names,
+        tool_configs=streaming_config.tool_configs,
+    )
+
+    # Create parser temporarily to get stop sequences for generation config
+    # The actual parser used during generation will be created inside vLLM actors
+    if tool_actors:
+        temp_parser = create_tool_parser(parser_type=streaming_config.tool_parser_type, tool_actors=tool_actors)
```
finbarrtimbers: Inline this, to make clear that it shouldn't be used?

open_instruct/grpo_fast.py
```diff
@@ -2384,6 +2400,24 @@ def main(
...
+        stop_seqs = temp_parser.stop_sequences()
+        logger.info(f"Adding tool stop sequences to config: {stop_seqs}")
+        streaming_config.stop_strings.extend(stop_seqs)
+
+    # Create generation configs AFTER adding tool stop sequences
```
finbarrtimbers: what about "AFTER maybe adding..."

open_instruct/vllm_utils.py
```diff
@@ -563,6 +539,20 @@ def _init_config(
         )
         self.reward_fn = reward_config.build() if reward_config else None
 
+        # Build mapping from tool names to actors for fast lookup
+        self.tool_actor_map = {}
+        if self.tool_actors:
+            call_name_futures = [actor.get_call_name.remote() for actor in self.tool_actors]
```
finbarrtimbers: inline this, as it's only used once?

open_instruct/vllm_utils.py
```diff
@@ -563,6 +539,20 @@ def _init_config(
...
+            call_names = ray.get(call_name_futures)
+            self.tool_actor_map = dict(zip(call_names, self.tool_actors))
+
+        # Create tool parser inside the actor (avoids serialization issues)
+        self.tool_parser = None
+        if self.tool_actors:
+            from open_instruct.grpo_fast import create_tool_parser
```
finbarrtimbers: It's a bit broken conceptually to have vllm_utils import from grpo_fast. Can we put create_tool_parser somewhere else? Maybe in tools/__init__.py or tools/utils.py?

open_instruct/vllm_utils.py
```diff
@@ -838,30 +828,41 @@ async def process_request(actor: LLMRayActor, sub_request_id: str, sampling_para
...
-        loop = asyncio.get_running_loop()
-        tool_result = await loop.run_in_executor(actor.executor, triggered_tool, output.text)
-
-        tool_called = True
-        num_calls += 1
-        timeout = timeout or tool_result.timeout
-        tool_error += "" if tool_result.error is None else tool_result.error
-        tool_output += tool_result.output
-        tool_runtime += tool_result.runtime
+        # Process each tool call
```
finbarrtimbers: Can we make this into a free function?

open_instruct/tools/utils.py
hamishivi: We can't really go through build because the ray remote instantion requires wrapping the class before calling init, and build() returns the new object directly. I just removed build() since I was mainly using it for testing.

open_instruct/data_loader.py
hamishivi: Was planning on doing this in a future PR! For now, the tool setup is pretty light weight, since we only have a simple parser and a simple tool.

finbarrtimbers: Can you add a # TODO(hamishivi): Pull this into a ToolsCOnfig? Otherwise that sgtm!

open_instruct/tools/utils.py
finbarrtimbers: SG!

mason.py
hamishivi: Making the config a dict string that you pass in, the previous setup would munge the strings no matter what i did, s.t. e.g.
```bash
    --tool_configs '{"api_endpoint": "https://open-instruct-tool-server-10554368204.us-central1.run.app/execute", "timeout": 3}' \
```
would not work. The shlex.quote function handles escaping things properly and fixes that.

open_instruct/data_loader.py
hamishivi: I just ended up implementing the toolsconfig to make some other comments easy to address!

open_instruct/data_loader.py
hamishivi: I'm just simplifying it for now (we take in a single int as a shared max call count), and I'll add in the actual max calls exceeded handling back in after this PR lands. Then, once we have some more tools, I miiight add the per-tool max call handling.

open_instruct/grpo_fast.py
hamishivi: Actually we didnt, but I moved the validation into the ToolsConfig post_init, and removed this.

open_instruct/grpo_fast.py
hamishivi: I simplified this a bit... this validation, yes, because the user is specifying some config dict that we need to actually instantiate the tool to check. I want to explicitly control the error to be thrown to tell the user *which* config is the problem.

open_instruct/grpo_fast.py
hamishivi: I sort of don't like that since it means the API is such that, even if you don't have tools, you have to explicitly pass in an empty list? We shouldn't make an empty list a default arg because of how python works with mutable default args (https://docs.astral.sh/ruff/rules/mutable-argument-default/).

open_instruct/vllm_utils.py
hamishivi: oop, yeah, missed this import... Moved create_tool_parser to parsers.py.

open_instruct/vllm_utils.py
hamishivi: Restructured into functions a bit!

open_instruct/grpo_fast.py
finbarrtimbers: ah yeah you're right, you're right

mason.py
```diff
@@ -470,7 +471,11 @@ def make_internal_command(command: list[str], args: argparse.Namespace, whoami:
                 filtered_command = maybe_download_tokenizer_from_gs_bucket(
                     filtered_command, args.auto_output_dir_path, whoami
                 )
-                caching_command = "python " + " ".join(filtered_command) + " --cache_dataset_only"
+                # we use shlex.quote to ensure that args with special characters are properly quoted
+                # this is important for tool config, which is a json dict.
```
finbarrtimbers: nit: `tool_configs`, right?

open_instruct/vllm_utils.py
finbarrtimbers: Why not add a "process_tools" function?

open_instruct/vllm_utils.py
```diff
@@ -838,30 +828,41 @@ async def process_request(actor: LLMRayActor, sub_request_id: str, sampling_para
...
+        tool_outputs_list: list[str] = []
+        for tool_call in tool_calls:
+            tool_actor = actor.tool_actor_map.get(tool_call.name)
+            if not tool_actor:
+                logger.warning(f"Tool {tool_call.name} called but not found in actor map")
+                continue
```
finbarrtimbers: Does 

```
tool_result: ToolOutput = await tool_actor.remote(**tool_call.args)
```

work?

open_instruct/grpo_fast.py
```diff
@@ -1414,7 +1418,7 @@ def setup_runtime_variables(args: Args, streaming_config: data_loader_lib.Stream
         args.hf_repo_url = f"https://huggingface.co/{args.hf_repo_id}/tree/{args.hf_repo_revision}"
     if args.with_tracking and args.wandb_entity is None:
         args.wandb_entity = maybe_use_ai2_wandb_entity()
-    args.tool_use = streaming_config.tools is not None and len(streaming_config.tools) > 0
+    args.tool_use = tools_config.enabled
```
finbarrtimbers: Can we just directly reference tools_config later? I'd rather not have multiple sources of truth for the same variable

open_instruct/grpo_fast.py
```diff
@@ -1500,42 +1504,46 @@ def setup_datasets(
...
+        # Build config from dictionary
+        try:
+            config = tool_config_class(**config)
+        except Exception as e:
+            raise ValueError(f"Invalid config for tool '{tool}': {e}") from e
 
-    return tool_objects
+        # The config is a dataclass, and may have performed additional validation
+        # or transformation of the args, which we then now pass to the tool.
+        kwarg_dict = asdict(config)
```
finbarrtimbers: kwarg_dict = asdict(config) | {"call_name": call_name}?

open_instruct/grpo_fast.py
```diff
@@ -1500,42 +1504,46 @@ def setup_datasets(
...
+        kwarg_dict["call_name"] = call_name
+        tool_actor = ray.remote(tool_config_class.tool_class).options(max_concurrency=512).remote(**kwarg_dict)
```
finbarrtimbers: Inline this, as it's only used once?

open_instruct/vllm_utils.py
hamishivi: Unfortunately not, from a quick manual test.

open_instruct/grpo_fast.py
hamishivi: Oh, good point. I don't even think this was used.

open_instruct/grpo_fast.py
hamishivi: fancy!

open_instruct/vllm_utils.py
hamishivi: I rejigged this a bit again. The main reason before was that the parser just checks for syntactically valid tool calls, but the model might hallucinate tool calls that are formatting correctly but otherwise wrong, in which case, we filter them out. I restructured a bit to split out this filtering from the execution, so now the flow is like:
parse out calls -> filter calls -> execute calls -> process outputs (inside a free function).

I think if we tried to push the full loop into a function, we end up having to re-declare and then return out a bunch of variables that are accumulated over the full loop, such as tool_runtime or num_calls, which is a bit uglier to me?

open_instruct/vllm_utils.py
```diff
@@ -838,42 +868,45 @@ async def process_request(actor: LLMRayActor, sub_request_id: str, sampling_para
...
-        if not actor.tools or not actor.max_tool_calls:
+        # check if we have tools to check for
+        if not actor.tool_actors or actor.tool_parser is None:
             break
 
-        triggered_tool, stop_str = get_triggered_tool(
-            output.text, actor.tools, actor.max_tool_calls, num_calls, sampling_params
-        )
-        if triggered_tool is None:
+        tool_calls = actor.tool_parser.get_tool_calls(output.text)
```
finbarrtimbers, line 875: Why not put all of this (down to the call to `process_tool_tokens`, including the for loop) into `process_tool_tokens`?

open_instruct/vllm_utils.py
```diff
@@ -838,42 +868,45 @@ async def process_request(actor: LLMRayActor, sub_request_id: str, sampling_para
...
-        tool_tokens = actor.llm_engine.tokenizer.encode(
-            "<output>\n" + tool_result.output + "</output>\n", add_special_tokens=False
-        )
-
-        tool_tokens, excess = truncate_tool_output_tokens(
-            tool_tokens,
+        # Execute tool calls
+        outputs: list[str] = []
+        for tool_call in tool_calls:
+            tool_result: ToolOutput = await actor.tool_actor_map[tool_call.name].__call__.remote(**tool_call.args)
```
finbarrtimbers: Can we rename `__call__` then? I don't like calling methods with underscores! Maybe `call` or `run`?

open_instruct/vllm_utils.py
finbarrtimbers: Can you rename `__call__` `call` then? (or whatever you want: `get`, `run`, etc.). I don't like referencing names that start with underscores as they're generally intended to be internal implementation details.

hamishivi: Sure... I guess conceptually I sort of see the tools as essentially functions with some state, and then from a higher level `__call__` seems appropriate since its the main reason for the tool to exist. If we didn't have the ray stuff, I feel that `Search(query=...)` is cleaner than `Search.run(query=...)`.

open_instruct/vllm_utils.py
hamishivi: I sort of like the current separation of execution vs parsing, which also matches the tool vs parser distinction.

open_instruct/vllm_utils.py
hamishivi: done. Added execute and then aliased `__call__` to it so the old interface still exists for non-ray scenarios.

PR #1345, opened 2026-01-13
Add serper search

hamishivi:
Adds a serper search tool (google search).
Also adds a utility function for making API calls, since this is a common pattern for tools.

### DIFF
--- a/open_instruct/tools/new_tools.py
+++ b/open_instruct/tools/new_tools.py
@@ -77,8 +77,8 @@ class PythonCodeTool(Tool):
                 runtime=time.time() - start_time,
             )
         else:
-            output = api_response.data.get("output", "")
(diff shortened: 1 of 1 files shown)


open_instruct/tools/new_tools.py
```diff
@@ -66,40 +64,26 @@ async def execute(self, code: str) -> ToolOutput:
...
+            result = ToolOutput(
+                output=api_response.error,
+                called=True,
+                error=api_response.error,
+                timeout=api_response.timed_out,
+                runtime=time.time() - start_time,
+            )
+        else:
+            output = api_response.data.get("output", "")
+            error = api_response.data.get("error") or ""
```
finbarrtimbers: Use the same pattern for both error and output?

PR #1347, opened 2026-01-13
Add s2 search tool

hamishivi:
Adds a tool for the S2 snippet search API. Minor change to get api utility.

### DIFF
--- a/open_instruct/tools/utils.py
+++ b/open_instruct/tools/utils.py
@@ -117,11 +117,13 @@ async def make_api_request(
     """
     if method not in ["GET", "POST"]:
         raise ValueError(f"Invalid method: {method}")
+    if method == "GET" and json_payload:
+        raise ValueError("JSON payload cannot be provided for GET requests")
     try:
         timeout = aiohttp.ClientTimeout(total=timeout_seconds)
         async with aiohttp.ClientSession(timeout=timeout) as session:
             if method == "GET":
-                request_context = session.get(url, params=params, json=json_payload, headers=headers)
+                request_context = session.get(url, params=params, headers=headers)
             else:
                 request_context = session.post(url, params=params, json=json_payload, headers=headers)
 


open_instruct/tools/new_tools.py
```diff
@@ -189,6 +188,82 @@ class JinaBrowseToolConfig(BaseToolConfig):
...
+
+    config_name = "s2_search"
+    description = "Searches Semantic Scholar for academic papers and citations"
+    parameters = {
+        "type": "object",
+        "properties": {"query": {"type": "string", "description": "The search query for Semantic Scholar"}},
+        "required": ["query"],
+    }
+
+    def __init__(self, call_name: str, num_results: int = 10, timeout: int = 60) -> None:
```
finbarrtimbers, line 206: Given that we've got the exact same __init__ for all the search tools, maybe it's worth making a `SearchTool` class that they all implement? Up to you. Unclear how much can be shared.

open_instruct/tools/new_tools.py
```diff
@@ -189,6 +188,82 @@ class JinaBrowseToolConfig(BaseToolConfig):
...
+
+        error = "" if snippets else "Query returned no results."
+        output = "\n".join(snippets).strip() if snippets else error
+
+        result = ToolOutput(output=output, called=True, error=error, timeout=False, runtime=time.time() - start_time)
+        _log_tool_call(self.call_name, query, result)
+        return result
+
+
+@dataclass
```
finbarrtimbers, line 255: Same comment here-- maybe we need a single `SearchToolConfig` class, where the only difference is the `tool_class`?

open_instruct/tools/new_tools.py
hamishivi: For now, I think I'll keep them separate since this is the last of the search tools I want to implement. Each tool does do slightly different parsing of the API results, but yeah I think some stuff can be shared. I'll consider it if I decide to do some extra work on it (e.g., updating logging or similar).

PR #1351, opened 2026-01-13
HFDataloader refactoring in preparation for DPO.

finbarrtimbers:
This is part of a series of PRs that are implementing the new Olmo-core based DPO.

In this PR, we refactor `HFDataloader`:

- Add to_device() function for moving batch tensors to a device
- Add collator parameter for custom batch collation
- Add device parameter for automatic tensor placement
- Fix fs_local_rank from hardcoded 0 to actual rank (critical for multi-node)
- Fix sharding logic to reshard dataset per epoch
- Improve docstrings explaining 'index' column requirement
- Fix collator for single-example iteration (batch_size=1) in DataPreparationActor and eval_data_loader

### Sharding refactor

Removed `_index_to_position` mapping and restructured `effective_size`:

**Old approach**: Shard first, shuffle second, maintain a mapping from original index → shuffled position. This was needed because `__getitem__` had to look up examples by their original index after shuffling.

**New approach**: Use `dataset.select(indices)` which creates a view where `dataset[i]` returns the example at `indices[i]` in the original dataset. The `index` field is preserved in each example's data, so no mapping needed.

`effective_size` moved into `_reshard()` because:
1. It's recomputed each epoch as excluded indices change
2. The new sharding shuffles globally first, then distributes batches round-robin to ranks - ensuring all ranks get equal work (the old shard-then-shuffle approach could give unequal sizes per rank)

Runs: 
1. Single GPU GRPO: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/open-instruct-dev/work/01KEWTKV4XM9G5B2H5QD60D4GX?taskId=01KEWTKV53CJY75KMS5M45S0X9&jobId=01KEWTKV8MPY4HF8QT677B7XZT)
2. Single GPU DPO: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/open-instruct-dev/work/01KEWSCB6MHXW90KC496RA6861?taskId=01KEWSCB6W7K10SMECT25R715V&jobId=01KEWSCBAHSV6X1SBFM4T5DBH0)
3. Multi-node GRPO: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/open-instruct-dev/work/01KEYW45C2DPMNP5QAGXH3XR6G?taskId=01KEYW45C8RHWJW4SXP740G3FK&jobId=01KEYW45R0QVVZQSK7JR84AKAA)
4. Multi-node DPO: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/open-instruct-dev/work/01KEYXDXH305FJ7KPMKX0TJ26T?taskId=01KEYXDXHA07EVCY37WYR9WN70&jobId=01KEYXDXNAAF2WJ7X5FTF1Z738)


open_instruct/data_loader.py
```diff
@@ -68,32 +78,54 @@ def __init__(
...
+        self._full_dataset = dataset
         self.seed = seed
         self._batch_size = batch_size
-        self.effective_size = len(self.dataset) - (len(self.dataset) % batch_size)
+        if batch_size % world_size != 0:
+            logger.warning(
+                f"Global batch size {batch_size} is not divisible by world size {world_size}. "
+                f"The effective global batch size will be {batch_size // world_size * world_size}."
+            )
+        self._per_rank_batch_size = batch_size // world_size
```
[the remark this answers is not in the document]
finbarrtimbers: Done! Thanks.

open_instruct/data_loader.py
```diff
@@ -68,32 +78,59 @@ def __init__(
...
     ) -> None:
         """Initialize the HFDataLoader.
 
         Args:
-            dataset: The HuggingFace Dataset to load data from.
+            dataset: The HuggingFace Dataset to load data from. Must have an 'index' column.
             batch_size: The global batch size.
             seed: Random seed for shuffling.
             rank: The rank of the current process in the distributed setup.
             world_size: Total number of processes in the distributed setup.
```
hamishivi:
```suggestion
            world_size: Total number of data-parallel processes in the distributed setup.
```

open_instruct/data_loader.py
```diff
@@ -154,30 +192,68 @@ def exclude_index(self, index: int) -> None:
...
+        rng.shuffle(all_indices)
+
+        global_size = len(all_indices)
+        total_batches = global_size // self._batch_size
+        usable_size = total_batches * self._batch_size
+
+        # Distribute examples from global batches to ranks. This is a form of strided sampling where each
+        # rank gets a subset of examples from each global batch, ensuring a diverse set of examples.
+        rank_indices = all_indices[:usable_size].reshape(total_batches, self._batch_size)
+        rank_indices = rank_indices[:, self.dp_rank :: self.dp_world_size].flatten()
```
hamishivi, line 228: nit: Can we be more consistent in using dp_world_size vs world_size? Since in __init__ we use world_size and rank. I don't mind one or the other (slight pref to dp_* to make the sharding clear).

open_instruct/data_loader.py
```diff
@@ -154,30 +192,68 @@ def exclude_index(self, index: int) -> None:
...
+        """Return the total number of tokens in the batch across all ranks.
+
+        Counts tokens from all keys containing 'input_ids' that are torch tensors.
+
+        Args:
+            batch: A batch dictionary containing input tensors.
 
         Returns:
-            The first item from the dataset.
+            Total number of tokens across all ranks, or None if no input_ids found.
```
hamishivi: Should we allow returning none, or just error if the user tries to run this function without there being an input_ids key?

finbarrtimbers: Let's do an error tbh! Much better.

finbarrtimbers: Ok done!

open_instruct/data_loader.py
finbarrtimbers: Done! Great suggestion.

PR #1353, opened 2026-01-14
Fix wandb_tracker AttributeError in dpo_tune_cache.py

finbarrtimbers:
We had a silly situation where we'd have to set two flags in `dpo_tune_cache.py` to get wandb logging: `--with_tracking` and `--report_to`. In this PR, we always set the proper values to report to Wandb when `--with_tracking` is set. 

We would get the error: 

```
2026-01-14T16:54:02.191Z [rank0]: Traceback (most recent call last):
2026-01-14T16:54:02.191Z [rank0]:   File "/stage/open_instruct/dpo_tune_cache.py", line 1226, in <module>
2026-01-14T16:54:02.191Z [rank0]:     main(args, tc)
2026-01-14T16:54:02.191Z [rank0]:   File "/stage/open_instruct/dpo_tune_cache.py", line 646, in main
2026-01-14T16:54:02.191Z [rank0]:     maybe_update_beaker_description(wandb_url=wandb_tracker.run.get_url())
2026-01-14T16:54:02.191Z [rank0]:                                               ^^^^^^^^^^^^^^^^^
2026-01-14T16:54:02.191Z [rank0]: AttributeError: 'GeneralTracker' object has no attribute 'run'
```

I ran `medium_dpo.sh` to verify: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/open-instruct-dev/work/01KEYS2P9T5G03FFDEDQ9YEWEM?).

### DIFF
--- a/configs/train_configs/tulu3/tulu3_dpo_8b.yaml
+++ b/configs/train_configs/tulu3/tulu3_dpo_8b.yaml
@@ -17,7 +17,6 @@ weight_decay: 0.0
 num_train_epochs: 1
 output_dir: output/dpo_8b
 with_tracking: true
-  - wandb
 logging_steps: 1
 use_lora: false
 dpo_loss_type: dpo_norm
--- a/configs/train_configs/dpo/default.yaml
+++ b/configs/train_configs/dpo/default.yaml
@@ -16,7 +16,6 @@ weight_decay: 0.0
 num_train_epochs: 3
 output_dir: output/dpo_7b_recreate2/
 with_tracking: true
-  - wandb
 logging_steps: 1
 checkpointing_steps: 1000
 keep_last_n_checkpoints: 3
--- a/configs/train_configs/dpo/final_best_70b_dpo_config.yaml
+++ b/configs/train_configs/dpo/final_best_70b_dpo_config.yaml
@@ -23,7 +23,6 @@ weight_decay: 0.0
 num_train_epochs: 1
 output_dir: /output
 with_tracking: true
-  - wandb
 logging_steps: 1
 use_lora: false
 dpo_loss_type: dpo_norm
--- a/configs/train_configs/dpo/final_best_8b_dpo_config.yaml
+++ b/configs/train_configs/dpo/final_best_8b_dpo_config.yaml
@@ -22,7 +22,6 @@ weight_decay: 0.0
 num_train_epochs: 1
 output_dir: /output
 with_tracking: true
-  - wandb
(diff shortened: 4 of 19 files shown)


scripts/train/debug/large_dpo.sh
```diff
@@ -14,6 +14,7 @@ uv run python mason.py \
     --preemptible \
     --num_nodes 4 \
     --budget ai2/oe-adapt \
+    --no_auto_dataset_cache \
```
hamishivi, line 17: Why this change? I think we should keep the caching in the larger debug runs?

hamishivi (changes requested):
some minor fixes

configs/train_configs/tulu3/tulu3_dpo_8b.yaml
```diff
@@ -17,7 +17,6 @@ weight_decay: 0.0
 num_train_epochs: 1
 output_dir: output/dpo_8b
 with_tracking: true
-report_to:
   - wandb
```
hamishivi: Don't you also need to remove this line? Likewise for all the above yamls.

scripts/train/debug/large_dpo.sh
finbarrtimbers: It doesn't run on Mac with caching, unfortunately, because of the vllm dep changes.

configs/train_configs/tulu3/tulu3_dpo_8b.yaml
finbarrtimbers: ah yeah ugh

scripts/train/debug/large_dpo.sh
hamishivi: It would be good to test the code path by default though...

finbarrtimbers: I think that'll involve pulling the caching code out of the main script so we can call it without importing everything. I've got a working PR: https://github.com/allenai/open-instruct/pull/1267.

configs/train_configs/tulu3/tulu3_dpo_8b.yaml
finbarrtimbers: Ok these have been fixed!

scripts/train/debug/large_dpo.sh
hamishivi: Yeah, makes sense!

PR #1355, opened 2026-01-14
Improve DPO loss functions with better type hints and docstrings

finbarrtimbers:
Now, we type check `dpo_utils.py`. Also adds tests for `dpo_utils.py`. 

Also adds a `concatenated_forward` implementation for Olmo-core models.

### DIFF
--- a/open_instruct/test_grpo_fast_gpu.py
+++ b/open_instruct/test_grpo_fast_gpu.py
@@ -133,7 +133,7 @@ class TestGeneration(TestGrpoFastBase):
         tokenizer_name = "Qwen/Qwen3-1.7B"
         tool_actors = (
             create_tools(
-                tools=["python"], tool_call_names=["</code>"], tool_configs=[{"api_endpoint": self.tool_api_endpoint}]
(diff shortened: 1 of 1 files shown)


open_instruct/dpo_utils.py
```diff
@@ -78,15 +78,33 @@ def dpo_loss(
...
+    policy_rejected_logps: torch.Tensor,
+    reference_chosen_logps: torch.Tensor,
+    reference_rejected_logps: torch.Tensor,
     beta: float,
     label_smoothing: float = 0.0,
     chosen_loss_mask: torch.BoolTensor = None,
     rejected_loss_mask: torch.BoolTensor = None,
-) -> tuple[torch.FloatTensor, torch.FloatTensor, torch.FloatTensor]:
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    """Compute the Weighted Preference Optimization (WPO) loss.
```
natolambert, line 90:
```suggestion
    """Compute the Weighted Preference Optimization (WPO) loss.
    Paper: https://arxiv.org/abs/2406.11827
```

PR #1358, opened 2026-01-14
Add OLMo-Core DPO training implementation

finbarrtimbers:
Implement DPO training using OLMo-core's native Trainer framework:
- Uses olmo_core.train.TrainModule as the base class
- HSDP (Hybrid Sharded Data Parallel) support for distributed training
- BeakerCallbackV2 integration for Beaker v2 compatibility
- Reference model logprob caching via TensorCache
- Multi-node distributed training support

Files:
- open_instruct/dpo.py: Main DPO training implementation
- open_instruct/test_dpo.py: Test suite for DPO training
- scripts/train/debug/dpo.sh: Single-node debug script
- scripts/train/debug/large_dpo.sh: Multi-node debug script
- mason.py: Add dpo.py to OPEN_INSTRUCT_COMMANDS, OLMO_CORE_JOBS

Note: This PR includes changes from dependent PRs (data_loader, dpo_config, dpo_utils) to ensure the branch is self-contained. After those PRs merge, this branch should be rebased.

### DIFF
--- a/open_instruct/dpo.py
+++ b/open_instruct/dpo.py
@@ -11,7 +11,7 @@ import pathlib
 import shutil
 import time
 from functools import partial
-from typing import Any, cast
+from typing import Any
 
 import peft
 import torch
@@ -542,8 +542,12 @@ def main(args: dpo_utils.ExperimentConfig, tc: dataset_transformation.TokenizerC
 
 
 if __name__ == "__main__":
-    parser = utils.ArgumentParserPlus([dpo_utils.ExperimentConfig, dataset_transformation.TokenizerConfig])
-    args, tc = cast(
-        tuple[dpo_utils.ExperimentConfig, dataset_transformation.TokenizerConfig], parser.parse_args_into_dataclasses()
-    )
+    import simple_parsing
+
+    parser = simple_parsing.ArgumentParser()
+    parser.add_arguments(dpo_utils.ExperimentConfig, dest="args")
+    parser.add_arguments(dataset_transformation.TokenizerConfig, dest="tokenizer_config")
+    ns = parser.parse_args()
+    args: dpo_utils.ExperimentConfig = ns.args
+    tc: dataset_transformation.TokenizerConfig = ns.tokenizer_config
     main(args, tc)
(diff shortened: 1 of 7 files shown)


open_instruct/dpo.py
```diff
@@ -0,0 +1,553 @@
...
+)
+
+from open_instruct import data_loader, dataset_transformation, dpo_utils, logger_utils, model_utils, utils
+from open_instruct.beaker_callback import BeakerCallbackV2
+from open_instruct.padding_free_collator import TensorDataCollatorWithFlatteningDPO
+
+logger = logger_utils.setup_logger(__name__)
+
+
+OLMO_MODEL_CONFIG_MAP: dict[str, str] = {
```
natolambert, line 46: Can we add a comment on where this list is maintained in olmo core so people can cross reference?
I'm sure no one loves this list approach

natolambert (approved):
Overall this looks okay to me, I may get to look more.
We shouldn't merge it until we have evals that look solid on a model trained with it. We know this comes with some work on finding new hyperparameters (maybe) and DPO has been sensitive to the LR.

**Feature suggestion: Support loading native OLMo-core checkpoints**
I *think* we convert the models to hf to evaluate, but could save conversion time if we have the sharded path lying around.


Currently the implementation only supports loading HuggingFace models via `load_hf_model()`:

```python
model_config = get_transformer_config(args.model.model_name_or_path, vocab_size)
model = model_config.build(init_device="cpu")
load_hf_model(args.model.model_name_or_path, model.state_dict(), work_dir=args.checkpoint.output_dir)
```

If we're doing SFT with OLMo-core, it would be faster to load the native sharded checkpoint directly into DPO rather than requiring conversion to HF format first.

**Suggested approach:**

1. Add a checkpoint format option:
```python
# In ModelConfig
checkpoint_format: Literal["hf", "olmo_core"] | None = None
"""Checkpoint format. If None, auto-detect based on directory structure."""
```

2. Auto-detect or branch on loading:
```python
if checkpoint_format == "olmo_core" or is_olmo_core_checkpoint(model_name_or_path):
    # Load native OLMo-core checkpoint (sharded, distributed-aware)
    trainer.load_checkpoint(model_name_or_path)  # or equivalent
else:
    # Existing HF path
    load_hf_model(model_name_or_path, model.state_dict(), ...)
```

This would enable a cleaner SFT → DPO pipeline when both stages use OLMo-core, avoiding the intermediate HF conversion step.

Not a blocker for this PR, but worth considering as a follow-up.

finbarrtimbers:
> Overall this looks okay to me, I may get to look more.
> We shouldn't merge it until we have evals that look solid on a model trained with it. We know this comes with some work on finding new hyperparameters (maybe) and DPO has been sensitive to the LR.

I think it's actually fine to merge, because it's a new implementation: `dpo.py`. We'll still have the existing one around. Wdyt?

PR #1362, opened 2026-01-15
Add vllm parser

hamishivi:
Adds a vllm-compatible parser! This allows us to re-use vllm's tool parsing logic instead of rolling our own.
It also adds tool definition plumbing into the dataset construction. This is needed since some models (qwen, olmo) can auto-populate their system prompt based on these tool definitions, and are trained to use these.

Note that right now it uses this dummy request from `_make_request`, which is required to get vllm to see the tool definitions. Going further than this would require wiring the messages list through our setup, which is tricky, as then we would also need to validate that the two are consistent with each other (...this is a broader problem that I would prefer to tackle in a future PR).

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -2239,10 +2239,7 @@ def run_training(
     save_final_model(args, policy_group, tokenizer, training_step, wandb_url, tc.chat_template_name)
 
 
-def initialize_tools(
-    tools_config: ToolsConfig,
-    tokenizer,
-) -> tuple[list, list, list[str]]:
+def initialize_tools(tools_config: ToolsConfig, tokenizer) -> tuple[list, list, list[str]]:
     """Initialize tool actors and get tool definitions and stop sequences.
 
     Args:


open_instruct/tools/parsers.py
```diff
@@ -102,29 +114,198 @@ def stop_sequences(self) -> list[str]:
...
-        return OpenInstructLegacyToolParser(tool_actors, output_wrap_name=output_wrap_name)
-    else:
-        # TODO: Implement other parser types (vllm_hermes, vllm_llama3_json, etc.)
-        raise ValueError(f"Parser type '{parser_type}' not yet implemented. Use 'legacy' for now.")
+        return OpenInstructLegacyToolParser(tool_actors, output_wrap_name="output")
+
+    if parser_type in VLLM_PARSERS:
+        return create_vllm_parser(parser_type, tokenizer, tool_definitions=tool_definitions)
+
+    raise ValueError(f"Unknown parser type: '{parser_type}'. Available: {get_available_parsers()}")
```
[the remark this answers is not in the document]
hamishivi: Nah, I like forcing the tokenizer to be set. We want to eventually sunset the legacy parser anyway.

open_instruct/tools/tests/test_parsers.py
```diff
@@ -246,6 +253,103 @@ def test_contains_legacy(self):
...
+        self.assertIn("vllm_llama3_json", parsers)
+        self.assertIn("vllm_olmo3", parsers)
+
+
+class TestVllmParserRegistry(unittest.TestCase):
+    """Tests for vLLM parser registry and helpers."""
+
+    def test_vllm_parsers_have_required_fields(self):
+        """Test that all registered vLLM parsers have required configuration."""
+        for name, config in VLLM_PARSERS.items():
```
finbarrtimbers: Let's use parameterized here and move the for loop outside of the test!

open_instruct/tools/tests/test_parsers.py
```diff
@@ -246,6 +253,103 @@ def test_contains_legacy(self):
...
+            self.assertIsInstance(config, VllmParserConfig, f"Parser {name} config is not VllmParserConfig")
+            self.assertTrue(config.import_path, f"Parser {name} missing import_path")
+            self.assertTrue(config.output_template, f"Parser {name} missing output_template")
+            self.assertIn("{}", config.output_template, f"Parser {name} output_template missing placeholder")
+            self.assertIsInstance(config.stop_sequences, list, f"Parser {name} stop_sequences is not a list")
```
finbarrtimbers: Let's isntead check that it has a non-zero length? We could, for instance, change this to be a set (or any iterable!) and I think the test should still pass conceptually.

open_instruct/tools/tests/test_parsers.py
```diff
@@ -246,6 +253,103 @@ def test_contains_legacy(self):
...
+class TestVllmParserRegistry(unittest.TestCase):
+    """Tests for vLLM parser registry and helpers."""
+
+    def test_vllm_parsers_have_required_fields(self):
+        """Test that all registered vLLM parsers have required configuration."""
+        for name, config in VLLM_PARSERS.items():
+            self.assertIsInstance(config, VllmParserConfig, f"Parser {name} config is not VllmParserConfig")
+            self.assertTrue(config.import_path, f"Parser {name} missing import_path")
+            self.assertTrue(config.output_template, f"Parser {name} missing output_template")
+            self.assertIn("{}", config.output_template, f"Parser {name} output_template missing placeholder")
```
finbarrtimbers: Can we instead try using the output_template?

open_instruct/tools/tests/test_parsers.py
```diff
@@ -246,6 +253,103 @@ def test_contains_legacy(self):
...
+
+
+class TestVllmParserRegistry(unittest.TestCase):
+    """Tests for vLLM parser registry and helpers."""
+
+    def test_vllm_parsers_have_required_fields(self):
+        """Test that all registered vLLM parsers have required configuration."""
+        for name, config in VLLM_PARSERS.items():
+            self.assertIsInstance(config, VllmParserConfig, f"Parser {name} config is not VllmParserConfig")
+            self.assertTrue(config.import_path, f"Parser {name} missing import_path")
```
finbarrtimbers: Can we instead try using the import_path? Or loading it?

open_instruct/tools/tests/test_parsers.py
```diff
@@ -246,6 +253,103 @@ def test_contains_legacy(self):
...
+            self.assertTrue(config.output_template, f"Parser {name} missing output_template")
+            self.assertIn("{}", config.output_template, f"Parser {name} output_template missing placeholder")
+            self.assertIsInstance(config.stop_sequences, list, f"Parser {name} stop_sequences is not a list")
+
+
+class TestVllmToolParser(unittest.TestCase):
+    """Tests for VllmToolParser class."""
+
+    def test_format_tool_outputs_single(self):
```
finbarrtimbers, line 293: I think you can combine a bunch of these tests as parameterized ones?

open_instruct/tools/tests/test_parsers.py
```diff
@@ -246,6 +253,103 @@ def test_contains_legacy(self):
...
+        """Set up mock actors for each test."""
+        self.patcher = patch("open_instruct.tools.parsers.ray")
+        self.mock_ray = self.patcher.start()
+        self.mock_ray.get.side_effect = lambda x: x
+
+    def tearDown(self):
+        """Stop the patcher."""
+        self.patcher.stop()
+
+    def test_create_legacy_parser(self):
```
finbarrtimbers: Make this parameterized and test all of them?

open_instruct/tools/parsers.py
```diff
@@ -102,29 +114,198 @@ def stop_sequences(self) -> list[str]:
...
+    """
+    Wraps a vLLM tool parser to extract tool calls and format responses.
+
+    This parser delegates to vLLM's native tool parsing implementations
+    (e.g., Hermes, Llama3, Qwen3) while providing a consistent interface.
+    """
+
+    def __init__(
+        self,
+        tool_parser: "VllmNativeToolParser",
```
finbarrtimbers: Is this a real class? Can we reference the class directly?

open_instruct/tools/parsers.py
```diff
@@ -102,29 +114,198 @@ def stop_sequences(self) -> list[str]:
...
+            tool_definitions: Tool definitions in OpenAI format for parsing.
+            output_postfix: Postfix to add after all tool outputs (e.g., generation prompt).
+            output_prefix: Prefix to add before all tool outputs.
+        """
+        self.tool_parser = tool_parser
+        self.output_formatter = output_formatter
+        self._stop_sequences = stop_sequences or []
+        self._tool_definitions = tool_definitions
+        self.output_postfix = output_postfix
+        self.output_prefix = output_prefix
```
finbarrtimbers: Why not make all of these private? 

Also, if __init__ is just setting these, make the class a dataclass?

open_instruct/tools/parsers.py
```diff
@@ -102,29 +114,198 @@ def stop_sequences(self) -> list[str]:
...
+        request = self._make_request()
+        result = self.tool_parser.extract_tool_calls(model_output=text, request=request)
+
+        if not result.tools_called:
+            return []
+
+        tool_calls = []
+        for call in result.tool_calls:
+            try:
+                args = json.loads(call.function.arguments)
```
finbarrtimbers: inline this?

open_instruct/tools/parsers.py
```diff
@@ -102,29 +114,198 @@ def stop_sequences(self) -> list[str]:
...
+        Args:
+            tool_outputs: List of tool output strings to format.
+
+        Returns:
+            Formatted string with prefix, all tool outputs, and postfix.
+        """
+        formatted_parts = [self.output_formatter(output) for output in tool_outputs]
+        return self.output_prefix + "".join(formatted_parts) + self.output_postfix
+
+    def stop_sequences(self) -> list[str]:
```
finbarrtimbers: Remove this and rename _stop_sequences to stop_sequences?

open_instruct/tools/parsers.py
```diff
@@ -102,29 +114,198 @@ def stop_sequences(self) -> list[str]:
...
+        """Format multiple tool outputs with prefix and postfix.
+
+        Args:
+            tool_outputs: List of tool output strings to format.
+
+        Returns:
+            Formatted string with prefix, all tool outputs, and postfix.
+        """
+        formatted_parts = [self.output_formatter(output) for output in tool_outputs]
+        return self.output_prefix + "".join(formatted_parts) + self.output_postfix
```
finbarrtimbers: f-string?

open_instruct/tools/parsers.py
```diff
@@ -102,29 +114,198 @@ def stop_sequences(self) -> list[str]:
...
+    def format_tool_outputs(self, tool_outputs: list[str]) -> str:
+        """Format multiple tool outputs with prefix and postfix.
+
+        Args:
+            tool_outputs: List of tool output strings to format.
+
+        Returns:
+            Formatted string with prefix, all tool outputs, and postfix.
+        """
+        formatted_parts = [self.output_formatter(output) for output in tool_outputs]
```
finbarrtimbers: Move the `"".join` here?

open_instruct/tools/utils.py
```diff
@@ -26,7 +26,7 @@ class ToolsConfig:
     """JSON strings for configuring each tool. Must match length of tools. Use '{}' for defaults."""
 
     tool_parser_type: str = "legacy"
-    """Type of tool parser to use: 'legacy' is the only option for now."""
+    """Type of tool parser to use. Options: "legacy", "vllm_hermes", "vllm_llama3_json", "vllm_olmo3"."""
```
finbarrtimbers: Let's instead say "see "parsers.get_available_parser()` for valid names? That way we don't have to keep the string up to date manually.

open_instruct/dataset_transformation.py
```diff
@@ -1327,6 +1327,7 @@ def rlvr_tokenize_v3(
     ground_truths_key: str = GROUND_TRUTHS_KEY,
     verifier_source_key: str = VERIFIER_SOURCE_KEY,
     system_prompt_override: Optional[str] = None,
+    tool_definitions: Optional[List[Dict[str, Any]]] = None,
```
finbarrtimbers: should be list[dict[str, Any]] | None = None

open_instruct/grpo_fast.py
```diff
@@ -2320,7 +2321,29 @@ def main(
...
+    )
+
+    # Create parser temporarily to get stop sequences for generation config
+    # The actual parser used during generation will be created inside vLLM actors
+    if tool_actors:
+        parser_stop_seqs = create_tool_parser(
+            parser_type=tools_config.tool_parser_type, tool_actors=tool_actors, tokenizer=tokenizer
+        ).stop_sequences()
+        logger.info(f"Adding tool stop sequences to config: {parser_stop_seqs}")
+        streaming_config.stop_strings.extend(parser_stop_seqs)
```
finbarrtimbers: Let's turn all of this into an `initialize_tools` function?

open_instruct/utils.py
```diff
@@ -84,6 +85,27 @@
...
+
+    Args:
+        import_path: Import path in format 'module.submodule:ClassName'.
+
+    Returns:
+        The imported class.
+
+    Raises:
+        ValueError: If the import path format is invalid.
+        ModuleNotFoundError: If the module cannot be found.
```
finbarrtimbers: Nit: These don't belong in the docstring as you're not raising them, but code you're calling is.

open_instruct/tools/parsers.py
hamishivi: Good idea, done!

open_instruct/tools/parsers.py
finbarrtimbers: Still applies!

finbarrtimbers (approved):

PR #1363, opened 2026-01-15
Support list of labels in F1Verifier 

hamishivi:
Allow the string_f1 verifier to accept a list of ground truth labels in addition to a single string. When given a list, the verifier returns the maximum F1 score across all labels, enabling multi-label evaluation.

### DIFF
--- a/open_instruct/ground_truth_utils.py
+++ b/open_instruct/ground_truth_utils.py
@@ -438,8 +438,7 @@ class F1Verifier(VerifierFunction):
         self, tokenized_prediction: list[int], prediction: str, label: str | list[str], query: str | None = None
     ) -> VerificationResult:
         prediction = remove_thinking_section(prediction)
-        # normalize label to a list
-        labels = label if isinstance(label, list) else [label]
+        labels: list[str] = label if isinstance(label, list) else [label]
         score = max(f1_score(prediction, str(lab))["f1"] for lab in labels)
         return VerificationResult(score=score)
 


open_instruct/ground_truth_utils.py
```diff
@@ -426,20 +426,26 @@ def __call__(
...
     ) -> VerificationResult:
         # remove thinking section from the prediction
         prediction = prediction.split("</think>")[-1]
         # remove answer tags from the prediction
         prediction = prediction.replace("<answer>", "").replace("</answer>", "")
-        # return f1 score
-        score = f1_score(prediction, label)["f1"]
+        # if label is a list, take the max F1 score over all labels
+        if isinstance(label, list):
+            score = max(f1_score(prediction, str(lab))["f1"] for lab in label)
```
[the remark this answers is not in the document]
hamishivi: This is an error in the dataset and we shouldn't handle it.

open_instruct/ground_truth_utils.py
```diff
@@ -426,20 +426,22 @@ def __call__(
...
+        self, tokenized_prediction: list[int], prediction: str, label: str | list[str], query: str | None = None
     ) -> VerificationResult:
-        # remove thinking section from the prediction
-        prediction = prediction.split("</think>")[-1]
-        # remove answer tags from the prediction
-        prediction = prediction.replace("<answer>", "").replace("</answer>", "")
-        # return f1 score
-        score = f1_score(prediction, label)["f1"]
+        prediction = remove_thinking_section(prediction)
+        if isinstance(label, list):
```
finbarrtimbers: Can you normalize this to a list and then go through the same code path always? 

Wdyt about making callers pass in a single element list instead of a string?

hamishivi: WRT list: I'd like to, but then some of our old datasets will just break. I think it's okay to be a bit flexible here?

Fixed the normalization thing.

open_instruct/ground_truth_utils.py
```diff
@@ -426,20 +426,21 @@ def __call__(
...
+        self, tokenized_prediction: list[int], prediction: str, label: str | list[str], query: str | None = None
     ) -> VerificationResult:
-        # remove thinking section from the prediction
-        prediction = prediction.split("</think>")[-1]
-        # remove answer tags from the prediction
-        prediction = prediction.replace("<answer>", "").replace("</answer>", "")
-        # return f1 score
-        score = f1_score(prediction, label)["f1"]
+        prediction = remove_thinking_section(prediction)
+        # normalize label to a list
```
finbarrtimbers: remove the comment and add a type annotation (`labels: list[str] = ...`) to the assignment on the next line?

PR #1364, opened 2026-01-15
Fix main bug

hamishivi:
I thought I had fixed this, but turns out there were some other issues. Generally simplified the code such that we don't allow `tools` to be `None`.

### DIFF
--- a/open_instruct/test_grpo_fast_gpu.py
+++ b/open_instruct/test_grpo_fast_gpu.py
@@ -133,7 +133,9 @@ class TestGeneration(TestGrpoFastBase):
 
         tokenizer_name = "Qwen/Qwen3-1.7B"
         tool_actors = (
-            create_tools([ParsedToolConfig(name="python", call_name="code", config={"api_endpoint": self.tool_api_endpoint})])
+            create_tools(
+                [ParsedToolConfig(name="python", call_name="code", config={"api_endpoint": self.tool_api_endpoint})]
+            )
             if use_tools
             else None
         )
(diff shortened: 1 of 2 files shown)


open_instruct/tools/utils.py
```diff
@@ -40,33 +40,34 @@ class ToolsConfig:
...
+        if len(self.tool_call_names) != len(self.tools):
+            raise ValueError(
+                f"tool_call_names must have same length as tools. "
+                f"Got {len(self.tool_call_names)} names for {len(self.tools)} tools."
+            )
+        if len(self.tool_configs) != len(self.tools):
+            raise ValueError(
+                f"tool_configs must have same length as tools. "
+                f"Got {len(self.tool_configs)} configs for {len(self.tools)} tools."
+            )
```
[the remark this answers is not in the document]
hamishivi: I don't really like the getattr thing.

open_instruct/tools/utils.py
```diff
@@ -15,10 +15,10 @@
 class ToolsConfig:
     """Configuration for tools used during generation."""
 
-    tools: list[str] | None = None
+    tools: list[str] = field(default_factory=list)
```
finbarrtimbers, line 32: I think we require that all of these have the same length, right? Can we instead have a list of a dataclass? That will enforce that requirement naturally. 

Something like:

```
tools: list[ToolAndInfo]

@dataclasses.dataclass
class ToolAndInfo:
    tool: str
    tool_call_name: str
    config: ToolConfig
```

hamishivi: This is exposed through the CLI, so I think we have to expose the lists anyway. I'm a little open to just always passing in the full config dict, but I do think thats a biiiiit annoying if some tool has reasonable defaults (e.g., some constant function).

finbarrtimbers: oh that's annoying. fair enough.

PR #1366, opened 2026-01-15
Per tool metrics logging

hamishivi:
Adds logging for each tool in the run, which is useful to give more visibility over what the model is doing.

### DIFF
--- a/open_instruct/test_streaming_data_loader_gpu.py
+++ b/open_instruct/test_streaming_data_loader_gpu.py
@@ -29,6 +29,7 @@ from open_instruct.dataset_transformation import (
 from open_instruct.ground_truth_utils import RewardConfig
 from open_instruct.grpo_fast import create_tools
 from open_instruct.test_grpo_fast import TestGrpoFastBase
+from open_instruct.tools.utils import ParsedToolConfig
 from open_instruct.utils import maybe_update_beaker_description
 from open_instruct.vllm_utils import SamplingConfig, create_vllm_engines
 
@@ -184,7 +185,7 @@ class TestStreamingDataLoaderGPU(TestGrpoFastBase):
         self._ray_queues.extend([param_prompt_Q, inference_results_Q, eval_results_Q])
 
         tool_actors = create_tools(
-            tools=["python"], tool_call_names=["code"], tool_configs=[{"api_endpoint": self.tool_api_endpoint}]
+            [ParsedToolConfig(name="python", call_name="code", config={"api_endpoint": self.tool_api_endpoint})]
         )
 
         pg = placement_group([{"GPU": 1, "CPU": 1}], strategy="PACK")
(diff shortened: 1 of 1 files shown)


open_instruct/data_loader.py
```diff
@@ -1088,17 +1131,16 @@ def _data_preparation_loop(self):
...
-                    ).mean(),
-                    "val/tool_runtimes_rate": np.array(result.request_info.tool_runtimes).mean(),
-                    "val/tool_calleds_rate": np.array(result.request_info.tool_calleds).mean(),
                     **reward_metrics,
                     **batch_metrics_prefixed,
                 }
 
+                step_metrics.update(compute_tool_metrics(
+                    result.request_info.tool_call_stats,
+                    num_rollouts=len(result.request_info.num_calls),
```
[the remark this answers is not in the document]
hamishivi: Hm, this is wrong I think? total_prompts on batch_stats doesn't give rollouts number?

open_instruct/data_loader.py
```diff
@@ -46,6 +47,43 @@
...
 
 
+def compute_tool_metrics(
+    tool_call_stats: list[list[data_types.ToolCallStats]], num_rollouts: int, tool_names: list[str]
+) -> dict[str, float]:
+    """Compute per-tool and aggregate tool metrics (tools/{name}/avg_calls_per_rollout, failure_rate, avg_runtime)."""
+    if not num_rollouts or not tool_names:
+        return {}
+
+    counts: defaultdict[str, int] = defaultdict(int)
```
finbarrtimbers: Up to you: but what do you think about having a dataclass (`ToolStatistics`) here, and you could have support for adding a rollout (or a sequence) to the class?

open_instruct/data_loader.py
```diff
@@ -46,6 +47,43 @@
...
+    runtimes: defaultdict[str, float] = defaultdict(float)
+
+    for rollout in tool_call_stats:
+        for s in rollout:
+            counts[s.tool_name] += 1
+            failures[s.tool_name] += not s.success
+            runtimes[s.tool_name] += s.runtime
+
+    metrics: dict[str, float] = {}
+    total_c = total_f = 0
```
finbarrtimbers: Let's use more informative names here than c/f/r and total_{c,f,r}?

finbarrtimbers (changes requested):

open_instruct/data_loader.py
hamishivi: Yeah, good idea! done.

open_instruct/data_loader.py
```diff
@@ -1088,17 +1094,15 @@ def _data_preparation_loop(self):
...
-                    "val/tool_runtimes_rate": np.array(result.request_info.tool_runtimes).mean(),
-                    "val/tool_calleds_rate": np.array(result.request_info.tool_calleds).mean(),
                     **reward_metrics,
                     **batch_metrics_prefixed,
                 }
 
+                tool_stats = ToolStatistics(tool_names=self.tool_names)
+                for rollout_stats in result.request_info.tool_call_stats:
+                    tool_stats.add_rollout(rollout_stats)
+                step_metrics.update(tool_stats.compute_metrics())
```
finbarrtimbers, line 1104: so clean!

PR #1376, opened 2026-01-15
Add dr tulu parser

hamishivi:
Adds a parser to handle using the `dr-agent-lib` library, which should allow DR Tulu-style training.
Also adds a dr-agent tool, which essentially is a wrapper around calls.

The reason for a different pattern here is that the dr-agent-lib basically handles all the tool detection, implementation, and parsing for us, so we delegate stuff down into it here. This is just a different pattern to e.g. the vllm tool handling. Nonetheless, I want to support it so we can train dr-tulu natively in open instruct in future.

The diff looks large, but it is mainly because we add a new (optional) dependency and update lock/requirements accordingly.

### DIFF
--- a/open_instruct/tools/tests/test_parsers.py
+++ b/open_instruct/tools/tests/test_parsers.py
@@ -492,7 +492,7 @@ class TestCreateToolParser(unittest.TestCase):
 
     def test_create_dr_tulu_parser(self):
         """Test creating dr_tulu parser."""
-        mock_actor = create_mock_tool_actor("search")
+        mock_actor = create_mock_tool_actor("dr_agent_mcp")
         mock_tokenizer = MagicMock()
 
         parser = create_tool_parser("dr_tulu", tokenizer=mock_tokenizer, tool_actors=[mock_actor])


open_instruct/tools/tools.py
```diff
@@ -354,10 +377,129 @@ class SerperSearchToolConfig(BaseToolConfig):
...
+        else:
+            result = ToolOutput(
+                output="\n".join(outputs),
+                called=True,
+                error="; ".join(errors) if errors else "",
+                timeout=any_timeout,
+                runtime=time.time() - start_time,
+            )
+        _log_tool_call(self.call_name, text, result)
+        return result
```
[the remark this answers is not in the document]
hamishivi: We only expect one MCP call at a time, so I think this is okay for now.

scripts/train/debug/tools/debug_dr_tulu_training.sh
finbarrtimbers, line 1: wdyt about having a scripts/train/dr_tulu dir?

open_instruct/tools/utils.py
```diff
@@ -50,6 +50,9 @@ class ToolsConfig:
     only_reward_good_outputs: bool = False
     """Only apply rewards to outputs from tools that didn't error."""
 
+    pass_tools_to_chat_template: bool = True
```
finbarrtimbers, line 53: Should this just always be true?

open_instruct/tools/utils.py
```diff
@@ -87,6 +90,15 @@ def __post_init__(self):
                 raise ValueError(f"Invalid tool_config for tool {tool_name} at index {i}: {e}") from e
             self._parsed_tools.append(ParsedToolConfig(name=tool_name, call_name=call_name, config=config))
 
+        has_mcp_tool = "dr_agent_mcp" in self.tools
```
finbarrtimbers: Can we pull this out into the dr tulu parser or somewhere else? It seems very specific

open_instruct/tools/utils.py
hamishivi: No, actually I end up setting it to false for qwen + dr tulu training, as then qwen adds its own tool calling format onto the end end of the system prompt, and it tries to do tool calls in a different format.

That is, sometimes we want to call tools in a different way to what the model trainer intended, but we can't remove the original tool call format from the system prompt without avoiding passing in tools to the chat template.

scripts/train/debug/tools/debug_dr_tulu_training.sh
hamishivi: Yeah maybe... I'll make a separate PR with debug scripts for each of the 3 parsers. I think for now, just having it under tools is okay, I don't expect adding lots of dr tulu training scripts for now?

PR #1377, opened 2026-01-15
Add Crawl4AIBrowseTool for fetching webpage content

hamishivi:
Add a new browse tool that uses the Crawl4AI Docker API to fetch and convert webpage content to clean markdown. 

This Docker API is actually a proxy server for crawl4ai, which is useful since (a) jina costs money, but we can host crawl4ai ourself; (b) ips can get burned when crawling so we always want to access crawl4ai through a proxy service. The proxy service itself was developed as part of the Dr Tulu project.

Based on: https://github.com/rlresearch/dr-tulu/blob/main/agent/dr_agent/tool_interface/mcp_tools.py

### DIFF
--- a/open_instruct/tools/tools.py
+++ b/open_instruct/tools/tools.py
@@ -166,7 +166,6 @@ class JinaBrowseTool(Tool):
             content = inner_data.get("content") or ""
             title = inner_data.get("title") or ""
 
-            # Format output with title if available
             if title and content:
                 content = f"# {title}\n\n{content}"
         else:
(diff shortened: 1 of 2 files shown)


open_instruct/tools/tools.py
```diff
@@ -354,10 +354,149 @@ class SerperSearchToolConfig(BaseToolConfig):
...
+            return result
+
+        start_time = time.time()
+
+        # Build request payload
+        payload = {
+            "urls": url.strip(),
+            "word_count_threshold": 10,
+            "bypass_cache": self.bypass_cache,
+            "remove_overlay_elements": True,
```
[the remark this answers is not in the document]
hamishivi: Not going to do this for now. These defaults match what we did in past. We can make configurable in future if needed.

open_instruct/tools/tools.py
```diff
@@ -354,10 +353,165 @@ class SerperSearchToolConfig(BaseToolConfig):
...
+            return result
+
+        start_time = time.time()
+
+        # Build request payload matching Crawl4AI Docker API format
+        # See: https://docs.crawl4ai.com/core/docker-deployment/
+        crawler_params: dict = {
+            "cache_mode": "bypass" if self.bypass_cache else "enabled",
+            "word_count_threshold": 10,
+            "exclude_social_media_links": True,
```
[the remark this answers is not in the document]
hamishivi: This is intentional, we just want to hardcode this preference for social media links for now, matching dr tulu training

open_instruct/tools/tools.py
```diff
@@ -354,10 +353,168 @@ class SerperSearchToolConfig(BaseToolConfig):
...
+            _log_tool_call(self.call_name, url, result)
+            return result
+
+        # Extract content from Crawl4AI response
+        # API returns: {"results": [{"url": ..., "markdown": ..., "success": ..., ...}]}
+        data = api_response.data
+        content = ""
+        error = ""
+
+        results = data.get("results", [])
```
finbarrtimbers: Can you refactor this into a function? It's a really dense chunk of code.

finbarrtimbers (changes requested):

open_instruct/tools/tools.py
hamishivi: done!

open_instruct/tools/tools.py
```diff
@@ -354,10 +353,168 @@ class SerperSearchToolConfig(BaseToolConfig):
...
+            _log_tool_call(self.call_name, url, result)
+            return result
+
+        content, error = self._parse_response(api_response.data)
+        output = error if error else content
+        result = ToolOutput(output=output, called=True, error=error, timeout=False, runtime=time.time() - start_time)
+        _log_tool_call(self.call_name, url, result)
+        return result
+
+    def _parse_response(self, data: dict) -> tuple[str, str]:
```
finbarrtimbers: Can you make this a free function? I think you only use self.include_html and self.max_length

finbarrtimbers (approved):

PR #1382, opened 2026-01-16
Add per-sample tools and active_tools support for datasets

hamishivi:
This PR allows us to specify what tools can be used on a per-sample level, which is useful for when we want to train our model to handle settings with varying tools available to it. We do this by reading a new column from a dataset: "tools", which is a list of the *call names* of the active tools. See https://huggingface.co/datasets/hamishivi/tulu_3_rewritten_tools_test for an example of such a column.

When we observe a tools column:
- we validate that all tools in the dataset are configured in the train command, error otherwise.
- we pass only the tools active for that sample to the chat prompt tokenizer (unless `pass_tools_to_chat_template=False`, in which case we just ignore them). Note that we allow empty lists (no active tools).
- We set a list of active tools for each sample, and **skip executing tools that aren't active**. In the future, we could consider adding some warning message to the model or similar. For now, I just treat it like an unparsed tool call: we dont do anything with it.

If we don't see a tools column, we just assume that all tools are active for all samples.

Also adds a new fingerprinting setup for the HF +open instruct cache to make updates to the data transformation a bit easier to deal with (increment a central variable).

### DIFF
--- a/open_instruct/dataset_transformation.py
+++ b/open_instruct/dataset_transformation.py
@@ -957,11 +957,8 @@ def validate_dataset_tools(dataset: Dataset, configured_tool_names: list[str], d
 
     dataset_tool_names: set[str] = set()
     for tools in dataset[TOOLS_COLUMN_KEY]:
-        if tools is None:
-            continue
-        for tool_name in tools:
-            if tool_name:
-                dataset_tool_names.add(tool_name)
+        if tools is not None:
+            dataset_tool_names.update(t for t in tools if t)
 
     unconfigured_tools = dataset_tool_names - set(configured_tool_names)
     if unconfigured_tools:


open_instruct/dataset_transformation.py
```diff
@@ -918,6 +920,74 @@ def remove_dataset_source_field(dataset: Dataset) -> Dataset:
...
+            f"These tools will be available but never used by this dataset."
+        )
+
+
+def dataset_has_tools_column(dataset: Dataset) -> bool:
+    """Check if a dataset has a 'tools' column for per-sample tool definitions."""
+    return TOOLS_COLUMN_KEY in dataset.column_names
+
+
+def get_active_tools_from_sample(sample_tools: list[str] | None) -> list[str] | None:
```
finbarrtimbers: This seems redundant. Just inline the logic?

open_instruct/dataset_transformation.py
```diff
@@ -1341,8 +1437,16 @@ def rlvr_tokenize_v3(
...
-        chat_template_kwargs["tools"] = tool_definitions
+    if pass_tools_to_chat_template and tool_definitions:
+        sample_active_tools = row.get(TOOLS_COLUMN_KEY)
+        if sample_active_tools is not None:
+            active_tool_names = set(sample_active_tools)
+            filtered_tools = [t for t in tool_definitions if t.get("function", {}).get("name") in active_tool_names]
+            if filtered_tools:
+                chat_template_kwargs["tools"] = filtered_tools
+        else:
+            chat_template_kwargs["tools"] = tool_definitions
```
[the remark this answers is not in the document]
hamishivi: I'll do this in future, previously we had the different filtering functions for backwards repro but less into that now.

open_instruct/dataset_transformation.py
```diff
@@ -918,6 +920,78 @@ def remove_dataset_source_field(dataset: Dataset) -> Dataset:
     return dataset
 
 
+TOOLS_COLUMN_KEY = "tools"
+
+# Cache version: increment this when transformation logic changes significantly
+# to invalidate old caches. v2: Added per-sample tool filtering in rlvr_tokenize_v3.
+DATASET_CACHE_VERSION = "v2"
```
finbarrtimbers, line 936: Smart!

open_instruct/dataset_transformation.py
```diff
@@ -918,6 +920,78 @@ def remove_dataset_source_field(dataset: Dataset) -> Dataset:
...
+        dataset_name: Name of the dataset for error messages.
+
+    Raises:
+        ValueError: If dataset contains tools that are not configured.
+    """
+    if TOOLS_COLUMN_KEY not in dataset.column_names:
+        return
+
+    dataset_tool_names = {
+        tool_name for tools in dataset[TOOLS_COLUMN_KEY] if tools for tool_name in tools if tool_name
```
finbarrtimbers: this is super dense. Can you split it into multiple comprehensions or a for loop?

open_instruct/dataset_transformation.py
```diff
@@ -918,6 +920,78 @@ def remove_dataset_source_field(dataset: Dataset) -> Dataset:
...
+    }
+    configured_set = set(configured_tool_names) if configured_tool_names else set()
```
finbarrtimbers: Do we need the conditional? What happens if it's falsey? Is it None or empty? (if it's None, the type annotation is wrong.)

open_instruct/dataset_transformation.py
```diff
@@ -918,6 +920,78 @@ def remove_dataset_source_field(dataset: Dataset) -> Dataset:
...
+
+    unconfigured_tools = dataset_tool_names - configured_set
```
finbarrtimbers: Inline this! And do we actually need the set difference? Instead, let's just check for equality?

open_instruct/dataset_transformation.py
```diff
@@ -918,6 +920,78 @@ def remove_dataset_source_field(dataset: Dataset) -> Dataset:
...
+    extraneous_tools = configured_set - dataset_tool_names
+    if extraneous_tools:
+        logger.warning(
+            f"Configured tools {sorted(extraneous_tools)} are not found in {dataset_name}'s "
+            f"'{TOOLS_COLUMN_KEY}' column. Tools found in dataset: {sorted(dataset_tool_names)}. "
+            f"These tools will be available but never used by this dataset."
+        )
+
+
+def dataset_has_tools_column(dataset: Dataset) -> bool:
```
finbarrtimbers: As it's only used once, let's inline this?

open_instruct/dataset_transformation.py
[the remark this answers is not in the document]
finbarrtimbers: Add a TODO please!

open_instruct/dataset_transformation.py
```diff
@@ -1551,24 +1655,35 @@ def get_dataset_v1(dc: DatasetConfig, tc: TokenizerConfig):
         fn_kwargs = {"tokenizer": tokenizer}
         fn_kwargs.update(fn_args)
 
+        # Compute a custom fingerprint that includes DATASET_CACHE_VERSION to invalidate
+        # HuggingFace's internal .map() cache when transformation logic changes significantly
+        fingerprint_data = (
```
finbarrtimbers: Inline this!

open_instruct/dataset_transformation.py
```diff
@@ -1551,24 +1655,35 @@ def get_dataset_v1(dc: DatasetConfig, tc: TokenizerConfig):
...
+            f"{DATASET_CACHE_VERSION}:{fn_name}:{dataset._fingerprint}:{json.dumps(fn_args, sort_keys=True)}"
+        )
+        new_fingerprint = hashlib.sha256(fingerprint_data.encode()).hexdigest()[:16]
+
         # perform the transformation
         target_columns = dataset.column_names if dc.target_columns is None else dc.target_columns
         # Always preserve dataset_source if it exists
         if DATASET_ORIGIN_KEY in dataset.column_names and DATASET_ORIGIN_KEY not in target_columns:
             target_columns = target_columns + [DATASET_ORIGIN_KEY]
+        if TOOLS_COLUMN_KEY in dataset.column_names and TOOLS_COLUMN_KEY not in target_columns:
```
finbarrtimbers: Can you combine this logic with the DATASET_ORIGIN_KEY logic above?

open_instruct/model_utils.py
```diff
@@ -94,6 +95,7 @@ def __getitem__(self, key: slice | int | list[int]) -> "Batch":
                 decoded_responses=self.decoded_responses[key] if self.decoded_responses is not None else None,
                 indices=self.indices[key] if self.indices is not None else None,
                 scores=self.scores[key] if self.scores is not None else None,
+                active_tools=self.active_tools[key] if self.active_tools is not None else None,
```
finbarrtimbers: Let's compute

active_tools = self.active_tools[key] if self.active_tools is not None else None

once and then reuse it throughout?

open_instruct/vllm_utils.py
```diff
@@ -872,11 +873,16 @@ async def process_request(actor: LLMRayActor, sub_request_id: str, sampling_para
     excess_tool_calls: dict[str, int] = {}
 
     base_request_id = split_request_id(sub_request_id)["base_id"]
-    original_prompt = actor.request_metadata[base_request_id]["prompt_token_ids"]
+    request_metadata = actor.request_metadata[base_request_id]
+    original_prompt = request_metadata["prompt_token_ids"]
+    active_tools = request_metadata.get("active_tools")
```
finbarrtimbers: Please use ["active_tools"]! only use .get if there's a fallback.

open_instruct/dataset_transformation.py
hamishivi: I think it's useful to tell the user what tools arent configured? The dataset might be large, and it might be hard to find the exact tool that you need to configure to get training going (e.g., one random sample out of 100k might have some additional tool, and then either you need to edit the data or add the tool).

open_instruct/dataset_transformation.py
hamishivi: Ah, good point. Removed.

open_instruct/dataset_transformation.py
```diff
@@ -918,6 +929,60 @@ def remove_dataset_source_field(dataset: Dataset) -> Dataset:
...
+        ValueError: If dataset contains tools that are not configured.
+    """
+    if TOOLS_COLUMN_KEY not in dataset.column_names:
+        return
+
+    dataset_tool_names: set[str] = set()
+    for tools in dataset[TOOLS_COLUMN_KEY]:
+        if tools is None:
+            continue
+        for tool_name in tools:
```
finbarrtimbers: make this a comprehension? 

dataset_tool_names.update(t for t in tools if f)

PR #1387, opened 2026-01-19
Add fp32 LM head option for GRPO

natolambert:
## Summary

Adds FP32 LM head projection option for GRPO training to reduce logprob mismatch between the vLLM generator and trainer forward pass.

### Background

In GRPO/RL training, the generator (vLLM) and trainer compute logprobs on the same sequences. Due to bf16 precision loss in the LM head projection (`hidden @ lm_head.T -> logits -> softmax -> logprobs`), these can diverge by 2-24+ nats, which affects training stability (see ScaleRL paper, Figure 3).

### The LM Head and Weight Tying

Many models (Qwen, LLaMA, etc.) use **weight tying**: the input embedding matrix (`embed_tokens`) and output projection (`lm_head`) share the same underlying weight tensor to save memory.

- **embed_tokens**: Lookup operation - `W[token_id]` returns that token's embedding vector
- **lm_head**: Matrix multiply - `hidden @ W.T` computes similarity to all token embeddings

Same matrix, different operations. This matters for the fp32 implementation because naively converting `lm_head` to fp32 would also convert the embedding layer.

### Two Modes Implemented

**Cache Mode** (`--fp32_lm_head true`):
- Keeps bf16 weights unchanged
- Patches forward pass to cast inputs/weights to fp32 on-the-fly
- vLLM maintains a separate `_open_instruct_fp32_weight` cache updated after each weight sync
- Safe with tied weights (no modification to underlying storage)
- Small compute overhead from casting each forward pass

**Permanent Mode** (`--fp32_lm_head true --fp32_lm_head_permanent true`):
- Unties lm_head from embed_tokens (creates independent copy if needed)
- Converts lm_head weights to fp32 in-place
- vLLM receives fp32 weights directly during sync
- No runtime casting overhead
- ~15-20% extra memory for the untied fp32 lm_head matrix

Both modes use roughly the same extra memory (~500MB for 0.5B model), but permanent avoids repeated casting.

## Changes

**Trainer side** (`model_utils.py`):
- `enable_fp32_lm_head(model, permanent=False)` - patches or converts lm_head
- `_untie_lm_head_if_needed(model)` - safely separates tied weights before fp32 conversion

**vLLM side** (`vllm_utils.py`, `vllm_utils_workerwrap.py`):
- `patch_vllm_for_fp32_logits()` - patches LogitsProcessor._get_logits for fp32 computation
- `_sync_fp32_lm_head()` - maintains fp32 cache or converts weights after trainer sync
- Env var `OPEN_INSTRUCT_FP32_LM_HEAD` controls mode ("1"=cache, "2"=permanent)

**GRPO** (`grpo_fast.py`):
- `--fp32_lm_head` flag enables fp32 LM head
- `--fp32_lm_head_permanent` flag selects permanent mode
- Applied to policy model, reference policy, and vLLM engines

## Test plan

- [x] `uv run pytest open_instruct/test_vllm_utils.py open_instruct/test_rl_utils.py`
- [x] GRPO smoke test: Qwen2.5-0.5B GSM8K with fp32 enabled

### Beaker GPU Tests (post-merge with main)

| Mode | Beaker Job | Status |
|------|------------|--------|
| Cache (`OPEN_INSTRUCT_FP32_LM_HEAD=1`) | [01KFHHEEKEGV6GR341XXKJ3RH9](https://beaker.org/orgs/ai2/workspaces/open-instruct-dev/work/01KFHHEEKEGV6GR341XXKJ3RH9) | ✅ Passed |
| Permanent (`OPEN_INSTRUCT_FP32_LM_HEAD=2`) | [01KFHHF9Z2QEVQ3A7H8SG606KC](https://beaker.org/orgs/ai2/workspaces/open-instruct-dev/work/01KFHHF9Z2QEVQ3A7H8SG606KC) | ✅ Passed |

### DGX Spark Long Runs (312 steps each, ~55 min per run)

| Run | Config | Wandb |
|-----|--------|-------|
| Long + NO FP32 | baseline | https://wandb.ai/ai2-llm/open_instruct_internal/runs/fw73k9s8 |
| Long + FP32 PERMANENT | `--fp32_lm_head --fp32_lm_head_permanent` | https://wandb.ai/ai2-llm/open_instruct_internal/runs/gmif0osr |

### DGX Spark Short Runs (78 steps each)

| Run | Config | Wandb |
|-----|--------|-------|
| Short + NO FP32 | baseline | https://wandb.ai/ai2-llm/open_instruct_internal/runs/ajujoluc |
| Short + FP32 PERMANENT | `--fp32_lm_head --fp32_lm_head_permanent` | https://wandb.ai/ai2-llm/open_instruct_internal/runs/r0i7x38r |

### DIFF
--- a/scripts/analysis/fp32-lm-head/run_multi_revision_analysis.sh
+++ b/scripts/analysis/fp32-lm-head/run_multi_revision_analysis.sh
@@ -15,7 +15,7 @@
 set -euo pipefail
 
 BEAKER_USER=$(beaker account whoami --format json | jq -r '.[0].name')
-BEAKER_IMAGE="${BEAKER_IMAGE:-${BEAKER_USER}/open_instruct_auto}"
+BEAKER_IMAGE="${1:-${BEAKER_USER}/open_instruct_auto}"
 
 # Model configuration
 MODEL="${MODEL:-allenai/Olmo-3-7B-Think}"


natolambert:
Note on vLLM fp32 LM head options we discussed:

- Option A (true fp32 matmul on the fly): patch `_get_logits` to use `F.linear(hidden_states.float(), lm_head.weight.float(), bias.float())` each call. Simple but slower (per-token cast + fp32 GEMM).
- Option B (cached fp32 head weights): keep an fp32 copy of `lm_head.weight` updated on weight sync; `_get_logits` uses cached fp32 weights with fp32 hidden_states. Lower runtime cost; modest extra VRAM (head size). Skips quantized heads.

You chose Option B because KV cache dominates memory in RL, so the head cache is relatively small while preserving fp32 head matmul.

TODO: current debug metric (`debug/vllm_vs_local_logprob_diff_*`) is per-sample overwrite; we should log batch-averaged stats for a smoother comparison curve.

PR #1390, opened 2026-01-20
Adds type checking for `dataset_transformation.py`. 

finbarrtimbers:
Also enables rule `UP006`, which bans using `from typing import List, Dict`, etc. instead of using the native `list`, `dict`, etc. types.


natolambert (approved):
LGTM

PR #1397, opened 2026-01-20
Add GRPO callbacks for OLMo-core Trainer (GRPO olmo-core: PR 3 of 5)

finbarrtimbers:
Adds callbacks for GRPO training with OLMo-core's Trainer:
- `VLLMWeightSyncCallback`: syncs weights to vLLM engines after each step
- `RefPolicyUpdateCallback`: Polyak averaging for reference policy updates

Based on PR #1412 (GRPOTrainModule)

GPU_TESTS=[01KFKCWJQKNEB71EZSA93XRKWF](https://beaker.org/orgs/ai2/workspaces/open-instruct-dev/work/01KFY0HAY4NGENR6XGA3Q0VPCS?taskId=01KFY0HAY99NJWX6EZFTGZ98KM&jobId=01KFY2M70ZKT2VEYZ8QK0GVMEC)

### DIFF
--- a/open_instruct/grpo_callbacks.py
+++ b/open_instruct/grpo_callbacks.py
@@ -1,0 +1,189 @@
+"""
+GRPO-specific callbacks for OLMo-core Trainer.
+
+These callbacks handle:
+- vLLM weight synchronization after each training step
+- Reference policy Polyak updates
+"""
+
+import contextlib
+import re
+from collections.abc import Callable
+from dataclasses import dataclass, field
+from typing import Any, cast
+
+import ray
+import ray.exceptions
+import torch
+import torch.distributed as dist
+import torch.nn as nn
+from olmo_core.train.callbacks import Callback
+from olmo_core.train.train_module import TransformerTrainModule
+from torch.distributed._composable.fsdp import FSDPModule
+from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
+
+from open_instruct import logger_utils, vllm_utils
+
+logger = logger_utils.setup_logger(__name__)
+
+_BLOCK_PATTERN = re.compile(r"blocks\.(\d+)\.(.*)")
+_OLMO_CORE_TO_HF_LAYER_MAPPINGS = {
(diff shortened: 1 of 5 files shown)


open_instruct/grpo_callbacks.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def post_step(self) -> None:
+        if self.trainer.global_step % self.update_interval != 0:
+            return
+
+        model = self.train_module.model
+        is_fsdp = isinstance(model, FSDP)
+
+        if is_fsdp:
+            with FSDP.summon_full_params(model, writeback=False, rank0_only=False):
```
hamishivi: If the model is really large, does this cause OOM? Since we had to turn off model gathering for e.g. 32B olmo.

open_instruct/vllm_utils.py
```diff
@@ -1202,26 +1222,30 @@ def broadcast_weights_to_vllm(
     """
     is_rank_0 = torch.distributed.get_rank() == 0
     params = list(model.named_parameters())
-    num_params = len(params)
-    all_refs: list[ray.ObjectRef] = []
+
+    if isinstance(model, FSDP) and not gather_whole_model:
```
hamishivi, line 1434: similar to the other comment. Doesn't this mean we can't scale with olmo-core?

open_instruct/grpo_callbacks.py
finbarrtimbers: Yes! This is definitely going to have to be fixed at some point.

finbarrtimbers: Ok fixed! I think the _actual_ issue is that FSDP2 doesn't support `whole_model_gather=True`.

open_instruct/vllm_utils.py
finbarrtimbers: We're good!

finbarrtimbers:
> Happy to merge this, although probably we want to more thoroughly test these code chunks once more the the GRPO implementation is in.

Agreed! I did add some test, but I agree completely.

PR #1398, opened 2026-01-20
Add OLMo-core Ray actor (GRPO olmo-core: PR 4 of 5)

finbarrtimbers:
## Summary
- Add `grpo_olmo_core_actor.py` with `PolicyTrainerOLMoCoreProcess` and `OLMoCoreModelGroup` classes
- Add file to type checking in `pyproject.toml`

This Ray actor wraps OLMo-core's training infrastructure for distributed GRPO training across multiple GPUs and nodes.

PR #1399, opened 2026-01-20
Add GRPO main entry point and scripts (GRPO olmo-core: PR 5 of 5)

finbarrtimbers:
## Summary
- Add `grpo.py` main training orchestration script using OLMo-core's Trainer with Ray actors
- Add test scripts: `single_gpu_grpo.sh`, `multi_node_grpo.sh`, `tool_grpo.sh`
- Unify duplicated functions between `grpo.py` and `grpo_fast.py`: `grpo.py` now calls `grpo_fast.setup_runtime_variables` and `grpo_fast.create_generation_configs` instead of maintaining its own copies
- Move `is_beaker_job()` guard into `maybe_get_beaker_config()` so callers don't need to wrap every call
- Fix bug in `grpo_fast.create_generation_configs` where `vllm_config` was referenced but not passed as a parameter
- Add `grpo.py` to type checking in `pyproject.toml`
- Update `CHANGELOG.md`

**Depends on**: #1398

GPU_TESTS=01KKY8PKQYXPDTJFCT37Q20E9X

Runs:

1. Single GPU GRPO: [Beaker](https://beaker.org/ex/01KKY8PKQYXPDTJFCT37Q20E9X)
2. Multi-node GRPO: [Beaker](https://beaker.org/ex/01KKYBG9FXJWXDR11BDHQRM5K9)

### DIFF
--- a/open_instruct/grpo.py
+++ b/open_instruct/grpo.py
@@ -42,8 +42,10 @@ from open_instruct.model_utils import ModelConfig, push_folder_to_hub
 
 logger = logger_utils.setup_logger(__name__)
 
+CLUSTER_STARTUP_TIMEOUT_S = 1200
 
-@backoff.on_predicate(backoff.constant, interval=5, max_time=300)
+
+@backoff.on_predicate(backoff.constant, interval=5, max_time=CLUSTER_STARTUP_TIMEOUT_S)
 def wait_for_gpus(expected_gpus: int) -> bool:
     """Poll the Ray cluster until ``expected_gpus`` GPUs are available.
 


open_instruct/grpo.py
```diff
@@ -0,0 +1,324 @@
...
+from open_instruct.dataset_transformation import TokenizerConfig
+from open_instruct.environments.tools.utils import EnvsConfig
+from open_instruct.ground_truth_utils import RewardConfig, build_all_verifiers
+from open_instruct.grpo_olmo_core_actor import OLMoCoreModelGroup
+from open_instruct.model_utils import ModelConfig, push_folder_to_hub
+
+logger = logger_utils.setup_logger(__name__)
+
+
+@backoff.on_predicate(backoff.constant, interval=5, max_time=300)
```
hamishivi: should the timeout be extended here to ~20min?

finbarrtimbers: good point. done.

PR #1400, opened 2026-01-20
Bumped a bunch of library versions and small bugfixes

finbarrtimbers:
I ran `uv lock --upgrade`. Then, ran the following experiments: 

1. Single GPU GRPO: [Beaker](https://beaker.org/ex/01KFGGSWHSHWNZ9SYZC72R7X6E)
2. Single GPU GRPO with tools: [Beaker](https://beaker.org/ex/01KFGHCSRF6FE3TRPW16BNRJ1B)

This revealed some bugs, which I fixed:

- `loss_denominator_type` was having issues because of the `str | float` setup. I changed it to always be a `str`, and if it's supposed to be a float, we convert it at the call-site, letting us get rid of the `get_denominator` function. 
- Updated `tool_grpo_fast.sh` to use the `SERPER_API_KEY` secret
- Fixed a few misc type annotation errors.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -46,7 +46,8 @@ jobs:
             echo "" >> comment.md
 
             # Get list of changed files, excluding verbose files
-            changed_files=$(diff -rq site-pr site-base | grep "^Files" | grep "differ$" | grep -v "search_index.json" | head -5)
+            # diff returns exit code 1 when differences exist; || true prevents pipefail from failing the script
+            changed_files=$(diff -rq site-pr site-base | grep "^Files" | grep "differ$" | grep -v "search_index.json" | head -5 || true)
 
             if [ -n "$changed_files" ]; then
               file_count=0


open_instruct/grpo_utils.py
```diff
@@ -181,7 +180,8 @@ def __post_init__(self):
                 "Cannot use both `use_vllm_logprobs` and `truncated_importance_sampling_ratio_cap`. "
                 "use_vllm_logprobs sets old_logprobs to vLLM logprobs, making importance sampling pointless."
             )
-        self.loss_denominator = utils.get_denominator(self.loss_denominator)
+        if self.loss_denominator != "token" and float(self.loss_denominator) <= 0:
+            raise ValueError(f"loss_denominator must be greater than 0 if not 'token', got: {self.loss_denominator}")
```
hamishivi:
```suggestion
            raise ValueError(f"loss_denominator must be a valid float greater than 0 if not 'token', got: {self.loss_denominator}")
```

PR #1403, opened 2026-01-20
Add optional wandb system metrics logging for generator process

jacob-morrison:
Add --log_generator_metrics flag to enable wandb logging from the first vLLM generator (engine_id=0) to collect GPU power, memory, and utilization metrics. When enabled, uses wandb groups to link the trainer and generator runs in the UI with -trainer and -generator suffixes.


hamishivi:
Closing since old, please re-open and ping on slack if you want this merged!

PR #1406, opened 2026-01-21
Add rollout saving to GRPO training

finbarrtimbers:
Implement the save_traces flag to save all generated RL rollouts to disk during GRPO training. For each generation, saves prompt_tokens, response_tokens, advantage, reward, step, sample_idx, prompt_idx, dataset, finish_reason, ground_truth, and tool_info.

- Add rollouts_save_path arg with default /weka/oe-adapt-default/allennlp/deletable_rollouts/
- Save metadata file with git commit, model name, and timestamp
- Save rollouts to sharded JSONL files (new file every 10k samples)
- Use async saving via ThreadPoolExecutor to not block training
- Add `RolloutMetadata` and `RolloutRecord` dataclasses for type safety
- Add GPU integration test validating rollout file creation and field presence

GPU_TESTS=[01KFGWBPSPES7P8X2BP3HGEYGC](https://beaker.org/orgs/ai2/workspaces/open-instruct-dev/work/01KFGWBPSPES7P8X2BP3HGEYGC?taskId=01KFGWBPSXVTFPET6K7XBAGAV2&jobId=01KFGWBQ1S8825KA8C6WZYQMYD)

### DIFF
--- a/open_instruct/utils.py
+++ b/open_instruct/utils.py
@@ -764,6 +764,7 @@ def get_wandb_tags() -> list[str]:
     if (git_commit := get_git_commit()) != "unknown":
         tags.append(f"commit: {git_commit}")
         try:
+            # try finding the pull request number on github
             prs = requests.get(f"https://api.github.com/search/issues?q=repo:allenai/open-instruct+is:pr+{git_commit}")
             prs.raise_for_status()
             prs = prs.json()


open_instruct/rl_utils.py
```diff
@@ -1,16 +1,164 @@
...
+from dataclasses import asdict, dataclass, field
+from typing import TYPE_CHECKING, Generic, TypeVar
 
 import numpy as np
 import torch
 
-from open_instruct import logger_utils
+from open_instruct import data_types, logger_utils, utils
+
+if TYPE_CHECKING:
```
hamishivi: do we need `if TYPE_CHECKING`?

open_instruct/rl_utils.py
```diff
@@ -1,16 +1,164 @@
...
+@dataclass
+class RolloutMetadata:
+    run_name: str
+    git_commit: str
+    model_name: str
+    timestamp: str
+
+
+@dataclass
+class RolloutRecord:
```
hamishivi, line 31: Do you think it would be reasonable to save the logprobs and/or vllm logprobs? One thing I think it would be interesting to do is being able to 'replay' the training run, which I think would need the logprobs?

hamishivi (commented):
Generally seems fine!

open_instruct/rl_utils.py
finbarrtimbers: Good idea, done!

open_instruct/rl_utils.py
finbarrtimbers: Removed! I also added a pre-commit hook to prevent these.

PR #1417, opened 2026-01-24
Add Slack notification on Beaker workflow failure

hamishivi:
## Summary

- Adds a Slack notification step to the Beaker Experiment Launch workflow that triggers only on failure
- Includes branch name, actor, commit SHA, and a button linking to the failed run

Now we can be alerted when main breaks 😆 

## Setup Required

After merging, add the `SLACK_WEBHOOK_URL` repository secret:

1. Create a Slack App at https://api.slack.com/apps
2. Enable Incoming Webhooks and add one to your target channel
3. Add the webhook URL as a repository secret named `SLACK_WEBHOOK_URL`

### DIFF
--- a/.github/workflows/beaker-experiment.yml
+++ b/.github/workflows/beaker-experiment.yml
@@ -330,22 +330,20 @@ jobs:
         run: |
           FAILED_EXPERIMENTS="${{ steps.wait.outputs.failed_experiment_names }}"
 
-          if [ "${{ steps.launch.outcome }}" = "failure" ]; then
+          # Skip Slack notification for timeouts (when experiments didn't finalize in time)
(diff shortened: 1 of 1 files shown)


.github/workflows/beaker-experiment.yml
```diff
@@ -323,3 +323,36 @@ jobs:
           else
             echo "❌ **Status:** Experiments failed or timed out (image not pushed)" >> $GITHUB_STEP_SUMMARY
           fi
+
+      - name: Notify Slack on Failure
```
finbarrtimbers, line 345: So I'm fine either way, but you might want to distinguish between a true failure and a timeout due to scheduling!

hamishivi: good point

PR #1418, opened 2026-01-24
Add automatic type coercion for tool arguments

hamishivi:
Adds safe_execute() method to Tool base class that coerces arguments to expected types based on JSON schema before calling execute(). This prevents crashes when models send wrong types (e.g. bool instead of string), and if the value can't be coerced, then we simply return the error to the model.

### DIFF
--- a/open_instruct/tools/tests/test_tools.py
+++ b/open_instruct/tools/tests/test_tools.py
@@ -31,6 +31,7 @@ from open_instruct.tools.utils import (
     ToolOutput,
     ToolsConfig,
     ToolStatistics,
+    coerce_args,
     get_openai_tool_definitions,
     make_api_request,
     truncate,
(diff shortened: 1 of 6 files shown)


open_instruct/tools/tests/test_tools.py
```diff
@@ -253,6 +253,111 @@ def test_tool_class_attribute(self):
         self.assertEqual(PythonCodeToolConfig.tool_class, PythonCodeTool)
 
 
+class TestToolArgCoercion(unittest.TestCase):
```
finbarrtimbers, line 257: Let's test behavior, not implementation! I think we could instead make tests that pass the wrong type in and verify they're caught?

open_instruct/tools/utils.py
```diff
@@ -394,14 +394,68 @@ def get_openai_tool_definitions(self) -> dict[str, Any]:
...
+            "boolean": bool,
+            "array": list,
+            "object": dict,
+        }
+
+        for name, value in kwargs.items():
+            if name not in properties:
+                continue
+            expected_type = properties[name].get("type")
+            if not expected_type or expected_type not in coercers:
```
finbarrtimbers: Why would expected_type ever be falsey?

open_instruct/tools/utils.py
```diff
@@ -394,14 +394,68 @@ def get_openai_tool_definitions(self) -> dict[str, Any]:
...
+                continue
+
+            target_type = coercers[expected_type]
+            if not isinstance(value, target_type):
+                # Special handling for boolean coercion from strings
```
finbarrtimbers: Let's make this a function and use this as the bool converter?

open_instruct/tools/utils.py
```diff
@@ -394,14 +394,68 @@ def get_openai_tool_definitions(self) -> dict[str, Any]:
         """Get tool definition in OpenAI format."""
         return get_openai_tool_definitions(self)
 
+    def _coerce_args(self, kwargs: dict[str, Any]) -> dict[str, Any]:
```
finbarrtimbers: Let's make this a free function? We only need the properties dict, right?

open_instruct/tools/utils.py
```diff
@@ -394,14 +394,68 @@ def get_openai_tool_definitions(self) -> dict[str, Any]:
...
+                    coerced[name] = target_type(value)
+
+        return coerced
+
     @abstractmethod
     async def execute(self, *args: Any, **kwargs: Any) -> ToolOutput:
         """Execute the tool, must be implemented by subclasses."""
         raise NotImplementedError("execute must be implemented by subclasses.")
 
+    async def safe_execute(self, *args: Any, **kwargs: Any) -> ToolOutput:
```
finbarrtimbers, line 452: Do we ever want to call execute vs safe_execute?

hamishivi: Good point! renamed execute to _execute. We still want it as a separate function so that subclasses don't have to worry about this arg coercion logic.

open_instruct/tools/utils.py
hamishivi: fair...

open_instruct/tools/tests/test_tools.py
hamishivi: updated!

open_instruct/tools/utils.py
```diff
@@ -395,13 +445,27 @@ def get_openai_tool_definitions(self) -> dict[str, Any]:
...
 
     @abstractmethod
-    async def execute(self, *args: Any, **kwargs: Any) -> ToolOutput:
-        """Execute the tool, must be implemented by subclasses."""
-        raise NotImplementedError("execute must be implemented by subclasses.")
+    async def _execute(self, *args: Any, **kwargs: Any) -> ToolOutput:
+        """Execute the tool. Must be implemented by subclasses."""
+        raise NotImplementedError("_execute must be implemented by subclasses.")
+
+    async def safe_execute(self, *args: Any, **kwargs: Any) -> ToolOutput:
```
finbarrtimbers, line 452: Let's just call this execute now?

PR #1425, opened 2026-01-26
Runs the benchmarks with the hybrid model

finbarrtimbers:
## Summary
- Add hybrid model (Olmo-Hybrid) support to the training pipeline
- Monkey-patch vLLM 0.18.0 `MambaSpec.dtypes` to fix hybrid model dtype serialization bug
- Forces tokens to be decoded to int, as now HF returns numpy dtypes
- Pass `trust_remote_code` through to vLLM engines for custom model/tokenizer support
- Add `min_tokens` to `SamplingConfig` with `extra_body` pass-through for vLLM API compat
- Upgrade dependencies: vllm>=0.18.0, transformers>=5.3.0, torch upper bound removal
- Add `return_dict=False` to all `apply_chat_template` calls for transformers 5.x compat
- Remove `configure_hf_hub_retry()` — the underlying `huggingface_hub.configure_http_backend()` was removed in huggingface_hub v1.0.0 (we now pull in v1.7.2 via the vllm/transformers upgrade). HF Hub v1.0+ has built-in retry logic via `http_backoff()` with the same 429/5xx exponential backoff behavior, and the pre-download fix from PR #1528 remains in place
- Add hybrid test scripts (GRPO + DPO) and production training scripts

## Runs

1. Multi-node GRPO: [Beaker](https://beaker.org/ex/01KMNGSSB8HAR6HKS1136BJFZH)
2. DPO 2-node: [Beaker](https://beaker.org/ex/01KMNDJTGBJRBPFEKS30RB1CD2)
3. Single GPU benchmark: [Beaker](https://beaker.org/ex/01KMND6YZ34FRTQ50BJXQP3J8W)

GPU_TESTS=[01KMNGSSB8HAR6HKS1136BJFZH](https://beaker.org/ex/01KMNGSSB8HAR6HKS1136BJFZH)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -56,13 +56,16 @@ jobs:
             "${{ github.repository }}" \
             "$PR_BODY"
 
+      - name: Setup Beaker (for override verification)
+        if: env.SKIP_REASON == 'override'
+        uses: allenai/setup-beaker@v2
+        with:
+          token: ${{ secrets.BEAKER_TOKEN }}
+          workspace: ai2/open-instruct-dev
+
       - name: Verify GPU_TESTS override experiment
         if: env.SKIP_REASON == 'override'
-        env:
-          BEAKER_TOKEN: ${{ secrets.BEAKER_TOKEN }}
-        run: |
-          echo "$BEAKER_TOKEN" | beaker account login --token-stdin
-          ./scripts/test/verify_gpu_tests_override.sh "${{ env.GPU_TESTS_EXP_ID }}"
+        run: ./scripts/test/verify_gpu_tests_override.sh "${{ env.GPU_TESTS_EXP_ID }}"
 
       - name: Checkout code
         if: env.RUN_GPU_TESTS == 'true'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 
 ### Added
+- Add hybrid model (Olmo-Hybrid) support: MambaSpec monkey-patch for vLLM dtype serialization, `trust_remote_code` pass-through to vLLM engines, `get_text_config()` for multimodal model support, dependency upgrades (vllm>=0.18.0, transformers>=5.3.0), `return_dict=False` for transformers 5.x compat, and hybrid test/production training scripts (https://github.com/allenai/open-instruct/pull/1425).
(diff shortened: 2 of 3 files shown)


open_instruct/vllm_utils.py
```diff
@@ -77,6 +78,27 @@
...
+#
+# The client-side decoder (in serial_utils._convert_result) dynamically imports
+# MambaSpec and calls msgspec.convert(data, MambaSpec), which checks the live
+# type annotations. By widening dtypes to tuple[torch.dtype, ...] (variable-
+# length), msgspec accepts any number of dtypes during deserialization.
+#
+# Upstream fix: change dtypes annotation in vllm/v1/kv_cache_interface.py.
+# ---------------------------------------------------------------------------
+MambaSpec.__dataclass_fields__["dtypes"].type = tuple[torch.dtype, ...]
+MambaSpec.__annotations__["dtypes"] = tuple[torch.dtype, ...]
```
hamishivi, line 100: Do we have an open PR to fix this on vllm side that we can track here?

open_instruct/vllm_utils.py
```diff
@@ -600,7 +623,14 @@ def __init__(
         self._setup_gpu_visibility(noset_visible_devices, distributed_executor_backend)
         self._setup_and_start_async_engine(args, bundle_indices, kwargs)
         self._init_openai_client()
-        self.inference_batch_size = self.get_kv_cache_info()
+        try:
+            self.inference_batch_size = self.get_kv_cache_info()
+        except Exception:
```
hamishivi: do we need the try catch (and fallback inference size) with the monkey-patch?

scripts/debug/repro_vllm_hybrid_dtype.py
```diff
@@ -0,0 +1,35 @@
+"""Minimal repro for vLLM hybrid model dtype serialization bug.
```
hamishivi: do we need to commit this file?

finbarrtimbers: No, I meant to remove this but didn't push.

open_instruct/vllm_utils.py
finbarrtimbers: As discussed in Slack, no.

open_instruct/vllm_utils.py
finbarrtimbers: Good question. Let me try removing it.

finbarrtimbers: It's fine! [Beaker](https://beaker.org/orgs/ai2/workspaces/open-instruct-dev/work/01KMNQE0P9A8MWXXNR34D2VENW).

PR #1434, opened 2026-01-28
Add documentation and warning for dataset_mixer_list format

natolambert:
## Summary
- Add docstring to `load_dataset_configs()` explaining the format
- Add inline comment at the parsing line  
- Add runtime warning when integer count <= 10 and dataset > 100 samples
- Add format documentation to tokenization script header

The dataset mixer uses a non-obvious convention where:
- Float values (with decimal): proportion of dataset (`"1.0"` = 100%)
- Integer values (no decimal): absolute sample count (`"1"` = 1 sample)

This is a common source of errors when users write `"1"` expecting the full dataset but get only 1 sample.

## Test plan
- [ ] Verify warning triggers with small integer counts
- [ ] Verify no warning with valid float proportions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 
 
 ### Added
-- Documentation and runtime warning for `dataset_mixer_list` format (float=proportion, int=count).
+- Documentation and runtime warning for `dataset_mixer_list` format (float=proportion, int=count) (https://github.com/allenai/open-instruct/pull/1434).
 
 ### Changed
 


open_instruct/dataset_transformation.py
```diff
@@ -1940,6 +1940,32 @@ def load_dataset_configs(
     target_columns: list[str] | None = None,
     dataset_config_seed: int = 42,
 ) -> list[DatasetConfig]:
+    """
+    Load and configure datasets from a mixer list.
+
+    Args:
+        dataset_mixer_list: Alternating list of [dataset_name, amount, dataset_name, amount, ...].
+            The 'amount' value determines how many samples to use:
```
finbarrtimbers, line 1948: This is so confusing, we should fix this

natolambert: Yeah I'm not opposed. We can leave this PR open and add it to it later, or merge and come back to it.

finbarrtimbers: Nah, let's do it later.

PR #1439, opened 2026-01-29
Fix eval responses lost on timeout in grpo_fast

### LINKED ISSUE #1385: Eval responses lost on timeout in grpo_fast
## Problem

In `grpo_fast.py`, `maybe_evaluate` uses a 10ms timeout for eval results. When this timeout hits:

1. **Results already pulled from queue are discarded** - `accumulate_inference_batches` stores them in local variables, but the `Empty` exception causes them to be lost
2. **Stale results accumulate** - Results that complete later stay in the queue and may be consumed on subsequent eval steps, mixing metrics from different model checkpoints

In practice, this usually means when eval responses take a while to generate, we never see eval responses.

## Location

- `grpo_fast.py:1917` - 0.01s timeout
- `data_loader.py:501` - `.get(timeout=timeout)` in loop that discards partial results on exception

## Possible fixes

This is tricky to precisely fix:

1. We could run the eval generations, and pause training/weight updates until they come back in. This would give us precise knowledge that step X has eval score Y. However, it would slow down training, especially if the samples take too long.
2. We could just allow evals to accumulate **across** steps. If the eval samples take really, really long, this could be a bit weird, since the model is changing as we run them, but I think this should be fine...?

Personally, I vote for (2).

hamishivi:
Fixes #1385. Now, we accumulate eval results across training steps. If we hit a new eval step but haven't yet cleared out partial results, we log a warning and just flush it out without logging. 

The only thing I'm unsure about: the training step associated with the logging is the training step we actually log at. This is technically wrong since the eval samples can be from whenever, but I think its better than pausing all training for the eval samples. The main use case is tracking performance, so I think it should still work (we want to observe that more training steps -> eval scores increasing).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 
 ### Added
+- NVIDIA H200 GPU support in `GPU_SPECS` (https://github.com/allenai/open-instruct/pull/1441).
 - Documentation and runtime warning for `dataset_mixer_list` format (float=proportion, int=count) (https://github.com/allenai/open-instruct/pull/1434).
 
 ### Changed
@@ -12,6 +13,7 @@ All notable changes to this project will be documented in this file.
 - Changed default wandb x-axis from `episode` to `training_step` for grpo_fast (https://github.com/allenai/open-instruct/pull/1437).
 
 ### Fixed
+- Fixed evaluation responses being lost on timeout in grpo_fast by requeuing partial results (https://github.com/allenai/open-instruct/pull/1439).
 - Beaker Experiment Launch now passes (https://github.com/allenai/open-instruct/pull/1424#pullrequestreview-3708034780).
 
 ### Removed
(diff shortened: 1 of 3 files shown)


open_instruct/grpo_fast.py
finbarrtimbers, line 1: is there a cleaner way of handling this? could we, for instance, check the number of results, and see if they're all done? or could we set a flag to handle this? 

There's just a lot of moving pieces in this PR!

open_instruct/grpo_fast.py
```diff
@@ -1727,6 +1727,7 @@ def maybe_evaluate(
             active_sampling=False,
             filter_zero_std_samples=False,
             replenish_prompts=False,
+            requeue_on_timeout=True,
```
finbarrtimbers: Should this just always be true?

finbarrtimbers (approved):

open_instruct/grpo_fast.py
hamishivi: Yeah fair. I'll leave it as an option but default it to true.

PR #1445, opened 2026-01-30
Add torch.compile support with max_length padding

finbarrtimbers:
Adds a `--compile_model` flag to `dpo.py`. 

It doesn't help performance at all, but it does let us use gradient checkpointing, which will be needed for large models ([Wandb](https://wandb.ai/ai2-llm/open_instruct_internal?nw=wiz8qvv5ze)):

Screenshot 2026-02-02 at 2 23 33 PM

Runs: 
1. Single GPU (01KGG5MAQ9ENC1Y7ARMN6BQF6W)
2. Multi-node (01KGG5RV1BER13M28XZXCJXS9Z)

In an earlier version of this PR, to avoid recompiles, we added a flag to pad the sequence lengths to have a fixed size. We ran an ablation comparing compilation with max length padding ([Beaker](https://beaker.org/ex/01KGFZEC4TERMHM4RQE9PFEQXF)) to compilation without ([Beaker](https://beaker.org/ex/01KGFZE980GW3TTAT0FWH6B819)), and found that max length padding hurts performance, but avoids recompiles:

| Metric | [With Padding ](https://beaker.org/ex/01KGFZEC4TERMHM4RQE9PFEQXF)| [Without Padding](https://beaker.org/ex/01KGFZE980GW3TTAT0FWH6B819) |
  |--------|--------------|-----------------|
  | MFU (avg) | 4.58% | 11.85% |
  | Seconds/step | 1.04s | 0.42s |
  | Recompilations | 48 | 396 |

As such, we removed the padding.

### DIFF
--- a/open_instruct/olmo_core_train_modules.py
+++ b/open_instruct/olmo_core_train_modules.py
@@ -127,6 +127,7 @@ class DPOTrainModule(TrainModule):
             self.record_metric("train/logps_chosen", policy_chosen_logps.mean().detach(), ReduceType.mean)
             self.record_metric("train/logps_rejected", policy_rejected_logps.mean().detach(), ReduceType.mean)
             token_count = self.trainer.data_loader.global_num_tokens_in_batch(batch)
+            assert token_count is not None
             self.record_metric("train/token_count", token_count, reduce_type=None)
 
             if self.args.loss_type.computes_reward_metrics:


open_instruct/dpo_utils.py
```diff
@@ -1045,26 +1050,11 @@ def separate_forward_olmo(
...
-        length: The target length for the specified dimension.
-        pad_value: The value to use for padding.
-        dim: The dimension along which to pad.
-
-    Returns:
-        The padded tensor, or the original tensor if already at least the target length.
-    """
-    if tensor.size(dim) >= length:
+def pad_to_length(tensor: torch.Tensor, length: int, pad_value: int | float) -> torch.Tensor:
+    """Pad a tensor to a specified length along the last dimension."""
```
hamishivi:
```suggestion
    """Right-pad a tensor to a specified length along the last dimension."""
```

open_instruct/dpo.py
```diff
@@ -308,6 +312,13 @@ def main(args: dpo_utils.ExperimentConfig, tc: dataset_transformation.TokenizerC
     if args.use_lora:
         raise ValueError("LoRA is not supported with OLMo-core DPO training. Use dpo_tune_cache.py instead.")
 
+    if args.packing and args.compile_model:
+        raise ValueError(
+            "packing and compile_model cannot be used together. "
+            "Packing creates variable-length batches which causes torch.compile to recompile on every batch. "
```
hamishivi, line 322: This is kinda surprising to me? I thought you could use varlen_attn to avoid torch.compile doing this (https://docs.pytorch.org/tutorials/intermediate/variable_length_attention_tutorial.html).

finbarrtimbers: Maybe it's bc we use flash attention?

PR #1447, opened 2026-01-30
Add gradient accumulation/microbatching support to `DPOTrainModule`

finbarrtimbers:
Adds gradient accumulation to `DPOTrainModule`. 

Runs
- Single GPU DPO: [Beaker](https://beaker.org/ex/01KGGWJ67GZHZ59W79GHDSM2SD) | [WandB](https://wandb.ai/ai2-llm/open_instruct_internal/runs/pj2n8gmj)
- Multi-node DPO (2x8 GPUs): [Beaker](https://beaker.org/ex/01KGGX5K0Y9S89RBRCPP8JFDA5) | [WandB](https://wandb.ai/ai2-llm/open_instruct_internal/runs/fsv3l9nf)


open_instruct/olmo_core_train_modules.py
```diff
@@ -71,65 +88,107 @@ def __init__(
...
+
+        for v in self._metrics.values():
+            v.zero_()
+
+        for micro_batch_idx, micro_batch in enumerate(micro_batches):
+            with self._train_microbatch_context(micro_batch_idx, num_micro_batches):
+                loss, step_metrics = self._compute_microbatch_loss(micro_batch)
+                for k, v in step_metrics.items():
+                    self._metrics[k] += v.detach()
+                (loss / num_micro_batches).backward()
```
hamishivi: I think we need to divide by total tokens in the batch, rather than the number of microbatches. This is because each microbatch is weighted evenly even though they have differing numbers of tokens in them.

open_instruct/olmo_core_train_modules.py
```diff
@@ -71,65 +88,107 @@ def __init__(
...
-                accuracy = (chosen_rewards > rejected_rewards).float().mean()
-                margin = (chosen_rewards - rejected_rewards).mean()
-                self.record_metric("train/rewards_chosen", chosen_rewards.mean().detach(), ReduceType.mean)
-                self.record_metric("train/rewards_rejected", rejected_rewards.mean().detach(), ReduceType.mean)
-                self.record_metric("train/rewards_accuracy", accuracy.detach(), ReduceType.mean)
-                self.record_metric("train/rewards_margin", margin.detach(), ReduceType.mean)
-
-            if self.dpo_config.load_balancing_loss and aux_loss is not None:
-                self.record_metric("train/aux_loss", aux_loss.detach(), ReduceType.mean)
+                margin = self._metrics["chosen_rewards"] - self._metrics["rejected_rewards"]
```
hamishivi: same as above for here - if we log every optimization step, then the metrics should be normalized by the total num tokens in the batch, which is not equivalent to averaging across the microbatches.

open_instruct/olmo_core_train_modules.py
finbarrtimbers: Nice catch, done.

open_instruct/olmo_core_train_modules.py
finbarrtimbers: Nice catch, done.

open_instruct/olmo_core_train_modules.py
```diff
@@ -71,65 +88,110 @@ def __init__(
...
+                    self._metrics[k] += v.detach() * weight
+                (loss * weight).backward()
+
+        self.model.post_batch(dry_run=dry_run)
+
         if not dry_run:
-            self.record_metric("train/loss", loss.detach(), ReduceType.mean)
-            self.record_metric("train/logps_chosen", policy_chosen_logps.mean().detach(), ReduceType.mean)
-            self.record_metric("train/logps_rejected", policy_rejected_logps.mean().detach(), ReduceType.mean)
+            self.record_metric("train/loss", self._metrics["loss"], ReduceType.mean)
```
hamishivi: should reduceType be mean? shouldnt be sum since we accounted for the token weighting above?

open_instruct/olmo_core_train_modules.py
```diff
@@ -71,65 +88,110 @@ def __init__(
...
+                loss, step_metrics = self._compute_microbatch_loss(micro_batch)
+                micro_tokens = (
+                    micro_batch["chosen_attention_mask"].sum() + micro_batch["rejected_attention_mask"].sum()
+                ).float()
+                weight = micro_tokens / total_tokens
+                for k, v in step_metrics.items():
+                    self._metrics[k] += v.detach() * weight
+                (loss * weight).backward()
+
+        self.model.post_batch(dry_run=dry_run)
```
hamishivi, line 170: is this where the optim step is taken?

open_instruct/olmo_core_train_modules.py
finbarrtimbers: ah yeah, nice catch. annoyingly Olmo-core doesn't provide a `ReduceType.weighted_mean`. So I had to roll my own.

open_instruct/olmo_core_train_modules.py
finbarrtimbers: No, it happens after, in the `TransformerTrainModule.optim_step` method. See here: 

https://github.com/allenai/OLMo-core/blob/3af842521375a373266673dda262debe0748a462/src/olmo_core/train/trainer.py#L1419

hamishivi (approved):

PR #1454, opened 2026-02-02
Fix training hang when inflight_updates is disabled

hamishivi:
## Summary

Fixes a training hang that occurs when `inflight_updates` is disabled (or not using vLLM's inflight batching).

**Root cause:** When `inflight_updates` is off, weight sync to vLLM engines is a blocking operation. During this time, vLLM engines cannot respond to other Ray remote calls. The health check at the start of each training step tries to call `check_background_threads.remote()` on vLLM engines, but they're blocked by the weight broadcast from the previous step, causing an indefinite hang.

For some reason, it seems that they never manage to return on this even once the weight broadcast is finished. So now I just added a stop polling for the weight sync manager. This fixes the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ All notable changes to this project will be documented in this file.
 - Changed default wandb x-axis from `episode` to `training_step` for grpo_fast (https://github.com/allenai/open-instruct/pull/1437).
 
 ### Fixed
+- Fixed training hang when `inflight_updates` is disabled by waiting for weight sync to complete before health check (https://github.com/allenai/open-instruct/pull/1454).
 - Fixed evaluation responses being lost on timeout in grpo_fast by requeuing partial results (https://github.com/allenai/open-instruct/pull/1439).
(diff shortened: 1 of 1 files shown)


open_instruct/grpo_fast.py
```diff
@@ -1847,6 +1848,12 @@ def run_training(
 
     def health_check_fn():
         [f.result() for f in [weight_sync_thread_future] if f.done()]
+        # Wait for weight sync to complete (should_stop becomes False)
```
finbarrtimbers, line 1851: Ray does support retries fyi:

# Retries up to 5 times on system errors, and retries if a ValueError is raised
@ray.remote(max_retries=5, retry_exceptions=[ValueError])
def possibly_failing_task(attempt_number):
    if attempt_number < 3:
        print(f"Attempt {attempt_number}: Failing with ValueError")
        raise ValueError("Simulated transient error")
    else:
        print(f"Attempt {attempt_number}: Success!")
        return "success"

Maybe we use that?

PR #1458, opened 2026-02-03
Refactor DPO to let TransformerTrainModule handle parallelism

finbarrtimbers:
Refactors `DPOTrainModule` to let `TransformerTrainModule` handle parallelism.

Disables TP and CP for DPO with `NotImplementedError` guards as they're broken currently:
  - **TP**: DTensor view ops are incompatible with `torch.compile` (we need to pad to a fixed size)
  - **CP**: Requires `batch_shard_by_document()` integration in the DPO training loop

Runs: 

- Single GPU DPO: [Beaker](https://beaker.org/ex/01KGJY3CQVY17FEMWNYDF281ET)
- Multi-node DPO: [Beaker](https://beaker.org/ex/01KGNE96G7KBHQ44F40S2HXV1D)

### DIFF
--- a/open_instruct/dpo.py
+++ b/open_instruct/dpo.py
@@ -378,7 +378,8 @@ def main(args: dpo_utils.ExperimentConfig, tc: dataset_transformation.TokenizerC
     )
 
     # Build reference cache after train_module init because TransformerTrainModule applies
-    # FSDP parallelism to the model, and we need the parallelized model for correct logprobs.
+    # FSDP parallelism to the model, and we need the parallelized model to calculate the
+    # logprobs in case the model is too big to fit in memory.
     logger.info("Caching reference logprobs...")
     train_module.reference_cache = dpo_utils.build_reference_logprobs_cache(model=train_module.model, **cache_kwargs)
 


open_instruct/dpo.py
```diff
@@ -304,6 +220,22 @@ def main(args: dpo_utils.ExperimentConfig, tc: dataset_transformation.TokenizerC
             "Either disable packing or disable compile_model."
         )
 
+    if args.tensor_parallel_degree > 1:
+        raise NotImplementedError(
+            "Tensor parallelism is not supported with DPO (DTensor view ops are incompatible with torch.compile)."
```
hamishivi, line 225: 😭😭😭😭😭😭😭

open_instruct/dpo.py
```diff
@@ -392,7 +323,9 @@ def main(args: dpo_utils.ExperimentConfig, tc: dataset_transformation.TokenizerC
         drop_last=False,
     )
 
-    forward_fn = dpo_utils.concatenated_forward_olmo if args.concatenated_forward else dpo_utils.separate_forward_olmo
+    forward_fn = partial(
```
hamishivi: why does this need to be a partial?

open_instruct/dpo.py
```diff
@@ -410,50 +343,55 @@ def main(args: dpo_utils.ExperimentConfig, tc: dataset_transformation.TokenizerC
...
+        ac_config=ac_config,
+        compile_model=args.compile_model,
         max_grad_norm=max_grad_norm,
         scheduler=scheduler,
         device=device,
     )
 
-    trainer_callbacks = _setup_callbacks(args, model, dp_world_size)
+    # Build reference cache after train_module init because TransformerTrainModule applies
+    # FSDP parallelism to the model, and we need the parallelized model for correct logprobs.
```
hamishivi: oh thats nasty. There's differents in logprobs based on sharding?

open_instruct/olmo_core_train_modules.py
```diff
@@ -65,16 +72,19 @@ def __init__(
...
 
         self.dpo_config = dpo_config
-        self.reference_cache = reference_cache
+        self.reference_cache: model_utils.TensorCache | None = None
 
         if dpo_config.packing:
             self._forward_fn = partial(dpo_utils.concatenated_forward_olmo, packing=True)
         elif dpo_config.concatenated_forward:
-            self._forward_fn = dpo_utils.concatenated_forward_olmo
+            self._forward_fn = partial(dpo_utils.concatenated_forward_olmo)
```
hamishivi: again, why partial here and below?

open_instruct/dpo.py
```diff
@@ -410,50 +343,55 @@ def main(args: dpo_utils.ExperimentConfig, tc: dataset_transformation.TokenizerC
...
+        param_dtype=DType.bfloat16,
+        reduce_dtype=DType.float32,
+        wrapping_strategy=transformer_config.TransformerDataParallelWrappingStrategy.blocks,
+    )
+    ac_config = (
+        transformer_config.TransformerActivationCheckpointingConfig(
+            mode=transformer_config.TransformerActivationCheckpointingMode.budget,
+            activation_memory_budget=args.activation_memory_budget,
+        )
+        if args.activation_memory_budget < 1.0 and args.compile_model
```
hamishivi, line 362: is there a good default for `activation_memory_budget` just to 'turn it on'?

finbarrtimbers: I'd say 0.5 is good. It's kinda arbitrary though, I'm just pulling that number out of thin air. Anything <1 will enable it.

finbarrtimbers: Higher numbers use more memory and are faster, lower numbers use less memory and are slower. So you want the highest possible number!

open_instruct/dpo.py
finbarrtimbers: Tell me about it 😿

open_instruct/dpo.py
finbarrtimbers: No, it's bc we need the calculate the logprobs with the sharded model in case it's too big to fit in memory. The comment is misleading. I fixed it.

open_instruct/olmo_core_train_modules.py
finbarrtimbers: Removed. This was from an earlier version of the code.

open_instruct/dpo.py
hamishivi: Could we add this statement to the arg comment?

hamishivi (approved):
LGTM!

PR #1461, opened 2026-02-04
Warn when eval prompts are queuing up

hamishivi:
## Summary
- Track whether the previous eval round was successfully collected
- When starting a new eval round before the previous one completes, use vLLM's `abort()` API to stop GPU work for stale requests
- Log a warning with the count of aborted requests and suggest increasing `local_eval_every`

## Test plan
- [ ] Run a training job with frequent eval (`local_eval_every=1`) and verify stale requests are aborted
- [ ] Check logs for the warning message when evals are cancelled

### DIFF
--- a/.claude/skills/update-pr-body/SKILL.md
+++ b/.claude/skills/update-pr-body/SKILL.md
@@ -6,6 +6,10 @@ allowed-tools: Bash(gh:*)
 
 # Update GitHub PR Body
 
+## Important
+
+**Do NOT use `gh pr edit --body`** - it fails with a GraphQL error about Projects (classic) deprecation. Always use the REST API approach below instead.
+
 ## Instructions
 
 When updating a GitHub PR body:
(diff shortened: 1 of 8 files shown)


open_instruct/vllm_utils.py
```diff
@@ -849,6 +849,31 @@ def check_background_threads(self) -> None:
...
+
+        eval_task_ids = [k for k in self.active_tasks if k.startswith("eval_")]
+        for task_id in eval_task_ids:
+            try:
+                self._run_async(self.llm_engine.abort(task_id))
+                aborted_count += 1
+            except Exception as e:
+                logger.debug(f"Failed to abort eval request {task_id}: {e}")
+
+            future = self.active_tasks.pop(task_id, None)
```
finbarrtimbers: what about: 

if (future := self.active_tasks.pop(task_id, None)) is not None:
    future.cancel()

open_instruct/vllm_utils.py
```diff
@@ -849,6 +849,31 @@ def check_background_threads(self) -> None:
...
+            if future is not None:
+                future.cancel()
+        eval_base_ids = [k for k in self.request_metadata if k.startswith("eval_")]
```
finbarrtimbers: Let's just combine the comprehension into the for-loop!

PR #1466, opened 2026-02-06
Enable packing + compile for DPO with wasted tokens metric

finbarrtimbers:
## Summary
- Enable packing + torch.compile for DPO training to reduce padding waste
- Fix cu_seq_lens offset bug when concatenating padded chosen/rejected sequences
- Add `tokens_per_second_per_gpu` metric to wandb

## DPO Packing Performance Results (post MFU-fix)

All runs on `finbarr/profile-dpo` branch, 2 nodes x 8 H100 GPUs, OLMo2-7B, 16k seq len, `activation_memory_budget=0.1`.

### Batch size comparison (no shard_degree, 16-way FSDP)

| Batch Size | Token Count | MFU (last) | TPS | TPS/GPU | sec/step | Experiment |
|------------|-------------|------------|-----|---------|----------|------------|
| bs=2 | 57,664 | 2.46% | 9,276 | 580 | 6.21s | [01KGSYNQ5RYZD1SVWWYDTQNY49](https://beaker.org/ex/01KGSYNQ5RYZD1SVWWYDTQNY49) |
| bs=4 | 58,640 | 2.51% | 9,465 | 592 | 6.19s | [01KGT1DNDKHZXR1PM2D3ADHD5M](https://beaker.org/ex/01KGT1DNDKHZXR1PM2D3ADHD5M) |
| bs=8 | 431,120 | 19.33% | 69,611 | 4,351 | 6.19s | [01KGT22ECAJPVPVQP5BRAAA9Q1](https://beaker.org/ex/01KGT22ECAJPVPVQP5BRAAA9Q1) |
| bs=16 | 453,184 | 19.71% | 73,067 | 4,567 | 6.19s | [01KGTEY8SZ3CWKHBX2KXD7HHAB](https://beaker.org/ex/01KGTEY8SZ3CWKHBX2KXD7HHAB) |
| bs=32 | 524,288 | 22.48% | 84,513 | 5,282 | 6.19s | [01KGTE43M9SBGD9D1Y7Z010EZ3](https://beaker.org/ex/01KGTE43M9SBGD9D1Y7Z010EZ3) |

### shard_degree=8 vs no shard_degree (bs=8, full epoch, 57 steps after warmup)

| Config | MFU | TPS | TPS/GPU | sec/step | Token Count | Experiment |
|--------|-----|-----|---------|----------|-------------|------------|
| no shard (16-way FSDP) | 8.82% ± 2.54 | 32,608 ± 9,141 | 2,038 ± 571 | 4.212 ± 0.010 | 233,151 ± 59,605 | [01KH1S31HR9Y6BRN8MTNY50PV3](https://beaker.org/ex/01KH1S31HR9Y6BRN8MTNY50PV3) / [wandb](https://wandb.ai/ai2-llm/open_instruct_internal/runs/tv9jy94p) |
| shard=8, replicas=2 (8-way FSDP) | 8.84% ± 2.55 | 32,670 ± 9,166 | 2,042 ± 573 | 4.210 ± 0.009 | 233,151 ± 59,605 | [01KH1QG1HYHEYVHRFBXC79MYA3](https://beaker.org/ex/01KH1QG1HYHEYVHRFBXC79MYA3) / [wandb](https://wandb.ai/ai2-llm/open_instruct_internal/runs/7nbk9jb9) |

**Result**: shard_degree=8 and no shard_degree produce **identical performance**. Over a full epoch (57 steps after warmup), all metrics are within noise. For DPO with 2 nodes, shard_degree has no effect — the default 16-way FSDP is fine.

### Binary search: activation_memory_budget with bs=4

Tested both with shard_degree=8 and without shard_degree (16-way FSDP). All budgets > 0.1 OOM on the same 12.25 GiB logits allocation from DPO's concatenated forward pass.

**With shard_degree=8 (8-way FSDP):**

| Budget | Result | Experiment |
|--------|--------|------------|
| 0.5 | OOM (688 MiB alloc failed) | [01KH1JHGY6STAF3GCCJTEPTWQS](https://beaker.org/ex/01KH1JHGY6STAF3GCCJTEPTWQS) |
| 0.3 | OOM (12.25 GiB alloc failed) | [01KH1K8WYQ2PA3Y2NZBG2GKPEF](https://beaker.org/ex/01KH1K8WYQ2PA3Y2NZBG2GKPEF) |
| 0.2 | OOM (12.25 GiB alloc failed) | [01KH1KRZ8J56RM81A0RNK6SQJA](https://beaker.org/ex/01KH1KRZ8J56RM81A0RNK6SQJA) |

**Without shard_degree (16-way FSDP):**

| Budget | Result | Experiment |
|--------|--------|------------|
| 0.5 | OOM (688 MiB alloc failed) | [01KH1VFZB95VQF38A35HA7HHPV](https://beaker.org/ex/01KH1VFZB95VQF38A35HA7HHPV) |
| 0.3 | OOM (12.25 GiB alloc failed) | [01KH1W3HV9DHGV87X8J6AK1GT8](https://beaker.org/ex/01KH1W3HV9DHGV87X8J6AK1GT8) |
| 0.2 | OOM (12.25 GiB alloc failed) | [01KH1WJ86T4Q3D9RNH171ZPY49](https://beaker.org/ex/01KH1WJ86T4Q3D9RNH171ZPY49) |
| 0.15 | OOM (12.25 GiB alloc failed) | [01KH1X0C2YTCR7K4HT5E9RJ95S](https://beaker.org/ex/01KH1X0C2YTCR7K4HT5E9RJ95S) |

**Conclusion**: The OOM bottleneck is the logits tensor in DPO's concatenated forward pass (~12.25 GiB = 4 seqs × 2 × 16384 tokens × 152064 vocab × 2 bytes). This allocation is independent of activation checkpointing, so increasing `activation_memory_budget` beyond 0.1 always fails for bs=4 at 16k seq len. The budget only controls how much recomputation vs caching is used for intermediate activations — it cannot reduce the logits tensor size.

### Analysis

- **Step time is constant (~6.19s) across all batch sizes** (without shard_degree) — the GPU does the same work regardless; the difference is in how many tokens are actually processed per step.
- **bs=2/4 waste ~87% of tokens** due to poor packing utilization at 16k seq len (only ~58k of 524k theoretical max tokens).
- **bs=32 hits 100% packing utilization** (524,288 = 16 GPUs × 16,384 tokens × 2 chosen+rejected).
- **bs=16 is the sweet spot** — 86% packing utilization with minimal sequence truncation (1-4 seqs dropped).
- **shard_degree=8 has no effect on throughput** — full-epoch comparison shows identical MFU/TPS/sec_per_step.
- **MFU and TPS are linearly consistent** post-fix (MFU/TPS ratio ~0.000265 across all batch sizes), confirming the sequence counting fix works correctly.
- **Note**: pre-fix runs (commits before 555d9560) reported inflated MFU/TPS because `token_count` was computed from config (theoretical max) rather than actual tokens processed.

## Bug fixes:

1. **Fixed cu_seq_lens offset bug** (`padding_free_collator.py`): When concatenating padded chosen/rejected sequences, the offset for rejected cu_seq_lens now correctly uses the padded length instead of actual sequence length.

2. **Fixed MFU calculation** (`olmo_core_callbacks.py`): Count actual sequences from batch cu_seq_lens instead of deriving from config, which overcounted when packing truncates sequences.

## Pre-filter DPO packing (no more mid-stream truncation)

When DPO packing is enabled, the collator now greedily packs as many **complete** examples as fit within `max_seq_length`, rather than packing all examples then truncating mid-stream. This prevents data loss from cutting sequences partway through.

Also fixed the reference cache data loader to use `batch_size=1` per rank when packing, ensuring no examples are dropped during cache building.

Runs:
1. Single GPU DPO: [Beaker](https://beaker.org/ex/01KHEJMG3HJ1MP6S3K0KMRGB1M)
2. Multi-node DPO: [Beaker](https://beaker.org/ex/01KH9RZD11EPNFEVJPZWT1A32G)

### DIFF
--- a/open_instruct/olmo_core_callbacks.py
+++ b/open_instruct/olmo_core_callbacks.py
@@ -184,8 +184,9 @@ class PerfCallback(Callback):
         tokens_per_second_avg = self._total_tokens_processed / total_time_elapsed
 
         logging_steps = self.trainer.metrics_collect_interval
-        num_sequences = self._interval_num_sequences
-        avg_sequence_length = total_tokens_step / num_sequences if num_sequences > 0 else 0
+        avg_sequence_length = (
+            total_tokens_step / self._interval_num_sequences if self._interval_num_sequences > 0 else 0
+        )
 
         mfu_result = self.model_dims.approximate_learner_utilization(
             total_tokens=total_tokens_step,


open_instruct/data_loader.py
```diff
@@ -272,7 +272,14 @@ def global_num_tokens_in_batch(self, batch: dict[str, Any]) -> int:
         """
         # attention_mask is 1 for all non-padding tokens, and 0 otherwise.
         # Using sum() excludes padding tokens from the count.
-        num_tokens = batch["attention_mask"].sum().item()
+        if "attention_mask" in batch:
+            num_tokens = batch["attention_mask"].sum().item()
+        elif "chosen_cu_seq_lens_k" in batch:
```
hamishivi: maybe its good to comment what the cu_seq_len vectors are/come from?

open_instruct/dpo.py
```diff
@@ -311,7 +310,10 @@ def main(args: dpo_utils.ExperimentConfig, tc: dataset_transformation.TokenizerC
         device=device,
     )
     # 4x batch size: forward-only (no backward), so no activation storage needed.
-    cache_batch_size = args.per_device_train_batch_size * 4 * dp_world_size
+    # With packing, use smaller batch size (half) to avoid truncation when packing long sequences.
```
hamishivi: What happens if per_device_train_batch_size is 1? It's not an unlikely use case

open_instruct/padding_free_collator.py
```diff
@@ -1,8 +1,125 @@
...
+        labels = labels[:, :max_seq_length]
+        if position_ids is not None:
+            position_ids = position_ids[:, :max_seq_length]
+        if seq_idx is not None:
+            seq_idx = seq_idx[:, :max_seq_length]
+
+        num_valid_seqs = num_features
+        sequences_dropped = 0
+        if cu_seq_lens is not None:
+            cu_seq_lens = torch.unique_consecutive(cu_seq_lens.clamp_max_(max_seq_length))
```
hamishivi: why do we need unique_consecutive?

open_instruct/padding_free_collator.py
```diff
@@ -1,8 +1,125 @@
...
+    input_ids: torch.Tensor
+    labels: torch.Tensor
+    position_ids: torch.Tensor | None
+    seq_idx: torch.Tensor | None
+    cu_seq_lens: torch.Tensor | None
+    num_valid_seqs: int
+    sequences_dropped: int
+
+
+def _fit_to_max_length(
```
hamishivi: can we add a doc string to this function? its a little dense

open_instruct/olmo_core_callbacks.py
```diff
@@ -146,18 +147,27 @@ def pre_train(self) -> None:
...
         self._prev_wall_clock_step_start = self._wall_clock_step_start
         self._wall_clock_step_start = self._batch_load_start
 
     def pre_step(self, batch: dict[str, Any]) -> None:
-        del batch
         self._batch_load_time = time.perf_counter() - self._batch_load_start
         self._prev_pre_step_time = self._pre_step_time
         self._pre_step_time = time.perf_counter()
         self._step_start_time = self._pre_step_time
+        if "chosen_cu_seq_lens_k" in batch:
```
hamishivi: I sort of dont like all these conditionals around dict entries, but not clear enough on the overall design to know if its easy to fix. Ideally we have more consistent makeups of the batch dicts, esp for DPO?

open_instruct/dpo_utils.py
```diff
@@ -1022,7 +1040,7 @@ def concatenated_forward_olmo(
         all_logps = pf_get_batch_logps(
             logits,
             concatenated_batch["concatenated_labels"],
-            concatenated_batch["concatenated_cu_seq_lens"],
+            concatenated_batch["concatenated_cu_seq_lens_k"],
```
hamishivi, line 1042: why the rename?

open_instruct/data_loader.py
finbarrtimbers: Added a comment!

open_instruct/dpo.py
finbarrtimbers: We clamp it to 1! See: `cache_batch_size = max(cache_batch_size, dp_world_size)`.

open_instruct/dpo_utils.py
finbarrtimbers: This is a bug fix. It was broken previously, but we didn't exercise that code path.

open_instruct/olmo_core_callbacks.py
finbarrtimbers: Yeah, I agree. I did a small refactor into functions; lmk if you think that's cleaner?

open_instruct/padding_free_collator.py
finbarrtimbers: Added! I agree about the density. I'm open to suggestions for how to refactor.

open_instruct/padding_free_collator.py
finbarrtimbers: I added a comment, but basically: 

```
cu_seq_lens = [0, 100, 250, 400]  # 3 sequences of lengths 100, 150, 150
 max_seq_length = 200
 after clamp_max_(200): [0, 100, 200, 200]  # seq 3 was entirely beyond the cut
 after unique_consecutive: [0, 100, 200]
```

hamishivi: ohhh i see. very clever.

open_instruct/data_loader.py
```diff
@@ -170,7 +171,10 @@ def _iter_batches(self) -> Iterable[dict[str, Any]]:
             example = self.dataset[i]
             batch_examples.append(example | {"prompt_id": f"{self._epoch}_{example['index']}"})
             if len(batch_examples) == self._per_rank_batch_size:
-                yield to_device(self._collator(batch_examples), self._device)
+                all_examples = self._overflow + batch_examples
```
hamishivi, line 174: same comment as in #1472 - should we adjust the iter batches loop to go until the overflow is exhausted?

open_instruct/olmo_core_train_modules.py
```diff
@@ -151,17 +151,18 @@ def train_batch(self, batch: dict[str, Any], dry_run: bool = False) -> None:
...
             v.zero_()
 
         for micro_batch_idx, micro_batch in enumerate(micro_batches):
             with self._train_microbatch_context(micro_batch_idx, num_micro_batches):
                 loss, step_metrics = self._compute_microbatch_loss(micro_batch)
-                micro_tokens = (
-                    micro_batch["chosen_attention_mask"].sum() + micro_batch["rejected_attention_mask"].sum()
-                ).float()
+                micro_tokens = torch.tensor(
+                    padding_free_collator.get_num_tokens(micro_batch), dtype=torch.float32, device=device
```
hamishivi: nit: token counts can probably be longs.

open_instruct/padding_free_collator.py
```diff
@@ -1,8 +1,67 @@
...
+    for s_idx, item in enumerate(features):
+        input_ids = item["input_ids"]
+        ret["input_ids"].append(input_ids)
+
+        # Labels are next-token shifted: insert a separator, then drop the first token.
+        label_source = item["labels"] if is_labels_provided else input_ids
+        ret["labels"].append(separator)
+        ret["labels"].append(label_source[1:])
+
+        if return_flash_attn_kwargs and cu_seq_lens is not None:
```
hamishivi: the logic around return_flash_attn_kwargs is weird to me, I don't like mutable returns. Shouldnt we just always return the cu_seq_lens and let the caller decide if they want to keep or not? Does it add much runtime?

open_instruct/data_loader.py
finbarrtimbers: Done!

open_instruct/padding_free_collator.py
finbarrtimbers: no, you're right. I'll do an unconditional return.

open_instruct/olmo_core_train_modules.py
finbarrtimbers: Yeah, you're right. The only issue here is that we have to do a manual all-reduce for all of the metrics that we want to calculate the weighted mean for, and that happens in float32 as a bunch of those metrics are float32s:

https://github.com/allenai/open-instruct/blob/3fb1effe0dad5a8708b76c335e64448d53ecd40e/open_instruct/olmo_core_train_modules.py#L173

We could make them all doubles. Idk do you think it's worth refactoring?

PR #1467, opened 2026-02-09
Enable tensor parallelism in DPO

finbarrtimbers:
## Summary

- Adds tensor parallelism (TP=2) support for DPO training with torch.compile + packing.
- Introduces `DPOLMHead`, a custom LM head subclass that computes per-token log-probabilities inside the compile boundary, handling all DTensor logic internally so callers never see DTensors.
- Unifies both HF and OLMo-core forward paths: `_get_batch_logps` and `pf_get_batch_logps` now receive 2D per-token logps instead of 3D logits.
- Removes the `NotImplementedError` that blocked `tensor_parallel_degree > 1`.
- Reverts AC config to origin/main style (removes temporary `elif` full-mode fallback that was only needed when compile was disabled with TP).
- Fixes aot_autograd aliasing bug: `ce_loss=per_token_logps.detach().clone()` in `DPOLMHead` to break storage sharing between output tensors.

## Performance comparison (2-node, 16x H100)

| Config | MFU% | TPS/gpu | sec/step | Beaker | wandb |
|---|---|---|---|---|---|
| **TP=2, bs16, shard4 (this PR)** | **16.9%** | **3,926** | **3.88s** | [01KHV6Z3](https://beaker.org/ex/01KHV6Z3NWD4XK282HG8W7ZCJN) | [5eyfqjp6](https://wandb.ai/ai2-llm/open_instruct_internal/runs/5eyfqjp6) |
| **TP=2, bs32, shard4 (this PR)** | **16.3%** | **3,860** | **3.87s** | [01KJKS4P](https://beaker.org/ex/01KJKS4PE1NXS3B6D88KH9Q3XF) | [rqt3t6w6](https://wandb.ai/ai2-llm/open_instruct_internal/runs/rqt3t6w6) |
| No TP, bs16 (baseline) | 18.9% | 4,374 | 6.19s | [01KGTEY8](https://beaker.org/ex/01KGTEY8SZ3CWKHBX2KXD7HHAB) | [qhq1629a](https://wandb.ai/ai2-llm/open_instruct_internal/runs/qhq1629a) |
| No TP, bs32 (baseline) | 20.3% | 4,780 | 6.19s | [01KGTE43](https://beaker.org/ex/01KGTE43M9SBGD9D1Y7Z010EZ3) | [3hpay79n](https://wandb.ai/ai2-llm/open_instruct_internal/runs/3hpay79n) |
| No TP, bs8, noshard (baseline) | 8.8% | 2,038 | 4.21s | [01KH1S31](https://beaker.org/ex/01KH1S31HR9Y6BRN8MTNY50PV3) | [tv9jy94p](https://wandb.ai/ai2-llm/open_instruct_internal/runs/tv9jy94p) |

> **Note on MFU/TPS-per-GPU:** these metrics now correctly divide by all 16 physical GPUs (÷16) rather than just the 8 data-parallel ranks (÷8). The earlier entry in this table used the uncorrected ÷8 calculation, making TP=2 appear to have 2× higher MFU/TPS-per-GPU than it actually does.

TP=2 achieves **1.6x faster step time** (3.88s vs 6.19s) vs the no-TP baseline, while enabling larger batch sizes that OOM without TP. Per-GPU MFU is slightly lower with TP due to all-gather communication overhead, but this is the expected trade-off for fitting larger batches.

### Why is TP=2 MFU lower than no-TP despite faster step time?

TP=2 reduces the data-parallel world size from 16 to 8 ranks: each TP group of 2 GPUs acts as a single "device". With `per_device_train_batch_size=16`:

- **No-TP**: 16 DP ranks × 16 seqs = 256 sequences/step (~452k tokens)
- **TP=2, bs16**: 8 DP ranks × 16 seqs = 128 sequences/step (~244k tokens)
- **TP=2, bs32**: 8 DP ranks × 32 seqs = 256 sequences/step (~452k tokens) — same as no-TP bs16, but 1.6× faster steps (3.87s vs 6.19s)

So even though TP=2 step time is 1.6× faster (3.88s vs 6.19s), TP=2 bs16 processes 2× fewer tokens per step, giving ~16% lower TPS/GPU (~3,926 vs ~4,566). MFU follows TPS/GPU directly.

TP=2 bs32 restores the same effective batch size as no-TP bs16 (256 sequences/step) while keeping the 1.6× faster step time — achieving **1.6× higher throughput** than no-TP bs16 at the same effective batch size.

The benefit of TP=2 is not raw throughput but **memory**: splitting model weights across 2 GPUs frees enough activation memory to fit batch sizes and sequence lengths that would OOM without TP at larger model sizes or longer contexts.

## How it works

`DPOLMHead` (subclass of olmo-core `LMHead`) overrides `forward()` to:
1. Compute logits via the existing `w_out` linear (which is TP-parallelized by `apply_tp`)
2. Extract local shards of logits and labels via `get_local_tensor()`
3. Compute per-token log-probabilities on local shards (no full-vocab materialization)
4. All-gather via `DTensor.from_local().redistribute(Replicate()).to_local()`

Since this all happens inside the compiled LM head module, there are no DTensor/compile backward incompatibilities.

## Test plan

- [x] `make style && make quality` passes
- [x] `uv run pytest open_instruct/test_padding_free_collator.py` passes
- [x] Multi-node experiment succeeds: [Beaker](https://beaker.org/ex/01KHV6Z3NWD4XK282HG8W7ZCJN), [wandb](https://wandb.ai/ai2-llm/open_instruct_internal/runs/5eyfqjp6)
- [x] TP=2 bs32 experiment succeeds: [Beaker](https://beaker.org/ex/01KJKS4PE1NXS3B6D88KH9Q3XF), [wandb](https://wandb.ai/ai2-llm/open_instruct_internal/runs/rqt3t6w6)

### DIFF
--- a/open_instruct/environments/tools/parsers.py
+++ b/open_instruct/environments/tools/parsers.py
@@ -143,7 +143,7 @@ class VllmToolParser(ToolParser):
 
         Usually these only need the list of tools.
         """
-        return ChatCompletionRequest(model="dummy", messages=[], tools=self._tool_definitions)
+        return ChatCompletionRequest(model="dummy", messages=[], tools=self._tool_definitions)  # type: ignore
 
     def get_tool_calls(self, text: str) -> list[EnvCall]:
         """Extract tool calls from model output.


open_instruct/dpo.py
```diff
@@ -357,12 +368,20 @@ def main(args: dpo_utils.ExperimentConfig, tc: dataset_transformation.TokenizerC
...
-    ac_config = (
-        transformer_config.TransformerActivationCheckpointingConfig(
+    if args.activation_memory_budget < 1.0 and args.compile_model:
+        ac_config = transformer_config.TransformerActivationCheckpointingConfig(
             mode=transformer_config.TransformerActivationCheckpointingMode.budget,
             activation_memory_budget=args.activation_memory_budget,
         )
-        if args.activation_memory_budget < 1.0 and args.compile_model
+    elif args.activation_memory_budget < 1.0:
+        ac_config = transformer_config.TransformerActivationCheckpointingConfig(
```
tyler-romero: I'd recommend logging a warning here to denote that the budget strategy is being overwritten

open_instruct/environments/tools/parsers.py
finbarrtimbers, line 1: This is needed to get the type checker to stop complaining. I have it in a separate PR: https://github.com/allenai/open-instruct/pull/1504/.

open_instruct/dpo_utils.py
```diff
@@ -805,16 +807,8 @@ def _get_batch_logps(logits: torch.Tensor, labels: torch.Tensor, average_log_pro
...
-    labels = labels[:, 1:].clone()
-    logits = logits[:, :-1, :]
-    loss_mask = labels != -100
-
-    # dummy token; we'll ignore the losses on these tokens later
-    labels[labels == -100] = 0
-
-    per_token_logps = log_softmax_and_gather(logits, labels)
+    per_token_logps = per_token_logps[:, :-1]
+    loss_mask = labels[:, 1:] != -100
```
hamishivi, line 811: Do we support sequence parallel and if so is this handled correctly here? Since the shifts assume you're operating over the entire sequence.

open_instruct/olmo_core_callbacks.py
```diff
@@ -192,7 +195,7 @@ def post_step(self) -> None:
             total_tokens=total_tokens_step,
             avg_sequence_length=avg_sequence_length,
             training_time=training_time,
-            num_training_gpus=self.num_training_gpus,
+            num_training_gpus=self.num_training_gpus * self._tp_degree,
```
hamishivi: Is this really the case? Seems weird.

open_instruct/olmo_core_train_modules.py
```diff
@@ -23,6 +26,55 @@
 logger = logger_utils.setup_logger(__name__)
 
 
+class DPOLMHead(LMHead):
```
hamishivi, line 29: Is it possible to push this into olmo-core in the future? I feel like this violates the modelling <-> data handling divide that we had before btw olmo-core <-> open-instruct

scripts/train/debug/dpo/multi_node.sh
```diff
@@ -48,4 +48,7 @@ uv run python mason.py \
     --with_tracking \
     --push_to_hub false \
     --try_launch_beaker_eval_jobs false \
+    --shard_degree 4 \
```
hamishivi: whats shard degree?

open_instruct/dpo_utils.py
```diff
@@ -1071,7 +1070,9 @@ def concatenated_forward_olmo(
     """Run the given model on the given batch of inputs, concatenating the chosen and rejected inputs together.
 
     We do this to avoid doing two forward passes, because it's faster for FSDP.
-    Uses OLMo-core Transformer interface: model(input_ids) returns logits tensor directly.
+    Uses OLMo-core Transformer interface. Passes labels so the LM head handles DTensor
+    correctly under TP+compile, then computes per-token log-probs on the local logits shard
+    to avoid materializing the full (S, vocab) tensor.
```
hamishivi, line 1075: does this imply we need to do a collective op later on to correctly compute the loss, or is that handled fine automatically?

open_instruct/dpo.py
finbarrtimbers: Was able to remove this entirely!

open_instruct/dpo_utils.py
finbarrtimbers: No, we don't support sequence parallel (currently).

open_instruct/olmo_core_train_modules.py
finbarrtimbers: Yes, I agree entirely. It is possible to push this into Olmo-core. I'll work on that once we have GRPO done (as presumably that will require similar changes).

scripts/train/debug/dpo/multi_node.sh
finbarrtimbers: Updated to name it `--fsdp_shard_degree`. It's the number of GPUs to shard the model over for FSDP. We also have `--num_replicas` (now `--fsdp_num_replicas`) which is the number of times we replicate the model. 

E.g. one config we could do is `--fsdp_num_replicas 4 --fsdp_shard_degree 8`, which would have 4 copies of the model, each sharded over 8 GPUs (i.e. we have a copy of the model per node).

open_instruct/olmo_core_callbacks.py
finbarrtimbers: It's a bad name. Should be `dp_world_size`. I have renamed it.

open_instruct/dpo_utils.py
finbarrtimbers: Yes, but it's handled within `DPOLMHead` automatically.

open_instruct/olmo_core_callbacks.py
hamishivi: dont see the rename?

open_instruct/olmo_core_train_modules.py
hamishivi: SG! Could we add a todo?

scripts/train/debug/dpo/multi_node.sh
hamishivi: cool!

open_instruct/olmo_core_callbacks.py
finbarrtimbers: Done!

open_instruct/olmo_core_train_modules.py
finbarrtimbers: Done!

hamishivi: i think not pushed?

PR #1472, opened 2026-02-13
Replaces `lambda` collators with a `single_example_collator` to use a consistent interface

finbarrtimbers:
This lets us avoid some branching in https://github.com/allenai/open-instruct/pull/1466.

Experiment links:
- GRPO single GPU (succeeded): https://beaker.org/ex/01KHC0ZXVCNVGWM6QJSNFWN09R
- DPO single GPU (queued): https://beaker.org/ex/01KHC100E8XNJABJ8GZBHQ8HFV

### DIFF
--- a/open_instruct/data_loader.py
+++ b/open_instruct/data_loader.py
@@ -176,6 +176,13 @@ class HFDataLoader(data_loader.DataLoaderBase):
                 self._overflow = all_examples[len(batch["index"]) :]
                 yield batch
                 batch_examples = []
+        while self._overflow:
+            batch = to_device(self._collator(self._overflow), self._device)
+            assert len(batch["index"]) > 0, (
+                f"Collator consumed 0 examples from {len(self._overflow)} overflow examples"
+            )
+            self._overflow = self._overflow[len(batch["index"]) :]
+            yield batch
 
     @property
     def total_batches(self) -> int:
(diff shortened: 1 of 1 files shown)


open_instruct/data_loader.py
```diff
@@ -537,10 +541,15 @@ class BatchStatistics:
     total_prompts: int
 
 
+def single_example_collator(examples: list[dict[str, Any]]) -> dict[str, Any]:
```
hamishivi, line 551: should we add a guard/assert that len examples is 1?

open_instruct/data_loader.py
```diff
@@ -170,7 +171,10 @@ def _iter_batches(self) -> Iterable[dict[str, Any]]:
             example = self.dataset[i]
             batch_examples.append(example | {"prompt_id": f"{self._epoch}_{example['index']}"})
             if len(batch_examples) == self._per_rank_batch_size:
-                yield to_device(self._collator(batch_examples), self._device)
+                all_examples = self._overflow + batch_examples
+                batch = to_device(self._collator(all_examples), self._device)
+                self._overflow = all_examples[len(batch["index"]) :]
```
hamishivi, line 176: should we add a final set of iterations to flush out the overflow at the end of the epoch?

finbarrtimbers: yeah, sure. it should be fine because we pack + pad, so it'll still be the same size as usual.

open_instruct/data_loader.py
finbarrtimbers: done!

PR #1474, opened 2026-02-17
Fix nodes_needed calculation for kv_cache_max_concurrency warning in grpo_fast

MohdElgaar:
When `kv_cache_max_concurrency` is lower than the expected batch size, grpo_fast logs a warning suggesting how many inference nodes to use. The `nodes_needed` calculation used floor division, which could round down and undercount. Changed to `math.ceil()`.


hamishivi (approved):
Thanks!

PR #1475, opened 2026-02-17
Add RL environment abstraction

hamishivi:
## Summary

First step toward merging `rl-environments-v2` in small, self-contained chunks. This PR adds the core RL environment abstraction without touching the training pipeline.

RL environments follow the OpenEnv API in exposing step/reset/state methods. Each RL env instance is a given sample, so we have some pool logic for managing that. Since I want to re-use the existing tool logic, we also end up editing the tool stuff to align it with the envs.

Next: integrating with the training loop more tightly.

### DIFF
--- a/tests/test_environments.py
+++ b/tests/test_environments.py
@@ -1,6 +1,5 @@
 """Unit tests for RL environments."""
 
-import asyncio
 import random
 
 import pytest
@@ -9,11 +8,6 @@ from open_instruct.environments import ENV_REGISTRY, get_env_class
 from open_instruct.environments.base import RLEnvironment, StepResult
 from open_instruct.environments.examples import CounterEnv, GuessNumberEnv
 from open_instruct.tools.utils import ToolCall
-
-
-def run_async(coro):
-    """Run async function in sync test."""
-    return asyncio.run(coro)
 
 
 class TestRegistry:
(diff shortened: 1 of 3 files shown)


open_instruct/tools/environments/base.py
```diff
@@ -0,0 +1,121 @@
...
+    @property
+    def final_reward(self) -> float:
+        return self.rewards[-1] if self.rewards else 0.0
+
+    @property
+    def total_reward(self) -> float:
+        return sum(self.rewards)
+
+
+ENV_REGISTRY: dict[str, type["RLEnvironment"]] = {}
```
finbarrtimbers: Can you move this after the RLEnvironment definition so we can use RLEnvironment as a type?

open_instruct/tools/environments/base.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def get_env_class(env_name: str | None = None, env_class: str | None = None) -> type[RLEnvironment]:
+    """Get environment class by registry name or import path."""
+    if env_name is not None:
+        if env_name not in ENV_REGISTRY:
+            raise ValueError(f"Environment '{env_name}' not found. Available: {list(ENV_REGISTRY.keys())}")
+        return ENV_REGISTRY[env_name]
+
+    if env_class is not None:
```
finbarrtimbers: Do we need both pathways, or can we force the caller to go through the registry?

open_instruct/tools/environments/pool.py
```diff
@@ -0,0 +1,92 @@
...
+        self._available = asyncio.Queue()
+        for actor in self._actors:
+            self._available.put_nowait(actor)
+
+        self._initialized = True
+        logger.info(f"Environment pool initialized with {self.pool_size} actors")
+
+    async def acquire(self) -> ray.actor.ActorHandle:
+        """Acquire an available environment actor (blocks until available)."""
+        if not self._initialized:
```
finbarrtimbers: Why do we need this? Why not always initialize automatically or in __init__?

open_instruct/tools/environments/pool.py
```diff
@@ -0,0 +1,92 @@
...
+            raise RuntimeError("Pool not initialized")
+        return await self._available.get()
+
+    def release(self, actor: ray.actor.ActorHandle) -> None:
+        """Release an environment actor back to the pool."""
+        if self._available is not None:
+            self._available.put_nowait(actor)
+
+    async def shutdown(self) -> None:
```
finbarrtimbers: Have you considered having this happen automatically by making this `__del__`?

open_instruct/tools/environments/pool.py
```diff
@@ -0,0 +1,92 @@
...
+
+        for actor in self._actors:
+            ray.kill(actor)
+
+        self._actors = []
+        self._available = None
+        self._initialized = False
+        logger.info("Environment pool shutdown complete")
+
+    @property
```
finbarrtimbers: What about renaming `self._env_class` to `self.env_class` and removing this definition?

open_instruct/tools/tools.py
finbarrtimbers, line 1: I don't like the `# type: ignore[override]`. Do we have any other options to fix this?

open_instruct/tools/utils.py
```diff
@@ -211,8 +214,11 @@ def compute_metrics(self) -> dict[str, float]:
 
 @dataclass
 class ToolCall:
+    """Parsed tool call from model output."""
+
     name: str
     args: dict[str, Any]
+    id: str | None = None
```
finbarrtimbers: Can you add a comment saying when this can be None?

open_instruct/tools/utils.py
```diff
@@ -408,25 +414,49 @@ def coerce_args(properties: dict[str, Any], kwargs: dict[str, Any]) -> dict[str,
 
 
 class Tool(ABC):
```
finbarrtimbers: I don't like having all of these defaults. Why are they necessary? 

I particularly don't like it for the containers.

open_instruct/environments/examples.py
finbarrtimbers, line 1: Add some tests here? 

Could just be a parameterized test that rips through 20 steps for each.

open_instruct/tools/environments/pool.py
```diff
@@ -0,0 +1,92 @@
...
+        self.pool_size = pool_size
+        self.env_name = env_name
+        self.env_class_path = env_class
+        self.env_kwargs = env_kwargs
+        self._env_class: type[RLEnvironment] | None = None
+        self._actors: list[ray.actor.ActorHandle] = []
+        self._available: asyncio.Queue[ray.actor.ActorHandle] | None = None
+        self._initialized = False
+
+    async def initialize(self) -> None:
```
finbarrtimbers: Why do we need this? Why not do this in `__init__`?

finbarrtimbers (changes requested):

open_instruct/tools/tools.py
hamishivi: We could make the signature always **kwargs and force the implementer to parse out the kwargs in the function body. But this also feels gross to me...

open_instruct/tools/environments/base.py
hamishivi: ah, good point. lets just use the registry.

open_instruct/tools/utils.py
hamishivi: oop, will force as string. A bit weird since I wanted to leave the vllm_utils integration out of this PR, where id will be passed through.

open_instruct/tools/environments/pool.py
hamishivi: Oh, good point. I think originally I wanted to lazy-init envs with heavy setup things. But actually I ended up changing this design, so now the pool can be sync.

open_instruct/tools/environments/pool.py
hamishivi: my understanding is that __del__ is a bit unreliable since we don't have guarantees on GC running? We could make it del and then explcitly call it, or leave as shutdown, i don't mind strongly.

open_instruct/tools/utils.py
hamishivi: Removed! I did some restructuring.

open_instruct/tools/tools.py
```diff
@@ -54,10 +55,10 @@ def __init__(self, call_name: str, api_endpoint: str, timeout: int = 3) -> None:
         self.api_endpoint = api_endpoint
         self.timeout = timeout
 
-    async def _execute(self, code: str) -> ToolOutput:
+    async def _execute(self, code: str) -> ExecutableOutput:  # type: ignore[override]
```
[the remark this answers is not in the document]
hamishivi: I dislike this since it makes implementing new tools annoying. but I can do this if we want.

open_instruct/environments/base.py
```diff
@@ -0,0 +1,110 @@
...
+from abc import abstractmethod
+from dataclasses import dataclass, field
+from typing import Any
+
+from open_instruct.executable import Executable, ExecutableOutput
+from open_instruct.tools.utils import ToolCall
+
+
+@dataclass
+class StepResult:
```
finbarrtimbers, line 13: Why not use the `OpenEnv` types here? E.g. `OpenEnv`'s `Observation` works for `StepResult`. 

Similarly, [`State`](https://github.com/meta-pytorch/OpenEnv/blob/5680f6470a153534bd6c6141a69ee50b46a7ccd5/src/openenv/core/env_server/types.py#L177), etc.

open_instruct/tools/environments/pool.py
finbarrtimbers: Ah, sorry. This was specifically referring to the `if not self._initialized` check.

open_instruct/tools/environments/pool.py
finbarrtimbers: Will we ever call `del` before the program exits? What are the practical concerns of unpredictable GC calls?

open_instruct/tools/tools.py
finbarrtimbers: I feel like that's better, though? That *is* the API, right?

tests/test_environments.py
finbarrtimbers, line 1: Could we put this in `tools/environments/test_examples.py`? I think that's more consistent with the test structure elsewhere in `open_instruct/`.

tests/test_environments.py
```diff
@@ -0,0 +1,239 @@
+"""Unit tests for RL environments."""
+
+import asyncio
+
+from open_instruct.tools.environments import ENV_REGISTRY, get_env_class
+from open_instruct.tools.environments.examples import CounterEnv, GuessNumberEnv
+from open_instruct.tools.utils import ToolCall
+
+
+def run_async(coro):
```
finbarrtimbers: I think this is redundant and doesn't add anything.

tests/test_environments.py
```diff
@@ -0,0 +1,239 @@
...
+    """Run async function in sync test."""
+    return asyncio.run(coro)
+
+
+class TestRegistry:
```
finbarrtimbers, line 13: Can you parameterize these?

open_instruct/tools/environments/base.py
finbarrtimbers: I don't think this was done?

tests/test_environments.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def test_get_env_class_no_args(self):
+        try:
+            get_env_class()
+            assert False, "Expected ValueError"
+        except ValueError:
+            pass
+
+
+class TestCounterEnv:
```
finbarrtimbers, line 81: I don't think you need specific tests like this. let's just do: 

```
class TestExampleEnvironments(unittestTestCase):

    @parameterize.parameterized([
        ENV_REGISTRY.values()
    ])
    def test_full_episode(self, env):
         observation = env.reset()
          while not observation.done():
               action = random.choice(observation.legal_actions())
               env.step(action)
```

I just want to make sure that the example environments don't silently break.

open_instruct/tools/tools.py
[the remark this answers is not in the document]
finbarrtimbers: I think we should, unfortunately. Or we can figure out a better high-level API.

open_instruct/tools/environments/pool.py
finbarrtimbers: What was the conclusion here?

open_instruct/tools/environments/pool.py
```diff
@@ -0,0 +1,92 @@
...
+        """Create all environment actors and call setup() on each."""
+        if self._initialized:
+            return
+
+        self._env_class = get_env_class(env_name=self.env_name, env_class=self.env_class_path)
+        actor_class = ray.remote(self._env_class)
+
+        logger.info(f"Creating {self.pool_size} environment actors")
+        self._actors = [actor_class.remote(**self.env_kwargs) for _ in range(self.pool_size)]
```
finbarrtimbers: Wdyt about lazily building the pool actors? Could build them in `acquire`?

tests/test_environments.py
hamishivi: Oh sure, I don't want it to be everything in the env registry though, because I plan to add envs that require docker, apis, etc.

open_instruct/tools/environments/pool.py
hamishivi: oh, that its clearer to me the expected pattern is something like:
```python
try:
   pool.acquire()
   ...
finally:
   pool.shutdown()
```

Rather than calling `del pool` at the `finally` clause. In this codebase, this will appear as the pool shutdown being called when the training dies.

open_instruct/tools/tools.py
hamishivi: `coerce_args` actually does a bunch of safety work to force kwargs into it. Could explicitly make it a dict, or a class that wraps around the dict and has the coerce validation fn on it.

PR #1477, opened 2026-02-18
Fix test single_example_collator to wrap index as tensor

hamishivi:
## Summary

The `single_example_collator` in `test_data_loader_gpu.py` returned the raw example dict with `index` as a plain `int`. `HFDataLoader._iter_batches` calls `len(batch["index"])` which fails with `TypeError: object of type 'int' has no len()`.

Fix: wrap index as `torch.tensor([index])` to match the production `single_example_collator` in `data_loader.py`.

## Test plan
- [x] The fix matches the production collator pattern

### DIFF
--- a/open_instruct/test_data_loader_gpu.py
+++ b/open_instruct/test_data_loader_gpu.py
@@ -79,7 +79,7 @@ class TestHFDataLoader(unittest.TestCase):
         for _dp_rank, loader in enumerate(loaders):
             rank_indices = []
             for batch in loader:
-                rank_indices.append(batch["index"])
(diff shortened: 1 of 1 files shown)


finbarrtimbers:
ahhh can you re-run the single GPU script? I can do it tomorrow. I want to make sure that doesn't break

finbarrtimbers:
ok ran the single GPU GRPO script: [Beaker](https://beaker.org/ex/01KHRR95S5EBK4A3GW236ZRM4W).

finbarrtimbers:
And it passed! So I'm re-adding to the merge queue.

PR #1480, opened 2026-02-19
Wait for vLLM update_weight RPCs before unpausing actors in weight sync

MohdElgaar:
The main thread could hang in `health_check_fn()` on `check_background_threads.remote()` because `should_stop` was set to False before all vLLM `update_weight` RPCs finished. vLLM engines stayed busy with queued updates and never processed the health check.

Fix: Wait for all `engine.update_weight` ObjectRefs returned by `broadcast_to_vllm()` to complete before calling `set_should_stop(False)`, so vLLM engines are idle before the health check runs.


MohdElgaar:
Note: this is only when `inflight_updates=False`. The stall never happens when `inflight_updates=True`.
https://github.com/allenai/open-instruct/blob/c4b10fc1a3e8185d328a716deefa15685ffce8dd/open_instruct/vllm_utils.py#L810-L817

hamishivi:
@MohdElgaar if you add to the changelog and fix the quality checks I'm happy to merge!

PR #1481, opened 2026-02-19
Fix eval_on_step_0 never working in grpo_fast

MohdElgaar:
The training loop starts at `training_step=1` (no step 0). The old condition required `training_step % local_eval_every == 0`, so with `eval_on_step_0=True` and `local_eval_every > 1`, evaluation never ran on step 1.

Fix: Run evaluation when `eval_on_step_0=True` and `training_step == resume_training_step`, independent of `local_eval_every`.

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -1897,9 +1897,11 @@ def run_training(
         health_check_time = time.perf_counter() - health_check_start
 
         if (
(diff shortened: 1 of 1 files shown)


hamishivi:
@MohdElgaar thanks a lot for the fix! I tweaked the setup a  bit and added a fix with #1485.

PR #1482, opened 2026-02-19
Remove vllm_num_engines from VLLMConfig; compute inline from cluster resources

finbarrtimbers:
Previously users had to manually pass `--vllm_num_engines` when launching `grpo_fast.py`, even though the value is fully determined by the rest of the config. We have three variables and three flags: 

1. `--num_nodes`
2. `--num_learners_per_node`
3. `--vllm_num_engines`
4. `--vllm_tensor_parallel_size`

We have an equation with 4 variables (total number of GPUs, number of learner GPUs, GPUs per actor, and total actor GPUs). We can uniquely determine it with 3 variables. 

This PR removes `--vllm_num_engines` and sets it automatically.

Runs:

1. Single GPU GRPO: [Beaker](https://beaker.org/ex/01KHSWCG9VTJ0DN00A69GTA7T2)
2. Multi-node GRPO: [Beaker](https://beaker.org/ex/01KHT38E8Z8ZX7NEYPP4JDY8PY)

### DIFF
--- a/open_instruct/benchmark_generators.py
+++ b/open_instruct/benchmark_generators.py
@@ -243,8 +243,7 @@ def setup_vllm_engines(
     assert model_config.model_name_or_path is not None
 
     total_gpus = int(ray.cluster_resources().get("GPU", 0))
-    learner_gpus = sum(args.num_learners_per_node)
-    vllm_num_engines = (total_gpus - learner_gpus) // vllm_config.vllm_tensor_parallel_size
+    vllm_num_engines = total_gpus // vllm_config.vllm_tensor_parallel_size
 
     vllm_engines = vllm_utils.create_vllm_engines(
         num_engines=vllm_num_engines,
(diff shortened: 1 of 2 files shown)


open_instruct/grpo_fast.py
```diff
@@ -2048,6 +2051,22 @@ def main(
     # We have to initialize ray earlier for constructing Tools (they are implemented as ray actors).
     ray.init(dashboard_host="0.0.0.0", runtime_env={"excludes": [".git/"], "env_vars": dict(os.environ)})
 
+    if args.single_gpu_mode:
+        vllm_num_engines = 1
+        logger.info("Auto-calculated vllm_num_engines=1 (single_gpu_mode)")
+    else:
+        total_gpus = int(ray.cluster_resources().get("GPU", 0))
```
hamishivi, line 2058: What happens if there's a launch issue and not every node starts up? The user could end up launching with less inference nodes than they expect.

Issue #1473, opened 2026-02-16
OLMo-3-7b RLVR: vllm_vs_local_logprob_diff_mean increase  (even with learning_rate=0), eventually causing rollout actor timeouts


lukashelff:
I am trying to replicate the `scripts/train/olmo3/7b_think_rl.sh` GRPO setup on a Slurm multi-node cluster using open-instruct and vLLM, but consistently observe a gradual divergence between vLLM rollout logprobs and locally computed policy logprobs. After approximately 450 training steps, the model appears to derail, with a steady increase in `vllm_vs_local_logprob_diff_mean`, eventually causing rollout actor timeouts (`LLMRayActor`) and training failure around step ~550.

Resuming training from earlier checkpoints did not resolve the issue. To further isolate the problem, I restarted training from a checkpoint at step 500 with weights fixed (`learning_rate=0`) to ensure no further learning occurs. However, the divergence in `vllm_vs_local_logprob_diff_mean` continued to increase even in this setting. Since the weights should remain constant when the learning rate is zero, I would expect the logprob differences between vLLM and the local policy forward pass to remain stable, rather than grow over time.

This suggests there may be some form of mismatch or drift between vLLM rollout logprob computation and the learner’s local forward pass that is independent of weight updates.

The figure below shows three runs:

* **Yellow:** base training run, which crashes around step ~550
* **Blue:** resumed run from step 400, which diverges and crashes similarly
* **Red:** resumed run from step 500 with `learning_rate=0`, where divergence still continues despite fixed weights

Overall, training metrics such as rewards and sequence lengths appear healthy for the first ~400 steps, before the logprob divergence begins to accelerate. The codebase is mostly identical to the current state of the repo.

## My Slurm launch script is below for reference:

```bash
#!/bin/bash
#SBATCH --job-name=RLVR-basev2
#SBATCH --partition=all
#SBATCH --nodes=8
#SBATCH --gpus-per-node=8
#SBATCH --ntasks-per-node=1
#SBATCH --cpus-per-task=112
#SBATCH --mem=1000G
#SBATCH --time=200:00:00
#SBATCH --output=logs/%x_%j.out
#SBATCH --error=logs/%x_%j.err
#SBATCH --qos=normal
#SBATCH --open-mode=append
#SBATCH --exclude=cn13,cn06,cn05
JOB_NAME="RLVR-soofi-basev2"


# --- 1. Variables & Paths ---
[...]

PR #1483, opened 2026-02-19
RLVR: fixing logprob divergence

lukashelff:
fix data indexing, leading to a mismatch between prompt-responses, and thus to logprob divergence between trainer and actor during GRPO with RLVR. (Issue #1473 )


hamishivi:
Thanks! Fixed in #1484 to make the change less major (just dropped the shuffle).

PR #1208, opened 2025-11-18
Implements an `OLMo-core` compatible data loader to load HF datasets from `get_cached_dataset_tulu`. 

finbarrtimbers:
There's not a ton of point to this, right now. But by making this change, we can subsequently: 

1) move `dpo_tune_cache.py` over to use this, which will (eventually) let that use the Olmo-core trainer
2) Give each LLMRayActor their own data_loader, so they can pull independently and not have to coordinate with `main`. This will let us remove `add_prompt_to_generator` and have them operate fully asynchronously. This will make switching the entire `grpo_fast.py` script over to the data_loader/trainer pattern easier.

Runs: 

- Single GPU GRPO: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/open-instruct-dev/work/01KAC7FADGWJFMVMF7Z6A7MHWK?taskId=01KAC7FADMXYSEG1C5S26VMPDE&jobId=01KAC7FANAP7ECR5VH70VN7T0F)
- Multi-node GRPO script: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/open-instruct-dev/work/01KAVP00AZWTA505ZGRND8QSKV?taskId=01KAVP00B2RDE2ACT982ZRD0H6&jobId=01KAVP00EHXXCT51MAWN6PSX2R)

---

> [!NOTE]
> Introduces an OLMo-core–compatible HF dataset loader and refactors GRPO to use it (replacing custom iterators/maps) with prompt_id-based request tracking; updates tests and dependencies.
> 
> - **Data Loading**:
>   - Add `HFDataLoader` (`open_instruct/data_loader.py`) implementing `DataLoaderBase` with HF `Dataset` sharding, shuffling, exclusion, checkpointing (`state_dict`/`load_state_dict`), optional auto-reshuffle, and per-example `prompt_id`.
> - **GRPO Pipeline Refactor** (`open_instruct/grpo_fast.py`):
>   - Remove `ShufflingIterator` and `PendingQueriesMap`; use `HFDataLoader` for prompt sourcing, replenishment, and no-resampling via `exclude_index`.
>   - Update `accumulate_inference_batches`, `data_preparation_thread`, `run_training`, and eval flow to consume from `HFDataLoader` and lookup examples by `dataset_index`.
>   - Simplify `add_prompt_to_generator(example, param_prompt_Q, generation_config, is_eval)`; batch/eval queuing now uses `prompt_id`.
[...]

PR #1493, opened 2026-02-24
Fix evals never queuing beyond the first time in grpo_fast

MohdElgaar:
This PR fixes a critical eval bug in `grpo_fast`: eval prompts were only enqueued on the first eval round, then silently dropped to zero on all subsequent rounds.

Before this change, eval effectively ran once per process lifetime. After the first full pass, subsequent eval rounds enqueued **0** prompts, and local eval never completes.

- **Critical fix:** call `eval_data_loader.reset()` after consuming the eval loader each round.
  - Without reset, `HFDataLoader` keeps `batches_processed` at end-of-epoch, so:
    - `for eval_example in iter(eval_data_loader):` works once,
    - then yields no samples forever after.
  - This is consistent with `OLMo-core` `DataLoaderBase` semantics, which require `reset()` after exhausting an epoch ([reference](https://github.com/allenai/OLMo-core/blob/main/src/olmo_core/data/data_loader.py#L190)).
- **Secondary fix:** pass `training_step` (instead of constant `0`) into eval prompt IDs to avoid cross-round request ID collisions/overwrites in vLLM metadata tracking.

This directly addresses the [failure mode](https://github.com/allenai/open-instruct/issues/1385#issuecomment-3841827982) discussed in issue #1385.


MohdElgaar:
I also had run logs from Nov 2025 showing that subsequent evals were running fine, so I checked what had changed. Apparently, this bug was introduced by #1208 (0f7ef0d2735c78233c90b9c622b1b1f981552775).

finbarrtimbers:
Nice catch! Thank you.

PR #1500, opened 2026-02-27
Enable multiple active environments per rollout

hamishivi:
## Summary

Now we can have multiple RL envs in one rollout! This might be useful if you want to e.g. combine the sandbox LM setup with RL games like wordle (see the new debug script for an example).

Our one exception: we only allow one active text environment per rollout. This is because if we had multiple text envs, who would own the rollout?

This also allows use to unify the tool and environment handling a bit (since before we handled tools differently to allow multiple tools per rollout), simplifying the code paths. Also adds a qwen 3 xml parser (for the wordle+sandbox script).

## testing

- [x] 1-gpu tool script
- [x] 1-gpu env script
- [ ] 8-gpu wordle script (https://beaker.org/orgs/ai2/workspaces/open-instruct-dev/work/01KK07KAQW7F5A37NRS2AWZMQ2?taskId=01KK07KAR4E09GQHZ68MSHHXBC&jobId=01KK07KAVKFB0812YHB5PFJ5PF)
- [x] 8-gpu wordle + sandbox script [new, specifically to test this code] (https://beaker.org/orgs/ai2/workspaces/open-instruct-dev/work/01KK07PPKECMS46TV71661PGRK?taskId=01KK07PPKJMC7CPNEBX6CKKN5V&jobId=01KK07PPPZM38YEPW8KMTNYRAG)

### DIFF
--- a/open_instruct/data_loader.py
+++ b/open_instruct/data_loader.py
@@ -597,7 +597,11 @@ def _aggregate_env_metrics(rollout_states: list[dict]) -> dict[str, float]:
             if k != "env_name" and isinstance(v, (int, float)):
                 bucket.setdefault(k, []).append(float(v))
 
-    return {f"env/{ename}/{k}": float(np.mean(vals)) for ename, metrics in env_metrics.items() for k, vals in metrics.items()}
+    return {
+        f"env/{ename}/{k}": float(np.mean(vals))
+        for ename, metrics in env_metrics.items()
+        for k, vals in metrics.items()
+    }
 
 
 def add_prompt_to_generator(


open_instruct/data_loader.py
```diff
@@ -557,6 +557,66 @@ def single_example_collator(examples: list[dict[str, Any]]) -> dict[str, Any]:
     return example | {"index": torch.tensor([example["index"]])}
 
 
+def _extract_env_configs(env_config: dict[str, Any] | None) -> list[dict[str, Any]]:
+    """Return env configs from canonical payload format."""
+    if env_config is None:
+        return []
+    if not isinstance(env_config, dict):
```
finbarrtimbers: I think this is overly defensive; let's let the type checker catch this.

open_instruct/data_loader.py
```diff
@@ -557,6 +557,66 @@ def single_example_collator(examples: list[dict[str, Any]]) -> dict[str, Any]:
...
+        raise TypeError(f"env_config must be a dict or None, got {type(env_config).__name__}")
+    if "env_name" in env_config:
```
finbarrtimbers: Let's remove the `.get("env_configs")` in the next line and have this raise a `KeyError`!

open_instruct/data_loader.py
```diff
@@ -557,6 +557,66 @@ def single_example_collator(examples: list[dict[str, Any]]) -> dict[str, Any]:
...
+        raise ValueError("env_config must use canonical form {'env_configs': [...]} at this stage.")
+    return [dict(cfg) for cfg in (env_config.get("env_configs") or [])]
+
+
+def _index_env_configs(envs: list[dict[str, Any]]) -> dict[str, dict[str, Any]]:
```
finbarrtimbers: Let's inline this, as it's only used once?

open_instruct/data_loader.py
```diff
@@ -557,6 +557,66 @@ def single_example_collator(examples: list[dict[str, Any]]) -> dict[str, Any]:
...
+    """Return env_name -> config mapping for quick lookup."""
+    return {cfg["env_name"]: cfg for cfg in envs}
+
+
+def _merge_sample_envs(
+    sample_envs: list[dict[str, Any]], base_envs_by_name: dict[str, dict[str, Any]]
+) -> list[dict[str, Any]]:
+    """Merge sample-provided env configs with same-name defaults from base config."""
+    merged_envs: list[dict[str, Any]] = []
```
finbarrtimbers: If we make `base_envs_by_name` a `collections.defaultdict` we can get rid of this line. 

Then, we can rewrite this as: 

```
merged_envs = [base_envs_by_name[sample_env["env_name"]] | sample_env for sample_env in sample_envs]
```

(and then we can inline this as it's only called once.)

open_instruct/data_loader.py
```diff
@@ -557,6 +557,66 @@ def single_example_collator(examples: list[dict[str, Any]]) -> dict[str, Any]:
...
+    base_env_config: dict[str, Any] | None, sample_env_config: dict[str, Any] | None
+) -> dict[str, Any] | None:
+    """Merge base and sample env config into canonical payload.
+
+    Canonical form:
+      {"max_steps": int | None, "env_configs": [<env dict>, ...]}
+    """
+    base_envs = _extract_env_configs(base_env_config)
+    base_envs_by_name = _index_env_configs(base_envs)
+    base_max_steps = base_env_config.get("max_steps") if base_env_config is not None else None
```
finbarrtimbers: Can you help me understand the logic here? Why would `base_env_config` be None (or not)?

open_instruct/data_loader.py
```diff
@@ -557,6 +557,66 @@ def single_example_collator(examples: list[dict[str, Any]]) -> dict[str, Any]:
...
+
+    if sample_env_config is None:
+        envs = base_envs
+        max_steps = base_max_steps
+    else:
+        if not isinstance(sample_env_config, dict):
+            raise TypeError(f"sample_env_config must be a dict or None, got {type(sample_env_config).__name__}")
+        max_steps = sample_env_config.get("max_steps", base_max_steps)
+        sample_envs = _extract_env_configs(sample_env_config)
+        envs = _merge_sample_envs(sample_envs, base_envs_by_name) if sample_envs else base_envs
```
finbarrtimbers: There's a lot of nested logic here, can we refactor this to have less nesting?

open_instruct/data_loader.py
```diff
@@ -557,6 +557,66 @@ def single_example_collator(examples: list[dict[str, Any]]) -> dict[str, Any]:
...
+
+    if not envs:
+        return None
+
+    payload: dict[str, Any] = {"env_configs": envs}
+    if max_steps is not None:
```
finbarrtimbers: What about always setting this to be None and letting the downstream caller handle that? Having a variable number of keys seems confusing. 

Then, payload could be a `dataclass` with known keys instead of a dict, which would be clearer imo.

Dockerfile
```diff
@@ -70,7 +70,7 @@ ENV UV_CACHE_DIR=/root/.cache/uv \
 RUN --mount=type=cache,target=${UV_CACHE_DIR} \
     --mount=type=bind,source=uv.lock,target=uv.lock \
     --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
-    uv run --frozen python -c "import nltk; nltk.download('punkt'); nltk.download('punkt_tab')"
+    uv run --frozen python -c "import nltk; nltk.download('punkt'); nltk.download('punkt_tab'); nltk.download('words')"
```
finbarrtimbers: what if we do: 

```
uv run --frozen python -m nltk.downloader punkt punkt_tab words
```

I verified locally that this works.

open_instruct/vllm_utils.py
```diff
@@ -887,49 +888,82 @@ async def process_request(actor: LLMRayActor, sub_request_id: str, sampling_para
     max_model_len = actor.llm_engine.model_config.max_model_len
 
     configured_tools = set(actor.pools.keys())
-    allowed_tools = configured_tools & set(active_tools) if active_tools is not None else configured_tools
+    allowed_tools = configured_tools & set(active_tools) if active_tools is not None else set(configured_tools)
 
-    max_steps = env_config.get("max_steps", actor.max_steps) if env_config else actor.max_steps
+    max_steps = actor.max_steps
```
finbarrtimbers: inline this?

open_instruct/vllm_utils.py
```diff
@@ -887,49 +888,82 @@ async def process_request(actor: LLMRayActor, sub_request_id: str, sampling_para
...
+    if env_config is not None and not isinstance(env_config, dict):
```
finbarrtimbers: Let the type checker handle this?

open_instruct/vllm_utils.py
```diff
@@ -887,49 +888,82 @@ async def process_request(actor: LLMRayActor, sub_request_id: str, sampling_para
...
+        raise TypeError(f"env_config must be a dict or None, got {type(env_config).__name__}")
+    if isinstance(env_config, dict):
```
finbarrtimbers: What is this if not a dict? 

I generally don't like having `isinstance` checks, do we have any other options?

open_instruct/vllm_utils.py
```diff
@@ -887,49 +888,82 @@ async def process_request(actor: LLMRayActor, sub_request_id: str, sampling_para
...
+    active_env_names: list[str] = []
+    text_env_names: list[str] = []
 
     output = None
     try:
-        # If env_config is present, acquire from the env's pool and call reset
-        if env_config is not None:
-            env_name = env_config["env_name"]
-            pool = actor.pools.get(env_name)
+        unknown_targets = set(env_config_by_name) - configured_tools
```
finbarrtimbers: Can you break this block up into functions? It's pretty dense.

finbarrtimbers (changes requested):

open_instruct/data_loader.py
hamishivi: base_env_config = job-level defaults set via CLI (or whatever job-level config you have)
sample_config = sample-level overrides (set in dataset).

But you're right, I should make this not nullable. Editing so that base_env_config always exists (just with defaults from cli, e.g. max steps is 5 by default).

open_instruct/data_loader.py
hamishivi: Done! I refactored things a bit and now it should be simpler.

open_instruct/vllm_utils.py
hamishivi: fixed via just making the configs a datatype.

open_instruct/vllm_utils.py
hamishivi: done! Tried to make the logic a bit clearer.

open_instruct/data_loader.py
```diff
@@ -1242,6 +1255,17 @@ def _data_preparation_loop(self):
                 env_metrics: dict[str, dict[str, list[float]]] = {}
                 for rs in result.request_info.rollout_states:
                     info = rs.get("info", {})
+                    multi_env_metrics = info.get("env_metrics")
```
finbarrtimbers: Could this become a function? I worry that `_data_preparation_loop` is getting cluttered.

hamishivi: oh, yeah, sure.

PR #1503, opened 2026-03-02
Rename TIS ratio cap, add low bound and hard filter flag

finbarrtimbers:
Renames `truncated_importance_sampling_ratio_cap` to `truncated_importance_sampling_ratio_high` (default 2.0, now enabled by default), adds `truncated_importance_sampling_ratio_low` for a floor on the mismatch ratio, and adds `truncated_importance_sampling_hard_filter` to zero out tokens outside [low, high] instead of clamping them (the GLM-5 pop operator).

The hard filter flag implements the IcePop technique from GLM-5: rather than clamping the training-inference mismatch ratio ρ to [low, high], it zeroes out any token where ρ falls outside that range. The soft (clamp) behavior remains the default.

tl;dr, if we want to support GLM5-style GRPO, we can do so by setting

```
--truncated_importance_sampling_hard_filter \
--truncated_importance_sampling_ratio_high=2.0 \
--truncated_importance_sampling_ratio_low=0.5
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 All notable changes to this project will be documented in this file.
 
 
+### Added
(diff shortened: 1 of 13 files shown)


hamishivi (approved):

hamishivi (changes requested):
I like this PR, but we need to fix up the merge conflicts!

PR #1504, opened 2026-03-02
Remove cast from vLLM tool definitions typing

finbarrtimbers:
Fix type-safety around vLLM tool definitions so we can pass the tools payload to `ChatCompletionRequest` without casting or `# type: ignore` hacks.


open_instruct/environments/tools/parsers.py
```diff
@@ -127,7 +129,7 @@ def __init__(
         tool_parser: VllmNativeToolParser,
         role_templates: dict[str, str],
         stop_sequences: list[str] | None = None,
-        tool_definitions: list[dict[str, Any]] | None = None,
```
hamishivi, line 141: This is good, but then we should touch the Tool/RlEnv classes and ensure their tool defns follow this typing instead of dict, or make the cast more explicit somewhere.

hamishivi (changes requested):

hamishivi (approved):

PR #1506, opened 2026-03-02
Port vLLM v1 AsyncMPClient weight-update flow

mnoukhov:
This was rushed and made with codex, not yet verified!

## Summary
- switch Open Instruct vLLM weight-update RPC path to vLLM v1 AsyncMPClient.collective_rpc_async
- keep compatibility fallback paths for non-v1 clients
- preserve existing inflight_updates behavior while updating RPC execution path

## Reference
- PipelineRL vLLM v1 reference: https://github.com/ServiceNow/PipelineRL/blob/39830c47ad2466973ba405e55f72ae612271ae08/pipelinerl/vllm1.py

## Validation
- uv run pytest open_instruct/test_vllm_utils.py
- make style && make quality


hamishivi:
We have now done this in another PR, closing.

PR #1507, opened 2026-03-02
new vllm and transformers 5.2.0

mnoukhov:


hamishivi:
Clsoing since now main is on 0.19 vllm and 5.4.0 transformers.

PR #1514, opened 2026-03-06
Fix GPU test failures

finbarrtimbers:
Fixes four GPU test failures found by running the test suite on Beaker.

1. `get_num_tokens` in `padding_free_collator.py` only checked for an exact `attention_mask` key, but DPO batches use `chosen_attention_mask` and `rejected_attention_mask`. Fixed to match all keys ending with `attention_mask` and sum across branches.

2. `concatenated_forward_olmo` and `separate_forward_olmo` in `dpo_utils.py` were computing logps from raw logits, bypassing the TP-safe `DPOLMHead` path. Reverted to pass `labels` to the model and use `output.loss` for per-token logps. The GPU tests now swap the model's lm_head to `DPOLMHead` so the labels path works correctly.

3. `test_dpo_utils_gpu.py` used a hand-rolled mock `OlmoStyleModel` that didn't match the real olmo-core Transformer interface. Replaced with `TransformerConfig.olmo2_30M(vocab_size=1000)`.

4. `test_olmo_core_callbacks_gpu.py` patched a nonexistent `get_rank` attribute on the module. The module uses `distributed_utils.get_rank()`, so fixed the patch target accordingly.

5. Removed unused combined `input_ids` and `attention_mask` keys from `DataCollatorForSeq2SeqDPO` that were causing double-counting in `get_num_tokens`.

6. Fixed `drop_last=True` in the reference logprobs cache dataloader — it was silently dropping samples when the dataset size wasn't divisible by the cache batch size, causing a `Missing indices` error during caching. Changed to `drop_last=False` since the cache must process every example.

7. Fixed flaky `test_streaming_dataloader_iteration_with_tools` — switched from Qwen3-1.7B to Qwen3-0.6B to avoid vLLM EngineCore OOM, and added missing `base_env_config` parameter to both streaming dataloader tests.

8. Implemented world-aware packing for DPO: instead of distributing examples to ranks and letting each rank pack independently (which can produce different batch counts due to variable-length overflow), we now pack globally first and distribute packed batches round-robin to ranks. This eliminates the need for `_collect_batches_with_sync` in the reference cache builder.

9. Added a multi-rank GPU test for the reference logprobs cache builder that verifies world-aware packing produces no missing indices and identical results across ranks.

Also fixed an outdated GPU test launch path in AGENTS.md.

Runs:

1. Single GPU DPO: [Beaker](https://beaker.org/ex/01KKEQ0C0W5ZNNFZE6NARECHPK)
2. Multi-node DPO: [Beaker](https://beaker.org/ex/01KKEQJJA26F83BHJB6GM39YHW)

GPU_TESTS=[01KKCVXQ493AZYQZYKDGPD03TF](https://beaker.org/ex/01KKCVXQ493AZYQZYKDGPD03TF)

### DIFF
--- a/open_instruct/data_loader.py
+++ b/open_instruct/data_loader.py
@@ -143,6 +143,7 @@ class HFDataLoader(data_loader.DataLoaderBase):
         self._excluded_indices: set[int] = set()
         self._overflow: list[dict[str, Any]] = []
         self._precomputed_batch_sizes: list[int] | None = None
+        self._num_padding_batches: int = 0
         self._epoch: int = 0
         self._current_iter: Iterator[dict[str, Any]] | None = None
         self._device = device
--- a/open_instruct/dpo_utils.py
+++ b/open_instruct/dpo_utils.py
@@ -573,7 +573,7 @@ def build_reference_logprobs_cache(
                     model, batch, average_log_prob=average_log_prob, **(forward_kwargs or {})
                 )
 
-            if batch["is_padding"]:
+            if batch.get("is_padding", False):
                 continue
 
             chosen_tensor[batch["index"]] = chosen_logps
@@ -618,8 +618,7 @@ def build_reference_logprobs_cache(
         logger.info(f"Saving reference logprobs cache to {cache_path}")
         cache.to_disk(cache_path)
 
-    if dist.is_initialized():
-        dist.barrier()
+    dist.barrier()
 
     return cache
 


open_instruct/dpo.py
```diff
@@ -321,7 +321,7 @@ def main(args: dpo_utils.ExperimentConfig, tc: dataset_transformation.TokenizerC
         work_dir=args.output_dir,
         collator=collator,
         device=device,
-        drop_last=True,
```
hamishivi, line 323: Does this stuff up any actual training jobs?

open_instruct/dpo_utils.py
```diff
@@ -505,6 +505,30 @@ def _get_batch_stats(batch: dict) -> tuple[int, int, list[int], list[int]]:
     return batch_tokens, batch_size, chosen_lengths, rejected_lengths
 
 
+def _collect_batches_with_sync(
+    dataloader: torch.utils.data.DataLoader, device: torch.device
+) -> tuple[list[dict], int]:
+    """Collect all batches and sync count across ranks so all ranks iterate the same number of times.
+
+    With packing, different ranks can end up with different batch counts due to
```
hamishivi: Can we have size-aware packing to avoid this? Basically, assign samples in a round-robin fashion to each rank, and then pack them

open_instruct/dpo_utils.py
```diff
@@ -575,6 +603,9 @@ def build_reference_logprobs_cache(
                     model, batch, average_log_prob=average_log_prob, **(forward_kwargs or {})
                 )
 
+            if not is_real:
```
hamishivi: see above for world-aware packing.

open_instruct/vllm_utils.py
```diff
@@ -1283,6 +1283,7 @@ def create_vllm_engines(
                 scheduling_strategy=scheduling_strategy,
                 runtime_env=ray.runtime_env.RuntimeEnv(
                     env_vars={
+                        **dict(os.environ),
```
hamishivi, line 1286: nice!

hamishivi (commented):
Generally looks good, but I think theres potentially a nicer way to handle the packing?

open_instruct/dpo.py
finbarrtimbers: Yes, it does. It causes the cache to not have all elements in it, so a training run will eventually fail. I'm not sure why we didn't catch it until now.

finbarrtimbers: (Note that this is `cache_dataloader`, not `data_loader`.)

hamishivi: ohhhh nice

open_instruct/dpo_utils.py
finbarrtimbers: Ah, great suggestion. Done. Can you take a look to make sure I implemented what you had in mind?

open_instruct/dpo_utils.py
finbarrtimbers: Done.

open_instruct/vllm_utils.py
finbarrtimbers: Thank you!

open_instruct/data_loader.py
```diff
@@ -260,6 +290,84 @@ def _reshard(self, epoch: int) -> None:
...
+            else:
+                new_total = current_total + sample_lengths[i]
+                would_exceed = len(current_batch) > 0 and new_total > max_seq_length
+
+            at_max_samples = len(current_batch) >= self._per_rank_batch_size
+
+            if would_exceed or at_max_samples:
+                batches.append(current_batch)
+                current_batch = [i]
+                if is_dpo:
```
hamishivi: nit: `is_dpo` is a bit ugly, is there a nicer way to split the dpo and non-dpo paths? I can't think of anything on the top of my head...

open_instruct/data_loader.py
```diff
@@ -260,6 +290,84 @@ def _reshard(self, epoch: int) -> None:
...
+            batches.append(current_batch)
+
+        num_batches = len(batches)
+        if self._drop_last:
+            num_batches = (num_batches // self.dp_world_size) * self.dp_world_size
+            batches = batches[:num_batches]
+        else:
+            if (remainder := num_batches % self.dp_world_size) > 0:
+                for _ in range(self.dp_world_size - remainder):
+                    batches.append(batches[-1])
```
hamishivi, line 341: so we duplicate the last batch when drop_last is false? We should at least document this (is it possible to add dummy batches here with losses masked?). I probably would prefer drop_last over this in most cases.

hamishivi (commented):
Looks great, one comment.

hamishivi (approved):

PR #1529, opened 2026-03-16
Add Slack alerts documentation

finbarrtimbers:
Documents the Slack alert functionality in GRPO/DPO.

Also modifies DPO and GRPO so that they have the same interface (set the URL as `SLACK_WEBHOOK_URL`, and enable with `--send_slack_alerts`, which defaults to False).


mnoukhov:
can we add documentation of where to find the existing Ai2 webhook url? 

For our own conversations in slack, I've found that there's actually an email that goes directly there (check the conversation with yourself -> click on your own face icon in the top bar -> integrations). Do we want to support this or does the webook allow for DMs?

image

mnoukhov (approved):
lgtm!

PR #1532, opened 2026-03-17
GRPO Improve vllm timeout handling

BrownianNotion:
On slower networks (eg. runpod), I have been experiencing two issues related to timeouts:
1. Actor `_check_health` doesn't retry on TimeoutError
2. No exception handling in weight sync thread means `ray.get(actor_manager.set_should_stop.remote(False))` is never called and hangs training perpetually.

This PR should resolve both these issues.


BrownianNotion:
@hamishivi take a quick look when you can :) not sure if these are still helpful as you guys have already merged some potentially related fixes, but jic

hamishivi:
Yeah thanks! I'll take a look in a bit. Trying to clear out older PRs, but just picking easier things for now. I'll try to review on monday.

hamishivi (approved):

PR #1525, opened 2026-03-12
Offline Distillation via DistillKit (Part One - Compression Helpers for Logit Capture)

wolfecameron:
## Summary

This PR is **Part 1** of a large effort to add support for offline distillation into OpenInstruct

The full distillation rollout will add:
- teacher logit sampling with vLLM.
- student offline distillation training.
- distillation data handling/loss infrastructure and launch scripts.

This Part 1 PR intentionally scopes to the **compression primitives foundation only** (no sampling/training pipeline yet).

## PR Scope

Adds the core compression module needed by later distillation stages:

- `open_instruct/distillkit/compression/config.py` for distribution quantization config/types and validation.
- `open_instruct/distillkit/compression/bitpack.py` for bit-packing and unpacking token index payloads.
- `open_instruct/distillkit/compression/monotonic_logprobs.py` for monotonic logprob compression/decompression utilities.
- `open_instruct/distillkit/compression/compressor.py` for sparse top-k compression/decompression wrapper logic.
- `open_instruct/distillkit/compression/__init__.py` and `open_instruct/distillkit/__init__.py` minimal exports for this scoped subset.
- `configs/compression/default.yml` default compression config to be consumed by upcoming sampling/training stages.
- `open_instruct/test_distillkit_compression.py` focused unit tests covering bitpack roundtrip, config roundtrip, and exact-only sparse compressor roundtrip.

## Attribution
- Contribution work from Netflix Research: Cameron Wolfe (`cameronwolfe@netflix.com`)
- DistillKit adaptation: Portions of the compression implementation were adapted from [DistillKit](https://github.com/arcee-ai/DistillKit), with attribution headers preserved in the relevant ported files.

PR #1534, opened 2026-03-18
Offline Distillation via DistillKit (Part Two - Teacher Logit Capture)

wolfecameron:
## Summary
This PR is **Part 2** of a larger effort to add offline distillation support to OpenInstruct. Part one is [here](https://github.com/allenai/open-instruct/pull/1525). This PR (part two) provides a script for capturing teacher logits that can be used for distillation training.

## PR Scope
Adds vLLM-based teacher logit capture with compressed top-k outputs:
- `open_instruct/sample_logits_vllm.py`
  - loads tokenized SFT-style data via existing dataset transformation/cache flow
  - samples prompt logprobs from vLLM
  - converts prompt logprobs to dense top-k tensors
  - compresses with `LogprobCompressor`
  - writes parquet shards + compression config artifact
- `open_instruct/distillkit/sample_common.py`
  - parquet schema helper for compressed logit rows
  - streaming parquet writer + local output handler utility
- `open_instruct/distillkit/test_sample_common.py`
  - unittest for data writers
- `scripts/train/distill/olmo3_32b_teacher_sample_logits.sh`
  - example launch script for teacher logit sampling
- `pyproject.toml`
  - adds dependencies (`pyarrow`, `pyyaml`, `tqdm`)
  - adds `open_instruct/distillkit/sample_common.py` to `ty` include list

## Attribution
- Contribution work from Netflix Research: Cameron Wolfe (`cameronwolfe@netflix.com`)
- DistillKit adaptation: Portions of the sampling/compression implementation were adapted from [DistillKit](https://github.com/arcee-ai/DistillKit), with attribution headers preserved in relevant files.


open_instruct/distillkit/sample_common.py
```diff
@@ -0,0 +1,113 @@
...
+    def write(self, row_data: dict[str, Any]) -> None:
+        """Enqueue one row for asynchronous parquet writing."""
+        self.queue.put(row_data)
+
+    def __enter__(self) -> "StreamingParquetWriter":
+        """Start the writer thread."""
+        self.thread.start()
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> bool:
```
finbarrtimbers: What does the returned bool represent?

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,305 @@
...
+from open_instruct.distillkit.sample_common import StreamingParquetWriter, compressed_logit_schema
+from open_instruct.utils import ArgumentParserPlus
+
+logger = logger_utils.setup_logger(__name__)
+
+LOGIT_SAMPLING_TARGET_COLUMNS = ["input_ids", "attention_mask", "labels", "messages", "text"]
+
+
+@dataclass
+class SampleLogitsArguments:
```
finbarrtimbers, line 47: Let's avoid using the `field` function here unless absolutely necessary; I think a number of the metadata is redundant given the helpful names. 

For the ones that aren't obvious, can we select more informative names to make the metadata redundant? e.g. `dtype` -> `model_dtype`, or `compression_config` -> `compression_config_path`?

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,305 @@
...
+    model_name_or_path: str = field(metadata={"help": "Path to model to sample logits from."})
+    compression_config: str = field(metadata={"help": "Path to compression config YAML file."})
+
+    model_revision: str | None = field(default=None, metadata={"help": "Optional model revision."})
+    dtype: str | None = field(default=None, metadata={"help": "Model dtype (e.g., bfloat16)."})
+    quantization: str | None = field(default=None, metadata={"help": "Optional vLLM quantization mode."})
+    tensor_parallel_size: int = field(default=1, metadata={"help": "Tensor parallel world size."})
+    pipeline_parallel_size: int = field(default=1, metadata={"help": "Pipeline parallel world size."})
+    enable_expert_parallel: bool = field(default=False, metadata={"help": "Enable expert parallelism for MoE."})
```
finbarrtimbers: Do we need this and `pipeline_parallel_size`? Let's instead just check if `pipeline_parallel_size > 1`.

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,305 @@
...
+    dataset_target_columns: list[str] = field(default_factory=lambda: LOGIT_SAMPLING_TARGET_COLUMNS.copy())
+    dataset_cache_mode: Literal["hf", "local"] = "local"
+    dataset_local_cache_dir: str = "local_dataset_cache"
+    dataset_config_hash: str | None = None
+    dataset_skip_cache: bool = False
+
+    max_seq_length: int = field(default=2048, metadata={"help": "Max context length for tokenization and vLLM."})
+    max_samples: int | None = field(default=None, metadata={"help": "Optional max samples to process."})
+    seed: int = field(default=42, metadata={"help": "Dataset shuffle seed."})
+    macrobatch_size: int = field(default=256, metadata={"help": "Number of prompts per vLLM call."})
```
finbarrtimbers: Let's instead use `num_unique_prompts` here, as that's the variable name we use in our GRPO implementation.

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,305 @@
...
+    max_workers: int | None = field(default=None, metadata={"help": "Background CPU writer workers."})
+
+    auto_vocab_size: bool = field(default=True, metadata={"help": "Auto-fix config vocab size to tokenizer vocab."})
+    use_flat_logprobs: bool = field(default=True, metadata={"help": "Use FlatLogprobs path when available."})
```
finbarrtimbers: Why would we want to use or not use this?

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,305 @@
...
+
+    output_dir: str = field(default="output/logits", metadata={"help": "Output root directory."})
+    exp_name: str = field(default="sample_logits_vllm", metadata={"help": "Output experiment name."})
+
+
+def process_prompt_logprobs(prompt_logprobs: PromptLogprobs, k: int) -> tuple[torch.Tensor, torch.Tensor]:
+    """Extract dense top-k token ids/logprobs from vLLM prompt logprob outputs."""
+    if isinstance(prompt_logprobs, FlatLogprobs):
```
finbarrtimbers: Can you make this into a separate function and dispatch to it? Or is there some way to combine the two branches?

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,305 @@
...
+    top_values = torch.full((len(valid_logprobs), k), fill_value=float("-inf"), dtype=torch.float32)
+    for seq_id, logprobs in enumerate(valid_logprobs):
+        if logprobs is None:
+            raise ValueError(f"Missing logprobs for token at position {seq_id + 1}.")
+        for tok_id, logprob in logprobs.items():
+            if logprob.rank is None or logprob.rank > k:
+                continue
+            top_indices[seq_id, logprob.rank - 1] = tok_id
+            top_values[seq_id, logprob.rank - 1] = logprob.logprob
+    return top_indices, top_values
```
finbarrtimbers, line 129: Can you add some comments here explaining what is going on?

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,305 @@
...
+
+
+def main(args: SampleLogitsArguments, tc: TokenizerConfig) -> None:
+    """Run vLLM prompt-logprob sampling and write compressed outputs to parquet."""
+    tc.tokenizer_revision = args.model_revision if tc.tokenizer_revision is None else tc.tokenizer_revision
```
finbarrtimbers, line 274: You could just do: 

```
tc.tokenizer_revision = tc.tokenizer_revision or args.model_revision
```

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,305 @@
...
+    tc.tokenizer_name_or_path = (
```
finbarrtimbers, line 275: Same here.

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,305 @@
...
+        args.model_name_or_path if tc.tokenizer_name_or_path is None else tc.tokenizer_name_or_path
+    )
+    tokenizer = tc.tokenizer
```
finbarrtimbers, line 278: Inline this, and refer to tc.tokenizer throughout?

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,305 @@
...
+        "trust_remote_code": tc.trust_remote_code,
+        "tensor_parallel_size": args.tensor_parallel_size,
+        "pipeline_parallel_size": args.pipeline_parallel_size,
+        "enable_expert_parallel": args.enable_expert_parallel,
+        "gpu_memory_utilization": args.gpu_memory_utilization,
+        "max_logprobs": k,
+        "logprobs_mode": "raw_logprobs",
+        "max_model_len": args.max_seq_length,
+    }
+    if args.dtype is not None:
```
finbarrtimbers: What happens if you pass None in here?

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,305 @@
...
+        presence_penalty=0.0,
+        repetition_penalty=1.0,
+        prompt_logprobs=k,
+        logprobs=k,
+        flat_logprobs=args.use_flat_logprobs,
+        max_tokens=1,
+        detokenize=False,
+        skip_special_tokens=False,
+    )
+
```
finbarrtimbers: Are any of these the default values? Do we need to specify all of these?

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,305 @@
...
+        row_out = compressor.compress_from_sparse(top_indices, top_values)
+        writer.write(
+            {
+                "input_ids": input_ids_sample if isinstance(input_ids_sample, list) else input_ids_sample.tolist(),
+                "compressed_logprobs": row_out["compressed_logprobs"].cpu().squeeze(0).tolist(),
+                "bytepacked_indices": row_out["bytepacked_indices"].cpu().squeeze(0).tolist(),
+                "messages": json.dumps(messages_sample) if messages_sample else "",
+                "text": text_sample or "",
+            }
+        )
```
finbarrtimbers: Please break this out to the file level context.

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,305 @@
...
+
+    output_dir = os.path.join(args.output_dir, args.exp_name)
+    try:
```
finbarrtimbers: Why do you need a try/finally here?

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,305 @@
...
+            {
+                "input_ids": input_ids_sample if isinstance(input_ids_sample, list) else input_ids_sample.tolist(),
+                "compressed_logprobs": row_out["compressed_logprobs"].cpu().squeeze(0).tolist(),
+                "bytepacked_indices": row_out["bytepacked_indices"].cpu().squeeze(0).tolist(),
+                "messages": json.dumps(messages_sample) if messages_sample else "",
+                "text": text_sample or "",
+            }
+        )
+
+    output_dir = os.path.join(args.output_dir, args.exp_name)
```
finbarrtimbers: Can you use Pathlib instead?

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,305 @@
...
+    try:
+        os.makedirs(output_dir, exist_ok=True)
+        with open(os.path.join(output_dir, "compression_config.yaml"), "w", encoding="utf-8") as f:
+            yaml.safe_dump(cfg.to_dict(), f, sort_keys=False)
+
+        with (
```
finbarrtimbers: Generally, this chunk of code is quite dense. Could you break it out into functions?

open_instruct/distillkit/sample_common.py
wolfecameron: This is context manager syntax in python for how you should handle exceptions:

True -> suppress exceptions
False -> propagate exceptions

So, returning false here basically means we don't want to suppress exceptions that occur in the context manager. However, upon looking into it further, returning None has the same behavior. So, I'll probably tweak this to just return nothing and add a comment explaining the behavior.

open_instruct/sample_logits_vllm.py
wolfecameron: Makes sense - I will update the names mentioned and change to using normal default and only retain field whenever the default factory is needed (e.g., for list defaults).

open_instruct/sample_logits_vllm.py
wolfecameron: I think we still need both of them right? One is for expert parallel (vllm infers size automatically but you have to enable it) and the other is for pipeline parallel.

open_instruct/sample_logits_vllm.py
wolfecameron: Earlier versions of vLLM did not have support for flat logprobs. Now, I think it's been around for long enough to assume this is supported and remove the legacy codepath. Will make this change.

open_instruct/sample_logits_vllm.py
wolfecameron: Do you mean the "messages" / "text" specifically? These are unique per prompt, so I don't think this would save any memory. And, the file size would be dominated by logprobs.

I could write this to a separate file that is linked by a shared id. However, this might make data loading harder for the actual distillation script in the future. For example, having the actual messages available is needed to compute the completion-only attention mask when pretokenizing the training dataset.

open_instruct/sample_logits_vllm.py
wolfecameron: good point - can be removed

wolfecameron:
Fixed the following items:
- remove return False from context manager __exit__ and return None instead
- dataclass cleanup (remove unnecessary field) + more informative renames
- rename macrobatch_size to num_unique_prompts
- assume flat logprobs always used
- remove unnecessary try/finally blocks
- use pathlib instead of os
- break sampling loop block into helper function to make it less dense

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,336 @@
...
+        for i in range(start_idx, end_idx):
+            rank = prompt_logprobs.ranks[i]
+            if rank is None or rank > k:
+                continue
+            seq_ids.append(seq_id)
+            rank_ids.append(rank - 1)
+            token_ids_to_copy.append(prompt_logprobs.token_ids[i])
+            logprobs_to_copy.append(prompt_logprobs.logprobs[i])
+
+    if seq_ids:
```
finbarrtimbers: When would seq_id not be truthy?

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,336 @@
...
+
+    # FlatLogprobs has flattened arrays with per-position start/end pointers
+    # use this to rebuild token position / rank coordinates
+    for pos_id in range(start_pos, len(prompt_logprobs)):
+        seq_id = pos_id - start_pos
+        start_idx = prompt_logprobs.start_indices[pos_id]
+        end_idx = prompt_logprobs.end_indices[pos_id]
+        for i in range(start_idx, end_idx):
+            rank = prompt_logprobs.ranks[i]
+            if rank is None or rank > k:
```
finbarrtimbers, line 115: When would rank be None? 

(if we hit rank > k, we can break, right?)

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,336 @@
...
+def load_and_validate_compression_config(
+    compression_config_path: str, tokenizer: Any, auto_vocab_size: bool
+) -> DistributionQuantizationConfig:
+    """Load compression config and align vocab sizing with tokenizer when requested."""
+    with Path(compression_config_path).open(encoding="utf-8") as f:
+        cfg = DistributionQuantizationConfig.from_dict(yaml.safe_load(f))
+
+    tok_vocab = tokenizer.get_vocab()
+    tok_vocab_size = max(len(tok_vocab) + 1, max(tok_vocab.values()))
+    if cfg.d != tok_vocab_size:
```
finbarrtimbers: What if you change this to be: 

```
if cfg.d == tok_vocab_size:
    # Nothing to do in this case
    pass
elif auto_vocab_size:
    ....
```

then we remove a level of nesting. Wdyt?

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,336 @@
...
+        ) as writer,
+        ThreadPoolExecutor(max_workers=max_workers) as executor,
+    ):
+        futures = []
+        for i0 in tqdm.tqdm(range(0, len(train_dataset), num_unique_prompts), desc="Logit Batches"):
+            batch_slice = train_dataset[i0 : i0 + num_unique_prompts]
+
+            # Convert HF dataset rows into vLLM prompt_token_ids payloads.
+            prompts = [
+                {"prompt_token_ids": x.tolist() if hasattr(x, "tolist") else list(x)}
```
finbarrtimbers: When does it have a tolist attribute? Any way we can predetermine this? Can we call `list(x)` always?

open_instruct/sample_logits_vllm.py
```diff
@@ -0,0 +1,336 @@
...
+                        req_out,
+                        input_ids_sample,
+                        messages_sample,
+                        text_sample,
+                        writer,
+                        compressor,
+                        k,
+                    )
+                )
+            while len(futures) > num_unique_prompts * 2:
```
finbarrtimbers, line 266: Can you add a comment describing why this is needed?

open_instruct/sample_logits_vllm.py
wolfecameron: Left a comment to clarify this and changed the check to "if token_ids_to_copy:" instead because that might be clearer. Basically will be None if there is no entry with rank < k (i.e., no token ids to copy). Theoretically, this shouldn't be the case, unless we do something like setting k = 0. I can also remove this if you'd prefer that.

open_instruct/sample_logits_vllm.py
wolfecameron: FlatLogprobs isn't ordered by rank in this case - we're basically stepping through the vector component by component (in the original ordering) and reading the rank of that component. So I don't think we can break here because we aren't traversing in rank order. 

We could iterate in rank order, but this would require sorting, which would cause more overhead than just traversing the list. 

In terms of being None, vLLM in their docs has the return type as int | None. I'm not sure when it would be None, but because the method has None within the return types I think it's probably better to leave it. 

Adding a comment to clarify these points

open_instruct/sample_logits_vllm.py
wolfecameron: tolist is a tensor attribute. I don't think we can change logic to list(x) because this has different behavior for tensors; e.g., 

x = torch.tensor([1, 2, 3, 4])
x.tolist()
-> [1, 2, 3, 4]
list(x)
-> [tensor(1), tensor(2), tensor(3), tensor(4)]

But, this is a good callout on being vague. I'm changing this to explicitly clarify the type expectation of:

if tensor -> x.tolist()
else -> list(x)

open_instruct/sample_logits_vllm.py
finbarrtimbers: > In terms of being None, vLLM in their docs has the return type as int | None. I'm not sure when it would be None, but because the method has None within the return types I think it's probably better to leave it.

Ah, that's obnoxious (on vllm's part). Leaving it sounds good.

open_instruct/sample_logits_vllm.py
finbarrtimbers: I'd prefer removing it, then!

open_instruct/sample_logits_vllm.py
finbarrtimbers: Thank you, I like that!

PR #1540, opened 2026-03-19
Log gradient norm in grpo_fast

mnoukhov:
## Summary
- log DeepSpeed gradient norm in grpo_fast after optimizer updates
- surface non-finite grad norms in the logged metric instead of filtering them out
- document that the collected grad norms may include nan/inf values

## Testing
- make style
- make quality

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -747,8 +747,7 @@ class PolicyTrainerRayProcess(RayProcess):
             batch_metrics = batch_data["metrics"]
             with torch.no_grad():
                 self._compute_loss_metrics(loss_stats_B, total_valid_tokens)
-                if grad_norms:
-                    self.local_metrics["optim/grad_norm"] = sum(grad_norms) / len(grad_norms)
+                self.local_metrics["optim/grad_norm"] = sum(grad_norms) / len(grad_norms)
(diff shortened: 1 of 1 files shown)


open_instruct/grpo_fast.py
```diff
@@ -725,6 +726,9 @@ def step(self):
                     self.model.backward(loss)
                     if is_accumulation_boundary:
                         self.model.step()
+                        grad_norm = self.model.get_global_grad_norm()
```
finbarrtimbers: When would this be None?

mnoukhov: I also thought that was weird and codex initially convinced me it was possible in theory

but I now realize that as long as we're using a decently normal optimizer, it won't be None so I will remove the check

open_instruct/grpo_fast.py
```diff
@@ -745,6 +747,8 @@ def step(self):
             batch_metrics = batch_data["metrics"]
             with torch.no_grad():
                 self._compute_loss_metrics(loss_stats_B, total_valid_tokens)
+                if grad_norms:
```
finbarrtimbers: Don't think this is needed?

mnoukhov: you're right, removing

PR #1542, opened 2026-03-20
Fix total_batch_size logging for sequence parallelism

hamishivi:
fixing logging on the batch size - taking into account SP.


open_instruct/finetune.py
```diff
@@ -730,7 +730,8 @@ def collate_fn(features):
         checkpointing_steps = int(checkpointing_steps)
 
     # Train!
-    total_batch_size = args.per_device_train_batch_size * accelerator.num_processes * args.gradient_accumulation_steps
+    dp_world_size = accelerator.num_processes // args.sequence_parallel_size
```
finbarrtimbers, line 733: Up to you ofc but I think we should refactor this somewhere so we can share it across DPO, GRPO, SFT

PR #1543, opened 2026-03-20
GRPO with Olmo-Core working

mnoukhov:
Added a local debug script `grpo.sh` that avoids using `ray_node_setup.sh` for simplicity and fixed up all the issues so it runs.

Still some confusing issues to figure out


open_instruct/grpo_utils.py
```diff
@@ -373,7 +395,9 @@ def calculate_token_counts(
     Copied from grpo_fast.py to share logic with olmo_core_train_modules.py.
     """
     accumulation_counts: dict[int, float] = {}
-    local_counts = [mask[:, 1:].sum().float() for mask in data_BT.response_masks]
+    local_counts = []
```
finbarrtimbers: Can we keep this as a comprehension?

open_instruct/olmo_core_train_modules.py
```diff
@@ -391,11 +399,15 @@ def train_batch(self, batch: dict[str, Any], dry_run: bool = False) -> None:
 
         for _epoch_idx in range(self.grpo_config.num_epochs):
             for sample_idx in range(num_samples):
+                query_responses = data_BT.query_responses[sample_idx]
```
finbarrtimbers: Let's keep these inlined!

open_instruct/grpo_utils.py
mnoukhov: done!

PR #1560, opened 2026-03-26
Switch back to CUDA 12.8.1, change libraries to cu128, pin cu128 FA3 wheels

mnoukhov:
## Summary
- switch Linux x86_64 torch resolution from cu129 to cu128 to match the working rlzero setup
- replace the flat flash-attn-3 wheel index with a direct Linux x86_64 wheel URL to avoid resolving to newer aarch64-only releases
- regenerate uv.lock and requirements.txt with the updated torch and NVIDIA cu12 package set

## Validation
- make style
- make quality

GPU_TESTS=bypass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,11 +30,8 @@ All notable changes to this project will be documented in this file.
 - Extended CONTRIBUTING.md with documentation on running tests, CI workflows, Beaker experiments, GRPO/DPO test scripts, and environment variables.
 
 ### Changed
-<<<<<<< michaeln/cuda128-fa3-wheel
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -30,7 +30,11 @@ All notable changes to this project will be documented in this file.
 - Extended CONTRIBUTING.md with documentation on running tests, CI workflows, Beaker experiments, GRPO/DPO test scripts, and environment variables.
 
 ### Changed
+<<<<<<< michaeln/cuda128-fa3-wheel
```
finbarrtimbers: Merge artifact!

PR #1565, opened 2026-03-26
Remove configure_hf_hub_retry (HF Hub handles retries internally)

finbarrtimbers:
## Summary
- Remove `configure_hf_hub_retry()` function from `utils.py` and all its callers in `grpo_fast.py` and `vllm_utils.py`
- Remove unused imports (`HTTPAdapter`, `Retry`)

## Context
`configure_hf_hub_retry()` was added in PR #1528 to prevent HF rate-limit crashes during Ray actor spawning. It called `huggingface_hub.configure_http_backend()` to inject a custom `requests.Session` with urllib3 `Retry`.

However, `configure_http_backend()` was **removed in huggingface_hub v1.0.0** (Oct 2024) when the library migrated from `requests` to `httpx`. Our pinned version (v1.7.2) no longer has this function, so **`configure_hf_hub_retry()` would crash at runtime if called**.

This is safe to remove because:
1. huggingface_hub v1.0+ has built-in retry logic via `http_backoff()` — same behavior (429/5xx, exponential backoff)
2. The real fix from PR #1528 (pre-downloading the model before Ray actors spawn) is still in place

GPU_TESTS=bypass

## Test plan
- [ ] Verify GRPO and vLLM inference work without explicit retry configuration

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file.
 - OLMo-core GRPO actor with Ray-distributed FSDP2 training (https://github.com/allenai/open-instruct/pull/1398).
 
 ### Fixed
+- Remove `configure_hf_hub_retry`; HF Hub handles retries internally (https://github.com/allenai/open-instruct/pull/1565).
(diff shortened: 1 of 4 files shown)


finbarrtimbers:
Moving this into PR #1425 — the removal must land together with the huggingface_hub upgrade (v1.7.2) that makes configure_http_backend unavailable.

PR #1567, opened 2026-03-27
Automatically configures flash attention

finbarrtimbers:
## Summary
- Remove `use_flash_attn` and `attn_backend` flags, auto-detect attention implementation from model config
- Add `attn_implementation` to `ModelConfig` with auto-detect default
- Unify attention backend detection across DPO, GRPO, and SFT codepaths
- Fix GPU test deadlock: start vLLM engine process group before local init
- Make dataset transformation tests fully offline with local fixtures

## DPO Runs

1. Single GPU DPO: [Beaker](https://beaker.org/ex/01KMRNQ25RBPJQ92Q84YAFQ24Q) - passed
2. Multi-node DPO: [Beaker](https://beaker.org/ex/01KMRP74SDEEA5ZCY5H9RJ3E8M) - hung (stuck at reference model `full_tensor().cpu()` after ~20min, no progress for 10+ hours)

## GRPO Runs

1. Single GPU GRPO: [Beaker](https://beaker.org/ex/01KMTH948Y57DJYE4R0MQJ4TWG) - passed
2. Single GPU GRPO with tools: skipped (missing Beaker secrets `finbarrt_SERPER_API_KEY`/`finbarrt_JINA_API_KEY`)
3. Multi-node GRPO: [Beaker](https://beaker.org/ex/01KMTHKBF734K6ZFCX6Y2J8PWS) - passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 
 ### Added
+- Auto-detect attention implementation from model config, removing `use_flash_attn` and `attn_backend` flags; add `flash-attn` v2 fallback for Blackwell GPU support (https://github.com/allenai/open-instruct/pull/1567).
 - Add hybrid model (Olmo-Hybrid) support: MambaSpec monkey-patch for vLLM dtype serialization, `trust_remote_code` pass-through to vLLM engines, `get_text_config()` for multimodal model support, dependency upgrades (vllm>=0.18.0, transformers>=5.3.0), `return_dict=False` for transformers 5.x compat, and hybrid test/production training scripts (https://github.com/allenai/open-instruct/pull/1425).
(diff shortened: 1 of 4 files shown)


open_instruct/model_utils.py
```diff
@@ -44,6 +45,30 @@
...
+@functools.lru_cache(maxsize=1)
+def detect_attn_implementation() -> str:
+    if transformers.utils.is_flash_attn_3_available():
+        result = "flash_attention_3"
+    elif transformers.utils.is_flash_attn_2_available():
+        result = "flash_attention_2"
+    else:
+        result = "sdpa"
+    logger.info(f"Auto-detected attention implementation: {result}")
+    return result
```
mnoukhov, line 72: Noticed that Flash Attention 3 only supports Hopper and doesn't currently support Blackwell GPUs!

If we want to run on B200s we should add back flash-attn-2 into pyproject.toml and double-check that we're on Hopper when activating FA3, otherwise drop down to FA2

finbarrtimbers: Added! Good point. I'll also add fa4 support in a subsequent PR.

PR #1569, opened 2026-03-30
Adds support for flash attention 4, and changes default to fa2 while we figure out learning issues. 

finbarrtimbers:
## Summary
- Adds Flash Attention 4 support for Blackwell GPUs (SM 10.0+), with auto-detection via flash_attn.cute
- Unifies attention backend representation on olmo_core.nn.attention.AttentionBackendName instead of HF string literals, eliminating duplicate type definitions
- Detection priority: FA2 (default) > FA4 (Blackwell) > FA3 (Hopper) > SDPA (fallback)
- Removes ai2/ceres from single GPU test cluster list

## Test plan
- [x] Single GPU GRPO on Beaker: [01KMZRRWCVK0MCMBC46E6AK31A](https://beaker.org/ex/01KMZRRWCVK0MCMBC46E6AK31A) (exit code 0)
- [ ] Multi-GPU and multi-node tests (pending)

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/open_instruct/model_utils.py
+++ b/open_instruct/model_utils.py
@@ -69,27 +69,24 @@ def _is_flash_attn_4_available() -> bool:
     return importlib.util.find_spec("flash_attn.cute") is not None
 
 
-def _is_blackwell_gpu() -> bool:
+@functools.lru_cache(maxsize=1)
+def _gpu_compute_major() -> int | None:
+    """Return the CUDA compute capability major version (e.g. 9=Hopper, 10=Blackwell)."""
     if not torch.cuda.is_available():
-        return False
+        return None
     major, _ = torch.cuda.get_device_capability()
-    return major >= 10
-
-
-def _is_hopper_gpu() -> bool:
-    if not torch.cuda.is_available():
-        return False
-    major, _ = torch.cuda.get_device_capability()
-    return major == 9
+    return major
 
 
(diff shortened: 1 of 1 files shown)


open_instruct/grpo_olmo_core_actor.py
```diff
@@ -57,7 +57,7 @@ def __init__(
         vllm_config: data_loader_lib.VLLMConfig,
         data_prep_actor_name: str,
         tokenizer: transformers.PreTrainedTokenizer,
-        attn_implementation: str = "flash_attention_3",
+        attn_implementation: model_utils.AttentionBackendName = model_utils.AttentionBackendName.flash_3,
```
mnoukhov: I don't think this needs a default value

open_instruct/grpo_olmo_core_actor.py
```diff
@@ -329,7 +327,7 @@ def __init__(
...
```
mnoukhov: ditto

finbarrtimbers: Nice catch, removed.

open_instruct/grpo_olmo_core_actor.py
finbarrtimbers: Nice catch, removed.

PR #1574, opened 2026-03-31
Add deprecation warning to finetune.py

finbarrtimbers:
## Summary
- Adds a `DeprecationWarning` to `finetune.py` directing users to the OLMo-core SFT implementation

## Test plan
- [x] `make style && make quality` passes
- [x] Change is additive-only (warning at startup), no functional impact

GPU_TESTS=bypass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 All notable changes to this project will be documented in this file.
 
 
+### Deprecated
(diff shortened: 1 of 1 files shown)


open_instruct/finetune.py
```diff
@@ -1022,6 +1023,12 @@ def collate_fn(features):
 
 
 if __name__ == "__main__":
+    warnings.warn(
```
hamishivi, line 1026: Seems good to me. I think finetune.py works fine these days, although. Let's not fully remove until we have a dpo-style integration.

PR #1576, opened 2026-03-31
Extract shared OLMo-core config classes and helpers

finbarrtimbers:
## Summary
- Extract shared config dataclasses (`ExperimentConfig`, `ModelConfig`, `TrainingConfig`, `DatasetConfig`, `LoggingConfig`, `CheckpointConfig`) and helper functions (`setup_model`, `load_dataset_distributed`, `setup_tokenizer_and_cache`, `build_checkpointer_callback`) into `olmo_core_utils.py`
- Add `max_seq_length` parameter to `HFDataLoader` for correct token-based batch validation
- Remove unused `CollatedBatchData.to()` method (now handled inline by callers)
- Remove unused `TestFlatArguments` test class
- Refactor DPO to use shared OLMo-core configs: replace local config classes in `dpo_utils.py` with imports from `olmo_core_utils`, rename `ExperimentConfig` to `DPOExperimentConfig`, and replace local helper functions in `dpo.py` with shared utilities

GPU_TESTS=[01KN308P2VFX4RXAW9YFKFQPQV](https://beaker.org/ex/01KN308P2VFX4RXAW9YFKFQPQV)

## Test plan
- [x] `make style && make quality` passes
- [x] `uv run pytest open_instruct/test_utils.py` passes
- [x] `uv run pytest open_instruct/test_dpo_utils.py` passes
- [x] GPU tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1578, opened 2026-03-31
Refactor GRPO: inline utilities and rename GRPOExperimentConfig

finbarrtimbers:
## Summary
- Rename `ExperimentConfig` to `GRPOExperimentConfig` throughout GRPO code
- Inline `mask_logprobs`, `compute_tis_weights`, `resolve_old_logprob`, `create_loss_stats`, `populate_sample_loss_stats`, and `compute_metrics_from_loss_stats` from `grpo_utils.py` into `grpo_fast.py` and `olmo_core_train_modules.py`
- Simplify metric computation in `GRPOTrainModule` to use explicit token-weighted averaging instead of generic helper functions
- Update `benchmark_generators.py` and test files to use the new config name

**Depends on #1576**

## Test plan
- [ ] `make style && make quality` passes
- [ ] `uv run pytest open_instruct/test_olmo_core_train_modules.py` passes
- [ ] GPU tests pass


open_instruct/grpo_fast.py
```diff
@@ -473,12 +473,11 @@ def calculate_token_counts(
 
         return accumulation_counts
 
-    def _compute_loss_metrics(self, loss_stats_B: dict[str, torch.Tensor], total_valid_tokens: int) -> None:
+    def _compute_loss_metrics(self, loss_stats_B: dict[str, torch.Tensor]) -> None:
         metrics = grpo_utils.compute_metrics_from_loss_stats(loss_stats_B, self.args)
         for k, v in metrics.items():
             self.local_metrics[k] = v
         self.local_metrics["lr"] = self.scheduler.get_last_lr()[0]
-        self.local_metrics["_token_count"] = total_valid_tokens
```
finbarrtimbers, line 479: This was redundant!

open_instruct/grpo_fast.py
```diff
@@ -543,7 +542,6 @@ def step(self):
         # Pre-compute token counts per sample (for weighted averaging across SP ranks)
         # This only needs to be done once since response_masks don't change across epochs
         token_counts_per_sample = torch.stack([mask[:, 1:].sum().float() for mask in data_BT.response_masks])
-        total_valid_tokens = token_counts_per_sample.sum().item()
```
finbarrtimbers, line 544: Also redundant!

open_instruct/olmo_core_train_modules.py
```diff
@@ -363,16 +363,6 @@ def train_batch(self, batch: dict[str, Any], dry_run: bool = False) -> None:
             else:
                 ref_logprobs_BT = None
 
-        with torch.no_grad():
```
finbarrtimbers, line 368: This was redundant, and the result was thrown away on the next line: 

1. Lines 370-391: Initializes old_logprobs_BT as a list of Nones, then:
    - If num_mini_batches > 1: recomputes logprobs (either from compute_logprobs or from vllm_logprobs) and
   fills the list.
    - If num_mini_batches == 1: leaves them all as None.
  2. Line 430: resolve_old_logprob() handles the None case — when num_mini_batches == 1 (or first epoch),
  it falls back to using vllm_logprobs or new_logprobs as the old logprob.

  So the deleted block was computing old_logprobs_BT into a variable that was immediately overwritten on
  the very next line (old_logprobs_BT: list[...] = [None for ...]). The computation was wasted — it was
  never used. It was likely left over from an earlier version of the code before the mini-batch logic was
  added.

open_instruct/grpo_utils.py
```diff
@@ -462,19 +462,21 @@ def calculate_token_counts(
     return accumulation_counts
 
 
+_SCALAR_LOSS_STAT_KEYS = ["kl_loss", "tis_ratio", "tis_clipfrac", "pg_clipfrac", "pg_loss", "loss", "ratio", "entropy"]
+
+_WEIGHTED_AVG_METRICS = [
```
hamishivi: is there a nicer way to do this? I feel like this is a bit overengineered - its hard to work out what the two entries in each list item are, etc.

open_instruct/olmo_core_train_modules.py
hamishivi: yeah old code was me being a bit defensive about keeping the same code path pre-fixing a bug.

hamishivi (commented):
One comment, rest looks good

open_instruct/grpo_utils.py
finbarrtimbers: Ok, PTAL!

open_instruct/grpo_utils.py
```diff
@@ -489,50 +494,40 @@ def populate_sample_loss_stats(
...
+    loss_stats_B: dict[str, torch.Tensor], token_counts: torch.Tensor
 ) -> dict[str, float]:
-    token_counts = loss_stats_B["token_count"]
     total_tokens = token_counts.sum()
     weights = token_counts / total_tokens if total_tokens > 0 else torch.zeros_like(token_counts)
 
     metrics: dict[str, float] = {}
-    if config.load_ref_policy:
-        for j in range(4):
-            metrics[f"objective/kl{j}_avg"] = (loss_stats_B["kl"][j] * weights).sum().item()
```
hamishivi, line 516: I think we also need to remove the `"objective/kl{j}_avg"` keys, or  also sum/norm them too, since they are in `loss_stats_B` but not normalised?

finbarrtimbers: I don't think so, as these are already normalized? I don't think this PR changed how we treat them at all. I could be wrong. Looking at [wandb](https://wandb.ai/ai2-llm/open_instruct_internal/workspace?nw=wubv57xoyj) the runs look similar.

finbarrtimbers: I took some time to manually trace through the code. Initially, we had this: 

```
loss_stats_B["kl"][:, sample_idx] = masked_mean(kl_4BT, response_mask).float()
loss_stats_B["kl_loss"][sample_idx] = loss_stats_B["kl"]
```

which later on became:

```
if config.load_ref_policy:
        for j in range(4):
            metrics[f"objective/kl{j}_avg"] = (loss_stats_B["kl"][j] * weights).sum().item()
metrics["loss/kl_avg"] = (loss_stats_B["kl_loss"] * weights).sum().item()
```

In the current version of the code, we include `"objective/kl{i}_avg"` in the keys and handle it with the rest of the values; in particular, we do:

```
kl_4BT = model_utils.estimate_kl(ref_logprobs_diff, ratio)
kl_values = masked_mean(kl_4BT, response_mask).float()
for j in range(4):
    loss_stats_B[f"objective/kl{j}_avg"][sample_idx] = kl_values[j]
loss_stats_B["loss/kl_avg"][sample_idx] = kl_values[config.kl_estimator] * config.beta
```

and then average with the other keys:

```
for key in loss_stats_B:
        metrics[key] = (loss_stats_B[key] * weights).sum().item()
```

So this seems equivalent to me!

hamishivi: Ah yep. I think the main thing is that now `populate_sample_loss_stats` creates the klX stats, rather than this func, which is more correct anyway.

Issue #1431, opened 2026-01-27
Single-node RLVR run hangs waiting for placement group / worker crash on Slurm


LalchandPandia:
Hi,

I am trying to run the RLVR code on a single node using the provided grpo_fast_8b_single_node.sh script, but the job either hangs indefinitely or crashes with a Ray worker error when launched under Slurm.

Below is my Slurm-side Ray setup (single node, 8 GPUs):
nodes_array=($nodes)
head_node=${nodes_array[0]}
NUM_GPUS=8

unset RAY_ADDRESS RAY_HEAD_NODE_ADDRESS RAY_NAMESPACE
export RAY_TEMP_DIR=/net/${SLURM_JOB_ID}
export MASTER_PORT=$(shuf -i 29500-29999 -n 1)
export RAY_NODE_PORT=$MASTER_PORT
export ip_head=$head_node:$MASTER_PORT
export RAY_ADDRESS=$head_node:$RAY_NODE_PORT

echo "IP Head: $ip_head"

ray stop --force

ray start --head \
  --node-ip-address=$head_node \
  --port=$RAY_NODE_PORT \
  --dashboard-host=0.0.0.0 \
  --num-gpus=$NUM_GPUS \
  --temp-dir=$RAY_TEMP_DIR

sleep 5
ray status

Issue 1: Job hangs waiting for placement group
When launching grpo_fast.py with the following configuration: 
--num_learners_per_node 6
--vllm_num_engines 2
--vllm_tensor_parallel_size 1
The logs report a placement bundle of:
[{'GPU': 6, 'CPU': 60}]
and then hang indefinitely with:
Waiting for placement group to be scheduled
This happens despite the node having 8 GPUs available.
Issue 2:  Worker crash with SYSTEM_ERROR

With a smaller configuration (e.g., num_learners_per_node=2, vllm_num_engines=1), the job progresses further but eventually crashes with:
A worker died or was killed while executing a task by an unexpected system error.
Worker exit type: SYSTEM_ERROR
Worker exit detail: Worker unexpectedly exits with a connection error code 2 (EOF).
Question:

a)Is there a recommended configuration for single-node RLVR runs (especially for num_learners_per_node, vllm_num_engines, and placement group sizing)?

b) Is the grpo_fast_8b_single_node.sh script expected to work as-is on Slurm, or are additional Ray/Slurm-specific adjustments required?
 Thanks!

LalchandPandia:
@hamishivi

[...]

PR #1586, opened 2026-04-03
grpo_fast: harden single-node startup resource checks and diagnostics

taivu1998:
## Summary
- add a small startup resource-planning module for `grpo_fast`
- preflight-check Ray-visible CPU/GPU resources before creating the learner placement group
- bound learner placement-group waits and raise actionable diagnostics instead of hanging indefinitely
- add focused unit tests for the single-node startup resource calculations

## Why
Issue #1431 reports one-node GRPO/RLVR runs hanging at `Waiting for placement group`. In the current startup path, `grpo_fast.py` asks Ray for the learner placement group before surfacing the effective startup resource requirements, so CPU/GPU visibility problems can look like opaque hangs.

This change keeps the existing scheduling model intact, but makes startup behavior much easier to diagnose and much safer on single-node runs.

## Root Cause
- the learner placement group is requested before learner actors, data prep, and vLLM engines are initialized
- the practical one-node topology is more than just `8 GPUs`; it also depends on Ray-visible CPU for the learner placement group and follow-on actors
- `grpo_fast.py` previously waited on `pg.ready()` without a bounded timeout or enough context to explain why scheduling was stuck

## What Changed
- added `open_instruct/grpo_fast_resource_plan.py` to centralize startup resource planning and shortfall reporting
- added a GRPO startup preflight in `open_instruct/grpo_fast.py` that waits for Ray to report the minimum required resources before creating the learner placement group
- replaced the unbounded learner placement-group wait with a bounded wait that includes cluster resources, available resources, runtime context, and likely-cause hints in the failure message
- extracted named CPU constants for learner actors and the data-prep actor instead of leaving those values as inline magic numbers
- added `open_instruct/test_grpo_fast_resource_plan.py` to cover the `6 learners / 2 vLLM` topology, `single_gpu_mode`, learner-PG CPU shortfalls, and full-topology CPU shortfalls

## Validation
- `uv run pytest open_instruct/test_grpo_fast_resource_plan.py -q`
- `make style`
- `make quality`

## Notes
- this is intentionally a focused startup-hardening change; it does not redesign the single-node scheduler or placement strategy yet
- I could not run a full end-to-end local `grpo_fast.py` + `vllm` startup on macOS in this environment, so the runtime validation here is targeted unit coverage plus repo checks
- `make quality` still reports one unrelated pre-existing `ty` warning in `open_instruct/environments/tools/parsers.py`

## Issue
- Addresses #1431

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -176,14 +176,14 @@ def _build_vlm_name_mapper(model_name: str):
 
 def _startup_debug_context(
     args: grpo_utils.GRPOExperimentConfig,
-    plan: grpo_fast_resource_plan.GrpoFastResourcePlan,
+    requirements: dict[str, Any],
     cluster_resources: dict[str, float] | None = None,
     available_resources: dict[str, float] | None = None,
 ) -> str:
     ray_address = os.environ.get("RAY_ADDRESS", "<unset>")
     return "\n".join(
         [
-            grpo_fast_resource_plan.format_grpo_fast_resource_plan(plan),
+            grpo_fast_resource_plan.format_grpo_fast_startup_requirements(requirements),
             f"Ray cluster resources: {grpo_fast_resource_plan.format_resource_snapshot(cluster_resources)}",
             f"Ray available resources: {grpo_fast_resource_plan.format_resource_snapshot(available_resources)}",
             (
@@ -196,7 +196,7 @@ def _startup_debug_context(
 
 def wait_for_grpo_fast_minimum_cluster_resources(
     args: grpo_utils.GRPOExperimentConfig,
-    plan: grpo_fast_resource_plan.GrpoFastResourcePlan,
+    requirements: dict[str, Any],
     timeout_s: float = CLUSTER_STARTUP_TIMEOUT_S,
     poll_interval_s: float = 5.0,
 ) -> dict[str, float]:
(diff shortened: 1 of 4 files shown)


taivu1998:
Hi @hamishivi, could you help review this PR? Thanks!

open_instruct/grpo_fast_resource_plan.py
```diff
@@ -0,0 +1,136 @@
+import dataclasses
```
hamishivi: I like the other change (waiting for resources, timing out), but I feel like this extra plan stuff is a bit more than we need. Rather, it would just fine to have a util function that logs the current resources we have + what we are waiting for, as opposed to adding a whole dataclass with extra logic?

open_instruct/grpo_fast.py
```diff
@@ -1191,10 +1270,12 @@ def create_model_and_optimizer(
...
 ]:
     """Create the model, optimizer, and vLLM engines."""
-    # Create placement group
-    bundles = [{"GPU": actor_num_gpus, "CPU": actor_num_gpus * 10} for actor_num_gpus in args.num_learners_per_node]
-    pg = placement_group(bundles, strategy="STRICT_SPREAD")
-    ray_get_with_progress([pg.ready()], desc="Waiting for placement group")
+    resource_plan = grpo_fast_resource_plan.build_grpo_fast_resource_plan(args, vllm_config)
+    wait_for_grpo_fast_minimum_cluster_resources(args, resource_plan)
+
+    # Create the learner placement group after Ray can already see enough resources for the full topology.
```
hamishivi: unecessary comment

hamishivi (changes requested):

taivu1998:
@hamishivi, Thanks for your comments. Could you help review this PR again? Thanks!

hamishivi (approved):

Issue #1214, opened 2025-11-19
LLM judge will start crashing under heavy load


finbarrtimbers:
From @jacob-morrison: 

> Crashing, it's the same issue we saw yesterday; after 20-25 steps we'd start getting a ton of those supposedly spurious dequeue errors: https://beaker.allen.ai/orgs/ai2/workspaces/olmo-instruct/work/01KAD98HKMW7TBMCTXYT378RAT/logs?jobId=01KAD98HQR2KTG7VXNPA5NHQSE
2025-11-19T06:58:45.001Z connections: ['deque([(<aiohttp.client_proto.ResponseHandler object at 0x7d806124f590>, 1027562.866828507), (<aiohttp.client_proto.ResponseHandler object at 0x7d806124f7d0>, 1027563.013807494), (<aiohttp.client_proto.ResponseHandler object at 0x7d806124fa10>, 1027563.015665895), (<aiohttp.client_proto.ResponseHandler object at 0x7d806124fb30>, 1027563.087693763), (<aiohttp.client_proto.ResponseHandler object at 0x7d80633c42f0>, 1027563.134183762), (<aiohttp.client_proto.ResponseHandler object at 0x7d806124f6b0>, 1027563.16644183), (<aiohttp.client_proto.ResponseHandler object at 0x7d80633c40b0>, 1027563.184146161), (<aiohttp.client_proto.ResponseHandler object at 0x7d806124f8f0>, 1027563.185138809)])']and then it'd just endlessly time out
2025-11-19T06:58:45.001Z connector: <aiohttp.connector.TCPConnector object at 0x7d790071a0f0>
2025-11-19T06:58:45.001Z 2025-11-19 06:58:45 - ERROR - base_events.py:1833 - Unclosed client session

BrownianNotion:
Hey @finbarrtimbers, is this issue about DPO/PPO? I've been experiencing some crashes in `grpo_fast.py` just wanted to confirm this wasn't a related issue (to my knowledge there's no LLM judge here).

finbarrtimbers:
Yes, GRPO. We're not sure exactly what's going on and need to track this down to reproduce it. 

Just so you know, PPO is a bit unmaintained and needs some love. We plan to rewrite it in Jan/Feb.

finbarrtimbers:
Share the crash errors you've seen? GRPO should be robust. It's heavily used.

BrownianNotion:
Sure, no rush though as it's Christmas.

[...]

PR #1587, opened 2026-04-03
Stabilize GRPO LLM judge calls by routing them through the guarded LiteLLM helper

taivu1998:
## Summary
- route `LMJudgeVerifier` through the shared semaphore-guarded LiteLLM async path instead of calling `litellm.acompletion(...)` directly
- add a raw-response LiteLLM helper so the judge path can preserve exception-driven retries and usage-based cost accounting
- remove the stale per-verifier client-cache cleanup path and replace it with an honest no-op
- add focused tests for raw helper behavior, semaphore limiting, wrapper defaults, and judge retry/cost preservation

## Why
Issue #1214 reports that heavy-load GRPO runs with an LLM judge degrade after roughly 20-25 steps into timeouts, `Unclosed client session` warnings, and `aiohttp` connector teardown failures. The strongest code-level mismatch was that the judge verifier bypassed the repo's existing guarded LiteLLM helper, while the rubric path already used a shared per-event-loop semaphore.

This change keeps the fix scoped to the most likely root cause:
- reuse the existing guarded LiteLLM path
- preserve judge-specific retry behavior instead of inheriting rubric defaults
- avoid adding new config or scheduling complexity until the minimal fix is validated

## What changed
- Added `run_litellm_async_raw(...)` in `open_instruct/rubrics/run_utils.py`
- Kept `run_litellm_async(...)` behavior compatible for existing rubric callers
- Migrated `LMJudgeVerifier` in `open_instruct/ground_truth_utils.py` to the shared raw helper
- Simplified stale judge cleanup code that no longer reflected real resource ownership
- Added targeted unit coverage in `open_instruct/rubrics/test_rubrics.py` and `open_instruct/test_ground_truth_utils.py`

## Validation
- `ruff check` on touched files
- `python -m compileall` on the worktree
- `pytest open_instruct/test_ground_truth_utils.py open_instruct/rubrics/test_rubrics.py -q`
  - result: `86 passed, 6 skipped`

## Notes
- This PR addresses issue #1214.
- GRPO collection tests were not runnable in this macOS environment because `vllm` is not installed locally.
- The local pytest run also required `DS_ACCELERATOR=cpu` to avoid a DeepSpeed macOS import crash.

GPU_TESTS=bypass

### DIFF
--- a/open_instruct/ground_truth_utils.py
+++ b/open_instruct/ground_truth_utils.py
@@ -22,7 +22,6 @@ from typing import Any, Literal
 
 import numpy as np
 import requests
-from litellm import acompletion
 
 from open_instruct import context_window_checker, logger_utils
 from open_instruct.if_functions import IF_FUNCTIONS_MAP
@@ -37,7 +36,7 @@ from open_instruct.math_utils import (
     remove_boxed,
 )
 from open_instruct.rubrics.prompts import RUBRIC_SCORING_PROMPT
-from open_instruct.rubrics.run_utils import extract_json_from_response, run_litellm_async
+from open_instruct.rubrics.run_utils import extract_json_from_response, run_litellm_async, run_litellm_async_raw
 from open_instruct.utils import extract_final_answer
 
 logger = logger_utils.setup_logger(__name__)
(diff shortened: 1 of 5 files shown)


taivu1998:
Hi @finbarrtimbers, could you help review this PR? Thank you!

taivu1998:
Hi @hamishivi, could you help review this PR? Thanks!

open_instruct/ground_truth_utils.py
```diff
@@ -856,17 +851,9 @@ def __call__(
     @classmethod
```
finbarrtimbers, line 851: Can we just delete the method?

finbarrtimbers (approved):
I think this is fine, but I want @hamishivi to verify as he wrote this code.

open_instruct/ground_truth_utils.py
hamishivi: yeah, if this is doing nothing we should delete.

hamishivi (approved):
Generally looks good with one fix.

PR #1589, opened 2026-04-06
Fix incorrect hardcoded checkpoint state path for multi-GPU DeepSpeed

### LINKED ISSUE #1262: Incorrect re-loading of the dataloader state
in `open_instruct/grpo_fast.py` we reload the data loader state with 
 
`checkpoint_path = os.path.join(args.checkpoint_state_dir, "global_0", "state.pt")` and then find it there but this is wrong if we have save with deepspeed on multiple GPUs in which case this isn't the right file

Chessing234:
## Summary
- Fixes the checkpoint reload in `main()` which used a hardcoded path `global_0/state.pt` that does not match DeepSpeed's actual checkpoint directory structure (`global_step{N}/mp_rank_00_model_states.pt`)
- Reads the `latest` file (maintained by DeepSpeed and `calibrate_checkpoint_state_dir`) to find the correct checkpoint tag, then loads the rank-0 model states file
- Includes a fallback glob search for robustness when the `latest` file is missing

Fixes #1262

## Test plan
- [ ] Verify checkpoint reload works correctly on single-GPU training runs
- [ ] Verify checkpoint reload works correctly on multi-GPU DeepSpeed training runs
- [ ] Verify the fallback glob path works when the `latest` file is absent

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file.
 - Add deprecation warning to `finetune.py` pointing users to the OLMo-core SFT implementation (https://github.com/allenai/open-instruct/pull/1574).
 
 ### Fixed
+- Fix incorrect hardcoded checkpoint state path for multi-GPU DeepSpeed resumption (https://github.com/allenai/open-instruct/pull/1589).
 - Fix shellcheck `$@` quoting in GRPO debug scripts (https://github.com/allenai/open-instruct/pull/1572).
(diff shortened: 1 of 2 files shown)


hamishivi (approved):

hamishivi:
@Chessing234 if you fix up the CI checks happy to merge :)

Chessing234:
Thanks! The CI failure was just due to a missing CHANGELOG entry, so I went ahead and added that to fix the checks. It looks like the PR has successfully been merged since then, but thanks again for your help and the review!

Chessing234:
Looks like all the CI checks are passing now! Let me know if there's anything else needed.

PR #1592, opened 2026-04-08
Fix Batch.__getitem__ active_tools for int and list indexing

Chessing234:
## Summary
- Fix `Batch.__getitem__` handling of `active_tools` for `int` and `list[int]` indexing in `model_utils.py`
- **Int indexing** (`batch[5]`): `active_tools` was not wrapped in a list, unlike all other fields — returned a bare element instead of a single-element list
- **List indexing** (`batch[[1,3,5]]`): crashed with `TypeError` because Python lists don't support list-based indexing (`my_list[[1,3,5]]`)
- Root cause: `active_tools` was computed once before the `isinstance` type dispatch (line 138), but each branch needs different handling (slice, wrap-in-list, or list comprehension)
- Fix: moved `active_tools` computation inline into each branch, matching the pattern used by all other optional fields (`raw_queries`, `decoded_responses`, `indices`, `scores`)

## Test plan
- [x] `uv run ruff check open_instruct/model_utils.py` — all checks passed
- [x] `uv run pytest open_instruct/test_model_utils.py` — 7/7 tests passed

GPU_TESTS=bypass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ All notable changes to this project will be documented in this file.
 - Add deprecation warning to `finetune.py` pointing users to the OLMo-core SFT implementation (https://github.com/allenai/open-instruct/pull/1574).
 
 ### Fixed
-- Fix `Batch.__getitem__` handling of `active_tools` for int and list indexing (https://github.com/allenai/open-instruct/pull/NEW).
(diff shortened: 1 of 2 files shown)


Chessing234:
Good catch — the CHANGELOG placeholder has been updated with the correct PR number (#1592). Should be good now!

hamishivi (approved):

PR #1597, opened 2026-04-10
Fix PreferenceDatasetProcessor.filter not checking rejected sequence length

Chessing234:
## Bug

`PreferenceDatasetProcessor.filter` uses a nested ternary to check sequence lengths:

```python
len(row[INPUT_IDS_CHOSEN_KEY]) <= self.config.max_token_length
if self.config.max_token_length is not None
else (
    len(row[INPUT_IDS_REJECTED_KEY]) <= self.config.max_token_length
    if self.config.max_token_length is not None  # ← duplicate condition
    else True
)
```

The inner `if self.config.max_token_length is not None` re-tests the exact condition already evaluated as `False` by the outer `else` branch. This makes the rejected-length check dead code — rejected sequences exceeding `max_token_length` are never filtered out.

## Fix

Check both chosen **and** rejected lengths in a single branch when `max_token_length` is set, matching the pattern used by `SFTDatasetProcessor.filter` (line 301-314).

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ All notable changes to this project will be documented in this file.
 - Add deprecation warning to `finetune.py` pointing users to the OLMo-core SFT implementation (https://github.com/allenai/open-instruct/pull/1574).
 
 ### Fixed
+- Fix `PreferenceDatasetProcessor.filter` dropping the rejected-sequence length check, so over-long rejected completions were no longer filtered (https://github.com/allenai/open-instruct/pull/1597).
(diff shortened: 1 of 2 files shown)


hamishivi (approved):

hamishivi:
@Chessing234 happy to merge if you add a changelog entry (fix CI)!

Chessing234:

PR #1598, opened 2026-04-10
Miscellaneous small fixes for training scripts

BrownianNotion:
Small self-explanatory fixes to the training scripts `finetune.py (deprecated), dpo_tune_cache.py, grpo_fast.py` (see changelog/commits for descriptions).

Regarding the optimizer param group fix in 9ddc55da6464ca7a2d0a10a6a11967a53796a4ac, the error we get is
```
[rank0]:   File "/workspace/open-instruct/.venv/lib/python3.12/site-packages/torch/optim/lr_scheduler.py", line 296, in _update_lr                                                                                                                   
[rank0]:     for param_group, lr in zip(self.optimizer.param_groups, values, strict=True): 
 ValueError: zip() argument 2 is longer than argument 1 
```
which seems to be due the addition of `strict=True` in the scheduler for torch 2.10. eg. https://github.com/huggingface/transformers/pull/43711

### DIFF
--- a/open_instruct/dpo_tune_cache.py
+++ b/open_instruct/dpo_tune_cache.py
@@ -39,7 +39,6 @@ import torch.utils.data
 import transformers
 from accelerate import Accelerator, DataLoaderConfiguration
 from accelerate.accelerator import GradientAccumulationPlugin
-from accelerate.logging import get_logger
 from accelerate.utils import DeepSpeedPlugin, InitProcessGroupKwargs, set_seed
 from huggingface_hub import HfApi
 from peft import LoraConfig, TaskType, get_peft_model, prepare_model_for_kbit_training
--- a/open_instruct/dpo_tune_cache.py
+++ b/open_instruct/dpo_tune_cache.py
@@ -39,6 +39,7 @@ import torch.utils.data
 import transformers
 from accelerate import Accelerator, DataLoaderConfiguration
 from accelerate.accelerator import GradientAccumulationPlugin
+from accelerate.logging import get_logger
 from accelerate.utils import DeepSpeedPlugin, InitProcessGroupKwargs, set_seed
 from huggingface_hub import HfApi
 from peft import LoraConfig, TaskType, get_peft_model, prepare_model_for_kbit_training
@@ -69,7 +70,7 @@ from open_instruct.utils import (
     maybe_use_ai2_wandb_entity,
 )
 
-logger = logger_utils.setup_logger(__name__)
+logger = get_logger(__name__)
 
 
 def build_deepspeed_config(
--- a/open_instruct/dpo_tune_cache.py
+++ b/open_instruct/dpo_tune_cache.py
@@ -423,6 +423,10 @@ def main(args: dpo_utils.DPOExperimentConfig, tc: TokenizerConfig):
         },
         {"params": [p for n, p in model.named_parameters() if any(nd in n for nd in no_decay)], "weight_decay": 0.0},
     ]
+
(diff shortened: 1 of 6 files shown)


open_instruct/dpo_tune_cache.py
```diff
@@ -69,7 +71,7 @@
     maybe_use_ai2_wandb_entity,
 )
 
-logger = logger_utils.setup_logger(__name__)
```
hamishivi, line 72: we want to use the open-instruct logger, not accelerate's logger here.

open_instruct/grpo_fast.py
```diff
@@ -252,6 +252,7 @@ def load(self, path: str, map_location=None):
             adam_offload=args.deepspeed_offload_optimizer,
             stage=args.deepspeed_stage,
             bf16=True,
+            max_norm=args.max_grad_norm if args.max_grad_norm is not None else 0.0,
```
hamishivi, line 255: else should be 1.0. Actually, we already set the default to 1.0 in the code in `main`, so we don't need this.

hamishivi (changes requested):
Changes look good, but just two nits!

open_instruct/dpo_tune_cache.py
BrownianNotion: Sure, I can change back to `logger_utils.setup_logger` + guard the step/loss logging with `is_main_process`

open_instruct/grpo_fast.py
BrownianNotion: Just double checking we don't ever want to turn off gradient clipping? I saw that the default is 1, but noticed that None value indicates no gradient clipping for DPO, so I thought I'd copy the behaviour from there.

hamishivi: ohh... actually looking at this more deeply, max_grad_norm is not implemented for grpo_fast.py here. We have it as an arg for the olmo-core GRPO implementation, but not the deepspeed one, so really we should just not add this.

BrownianNotion: Sorry, not sure what you mean here, would you be able to clarify why we shouldn't pass the max_grad_norm value to deepspeed? 

Currently, the value is set but isn't passed, so it does nothing - I thought this was a bug.

Is the concern here that, if we set the value, it is misleading for the case where we use `grpo_fast.py` without deepspeed, since it wouldn't do anything?

hamishivi: hmm... I just double checked if it works despite accelerate also having its own grad clipping, but it seems internally it does indeed [just rely on deepspeed](https://github.com/huggingface/accelerate/blob/v1.13.0/src/accelerate/accelerator.py#L2970-L2974).

Oh, and seems that I thiiiink setting manually to none should work... sorry for the noise, actually I think this is fine.

hamishivi (approved):

hamishivi:
@BrownianNotion approved! Can you add a changelog entry and I'm happy to merge (assuming rest of CI passes okay).

open_instruct/grpo_fast.py
BrownianNotion: My impression was that since we are using Ray/torch.distributed in `grpo_fast.py`, the code isn't going through accelerate and we don't run into this conflict on the grad norm parameter. Let me know if I've misunderstood though, good to double check!

BrownianNotion:
> @BrownianNotion approved! Can you add a changelog entry and I'm happy to merge (assuming rest of CI passes okay).

Sorry, had the changelog entry but forgot to add the PR url 😅 fixed now

PR #1603, opened 2026-04-11
Fix: deterministic downsampling

mnoukhov:
Currently for `dataset_mixer_list dataset_name number_of_samples` if the `number_of_samples` is less than the size of the dataset, we randomly downsample. This means repeated runs aren't deterministic if you're doing debugging on a small subset of the data because you're sampling different data each time.

Switch to deterministic downsampling but keep random upsampling as we likely never do debugging with upsampling


open_instruct/dataset_transformation.py
```diff
@@ -1667,9 +1667,12 @@ def update_range(self, dataset_range: int):
         self.is_upsampled = dataset_range > original_size
 
     def select_samples(self, target_size: int):
-        """Upsample dataset to target_size by repeating samples."""
+        """Select samples deterministically for downsampling and repeat with random indices for upsampling."""
         original_size = len(self.dataset)
 
+        if target_size <= original_size:
+            return self.dataset.select(range(target_size))
```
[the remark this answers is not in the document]
mnoukhov: done

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 - Fix shellcheck `$@` quoting in GRPO debug scripts (https://github.com/allenai/open-instruct/pull/1572).
 - Add `--no_auto_dataset_cache` to GRPO and SFT integration test scripts to avoid HuggingFace 504 timeouts on CI runner (https://github.com/allenai/open-instruct/pull/1571).
+- Deterministic downsampling of dataset to enable reproducibility (https://github.com/allenai/open-instruct/pull/1603)
```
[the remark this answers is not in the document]
mnoukhov: done

finbarrtimbers (changes requested):
Please add a test, but otherwise, LGTM.

open_instruct/dataset_transformation.py
```diff
@@ -1651,34 +1651,40 @@ def update_range(self, dataset_range: int):
...
-        # Calculate how many full repeats and how many extra samples
-        full_repeats = target_size // original_size
-        extra_samples = target_size % original_size
+        if target_size == original_size:
+            return self.dataset
 
-        # Create indices for upsampling
-        indices = []
+        elif target_size < original_size:
+            return self.dataset.select(range(target_size))
```
hamishivi, line 1661: Rather than this, can we just set a set a seed that doesn't change and shuffle, then pick?

We can run into issues where the dataset as uploaded is not shuffled, and so picking the first n samples ends up not really being a random sample. We've run into issues in the past where someone put in e.g. a 50% downsample and just missed half the sources in the dataset.

hamishivi (changes requested):

open_instruct/dataset_transformation.py
mnoukhov: Hmm, I generally specifically want it not shuffled but can add this as an arg

hamishivi: yea happy to have a no_shuffle arg or something similar!

PR #1604, opened 2026-04-12
Fix find_shared_text off-by-one and crash on empty input

Chessing234:
## Summary

`find_shared_text` in `reward_modeling_eval.py` has two bugs caused by the `for`/`break` pattern:

1. **Off-by-one when one string is a prefix of the other**: When the loop completes without hitting `break`, `i` is left at `min(len(a), len(b)) - 1`, so `chosen_text[:i]` drops the last shared character.

   ```python
   >>> find_shared_text("hello world", "hello world extra")
   'hello worl'   # missing final 'd'
   ```

2. **`UnboundLocalError` when either input is empty**: `range(0)` never executes, so `i` is never assigned, and `chosen_text[:i]` crashes.

   ```python
   >>> find_shared_text("", "abc")
   UnboundLocalError: cannot access local variable 'i'
   ```

**Fix**: Use early `return` on mismatch inside the loop. Fall through to `return chosen_text[:min_len]` when one string is a prefix of the other (or either is empty).

## Test plan

- [x] Verified both failure modes in Python REPL
- [x] Verified fix returns correct results for: identical strings, prefix relationship, mismatch, empty inputs

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


hamishivi (approved):
Happy to merge once you add a changelog entry.

Chessing234:

Chessing234:
Ah, I missed the suggestion to use `zip` and `enumerate` and add a return type hint earlier! I'll open a quick follow-up PR to clean that up.

PR #1605, opened 2026-04-13
Fix is_equiv returning None instead of False on ValueError

Chessing234:
## Bug

`is_equiv()` in `open_instruct/math_utils.py` has a `-> bool` return type but implicitly returns `None` when `sympy.simplify()` raises `ValueError`. The `except ValueError` handler (line 198-199) logs a debug message but lacks a `return False` statement, causing execution to fall through and exit the function without returning.

## Root cause

Every other error path in `is_equiv` (`LaTeXParsingError`, `SympifyError`, `TypeError`, `TimeoutError`, generic `Exception`) explicitly returns `False`. The `ValueError` handler was the only one missing its `return False`.

## Fix

Add `return False` after the debug log in the `except ValueError` handler, consistent with all other error paths in the function.


hamishivi (approved):
Please add a changelog entry and we are good!

Chessing234:

PR #1608, opened 2026-04-13
Warn about checkpoint disk space only on the first checkpoint

mnoukhov:
## Summary
- warn about low disk space only on the first checkpoint attempt of a GRPO run
- keep  stateless and scope the suppression to the checkpoint loop
- keep unit coverage for the disk-space warning helper itself

GPU_TESTS=bypass


finbarrtimbers:
My original goal for this was to catch the case where the disk fills up on a long run.

mnoukhov:
I've been getting 100+ messages a day. I can just turn it off in my runs, but it just gets annoying and there's nothing you can do to turn it off or stop it once the run is going (especially if you're not the one who's using up the memory)

PR #1609, opened 2026-04-14
Dr tulu replication

hamishivi:
Add script for DR Tulu training w/ Qwen 3.5 + evolving rubrics. Also fixes up somethings:
- allows setting different environment pool sizes for each tool
- adds vllm parser for qwen 3.5-style tool calling
- ups the acquire timeout, since sometimes rollouts are just slow
- extra robustness around the tool formatting.

testing now


open_instruct/ground_truth_utils.py
```diff
@@ -1125,6 +1125,13 @@ async def async_call(
             logger.warning("No rubrics found in ground truth")
             return VerificationResult(score=0.0)
 
+        # Extract content from <answer> tags if present, otherwise use full response.
+        # Use the last match in case the model outputs multiple answer blocks.
+        response_for_scoring = prediction
+        answer_matches = re.findall(r"<answer>(.*?)</answer>", prediction, re.DOTALL)
+        if answer_matches:
```
finbarrtimbers: what about: 

```
if answer_matches := re.findall(r"<answer>(.*?)</answer>", prediction, re.DOTALL):
    response_for_scoring = answer_matches[-1].strip()
else:
    response_for_scoring = prediction
```

open_instruct/grpo_fast.py
```diff
@@ -459,7 +459,6 @@ def setup_model_update_group(self, vllm_engines):
         torch.distributed.barrier()
 
     def broadcast_to_vllm(self):
-        # avoid OOM
```
finbarrtimbers, line 462: add back plz

open_instruct/grpo_fast.py
```diff
@@ -1135,9 +1134,13 @@ def create_tool_pools(
             available_tools = ", ".join(TOOL_REGISTRY.keys())
             raise ValueError(f"Unknown tool: {parsed_tool.name}. Available tools: {available_tools}")
 
+        # Extract pool_size_override before passing config to the dataclass
```
finbarrtimbers: What happens if you don't extract `per_tool_pool_size`?

Also, what about adding `pool_size` method to the `tool_config_class` that has this logic? Can we then have `EnvironmentPool.__init__` take `tool_class_config` (or `tool_class`) as the only input to the c'tor?

open_instruct/vllm_utils.py
```diff
@@ -182,8 +182,17 @@ def process_tool_tokens(
...
+    tool_outputs = [str(o) if o is not None else "" for o in tool_outputs]
     formatted_output = tool_parser.format_tool_outputs(tool_outputs, role=role)
-    tokens = tokenizer.encode(formatted_output, add_special_tokens=False)
+    if not isinstance(formatted_output, str):
+        formatted_output = str(formatted_output)
+    # Strip null bytes and other characters that the fast tokenizer's Rust backend rejects.
+    formatted_output = formatted_output.replace("\x00", "")
+    try:
+        tokens = tokenizer.encode(formatted_output, add_special_tokens=False)
+    except TypeError:
```
finbarrtimbers, line 188: This feels overly defensive! I think it's fine to crash?

open_instruct/vllm_utils.py
```diff
@@ -182,8 +182,17 @@ def process_tool_tokens(
     Returns:
         Tuple of (tokens, logprobs, masks, excess).
     """
+    tool_outputs = [str(o) if o is not None else "" for o in tool_outputs]
```
finbarrtimbers: Should this logic live here or inside the tool_parser?

open_instruct/vllm_utils.py
```diff
@@ -182,8 +182,17 @@ def process_tool_tokens(
...
     formatted_output = tool_parser.format_tool_outputs(tool_outputs, role=role)
-    tokens = tokenizer.encode(formatted_output, add_special_tokens=False)
+    if not isinstance(formatted_output, str):
```
finbarrtimbers: Can we force `tool_parser` to handle this?

open_instruct/vllm_utils.py
```diff
@@ -751,7 +764,11 @@ def _should_stop(self) -> bool:
     def process_from_queue(self) -> None:
         finalize_futures: list[futures.Future] = []
         while True:
-            completion_future = accumulate_completions(self, self.completion_queue.get())
+            try:
```
finbarrtimbers: Why change this? I thought the hanging forever behaviour was desired?

finbarrtimbers (changes requested):

open_instruct/ground_truth_utils.py
hamishivi: ohhh very nice

open_instruct/vllm_utils.py
hamishivi: I had some problems where this crashed my job... Ideally I feel that we should be very defensive about model outputs/tool stuff, since we would rather pass errors to the model and allow it to recover rather than crash our own loop?

open_instruct/vllm_utils.py
hamishivi: oh, oop, I think this was from some tests to see if this was causing issues with weight sync (it was not).

open_instruct/environments/tools/parsers.py
```diff
@@ -39,14 +39,25 @@ def get_tool_calls(self, text: str) -> list[EnvCall]:
         """Extract tool calls from model outputs."""
         pass
 
-    @abstractmethod
-    def format_tool_outputs(self, tool_outputs: list[str], role: str = "tool") -> str:
+    def format_tool_outputs(self, tool_outputs: list[Any], role: str = "tool") -> str:
```
finbarrtimbers, line 42: list[str]?

open_instruct/grpo_fast.py
```diff
@@ -1155,8 +1155,10 @@ def create_tool_pools(
             configs_to_create.append((config, parsed_tool.call_name, tool_config_class.tool_class))
 
         for cfg, call_name, tool_class in configs_to_create:
+            effective_pool_size = cfg.pool_size if cfg.pool_size is not None else pool_size
             kwargs = asdict(cfg) | {"call_name": call_name}
```
finbarrtimbers: what about: 

```
kwargs = asdict(cfg) | {"call_name": call_name, "pool_size": cfg.pool_size or pool_size}
pools[call_name] = EnvironmentPool.remote(actor_class=tool_class, **kwargs)
```

finbarrtimbers (approved):

open_instruct/environments/tools/parsers.py
hamishivi: Nah, this joins together multiple tool outputs (if we had them)

PR #1611, opened 2026-04-14
Fix DataPreparationActor hanging on shutdown

hamishivi:
## Summary
- Added `request_shutdown()` method to `DataPreparationActor` and call it from `cleanup_training_resources()` before signaling other actors to stop.

Basically, when the trainer is really lagging behind, the data loader waits for it to progress, but if the job crashes or dies for some reason, the data loader doesn't receive the shutdown request because its stuck waiting for the trainer to progress a step before it starts checking the queue again.

## Test plan
- [x] Run `make style && make quality` (verified locally — passes)
- [x] Run single-GPU debug script: `./scripts/train/build_image_and_launch.sh scripts/train/debug/single_gpu_on_beaker.sh`
- [x] Verify the job exits cleanly after training completes (no "waiting for step to be consumed" spam in logs)


finbarrtimbers (changes requested):
What if we just call `ray.kill` in `cleanup_training_resources`? I don't think `DataPreparationActor` owns the vllm engines so it doesn't matter.

finbarrtimbers (approved):

PR #1612, opened 2026-04-15
Fix verify_sentence_constraint missing ! as a sentence terminator

Chessing234:
## Bug

\`verify_sentence_constraint\` in \`open_instruct/if_functions.py\` undercounts sentences in any response that ends one or more of its sentences with \`!\`.

For example:

\`\`\`python
>>> verify_sentence_constraint("Hello! How are you? I am fine.", 3, "around")
False   # actual sentence count is reported as 2
\`\`\`

## Root cause

The regex used to split \`text\` into sentences only recognises \`.\` and \`?\` as sentence terminators:

\`\`\`python
sentences = re.split(r"(?<!\w\.\w.)(?<![A-Z][a-z]\.)(?<=\.|\?)\s", text)
\`\`\`

The positive lookbehind \`(?<=\.|\?)\` does not include \`!\`, so \`"! "\` is never treated as a sentence boundary. Anything matching \`…! …\` stays glued to the next sentence in the split result, and \`len(sentences)\` is therefore too low whenever the response uses exclamations.

\`verify_sentence_constraint\` is wired into \`IF_FUNCTIONS_MAP\` and consumed by the \`IFEvalVerifier\` / \`IFEvalGroundTruthVerifier\` path in \`ground_truth_utils.py\`, so the undercount silently biases IFEval "around / at least / at most N sentences" rewards during RL / evaluation.

## Fix

Extend the lookbehind alternation to include \`!\`:

\`\`\`python
sentences = re.split(r"(?<!\w\.\w.)(?<![A-Z][a-z]\.)(?<=\.|\?|!)\s", text)
\`\`\`

The two negative lookbehinds guarding abbreviations (\`\w\.\w.\` and \`[A-Z][a-z]\.\`) are unchanged, so abbreviations such as \`U.S.\`, \`Mr.\`, \`Dr.\` continue to not trigger a split. Verified:

\`\`\`python
"Hello! How are you? I am fine."
  before: ['Hello! How are you?', 'I am fine.']             # 2
  after:  ['Hello!', 'How are you?', 'I am fine.']           # 3

"Go to the U.S. now! What do you think?"
  before: ['Go to the U.S. now! What do you think?']         # 1
  after:  ['Go to the U.S. now!', 'What do you think?']      # 2
\`\`\`

GPU_TESTS=bypass


hamishivi (approved):
Thanks!

PR #1614, opened 2026-04-15
Split `accumulate_inference_batches` into `process_single_result` and `combine_processed_results`.

finbarrtimbers:
This refactoring is done to make it easier to add a replay buffer, which will slot in between these two functions.

Runs:

1. Single GPU GRPO: [Beaker](https://beaker.org/ex/01KQJRBA2S3B8T6P2HBN4MB0WY)
2. Large multi-node GRPO: [Beaker](https://beaker.org/ex/01KQKFQ9XVBEKY70GEB3SYQY9W)


open_instruct/data_loader.py
```diff
@@ -737,40 +737,110 @@ def add_prompt_to_generator(
     )
 
 
-def accumulate_inference_batches(
-    inference_results_Q: ray_queue.Queue,
+@dataclass
+class Group:
```
hamishivi, line 741: ah this is nice to have!

hamishivi (approved):

PR #1616, opened 2026-04-15
Simplifies the model step logic

finbarrtimbers:
Simplifies model_step handling in `vllm_utils.py`: `set_model_step` is now awaited synchronously inside `_broadcast_weights_ipc` and the NCCL branches of `broadcast_weights_to_vllm`, and the `model_step` kwarg has been dropped from `update_weights` (the separate `set_model_step` RPC is the sole path). Also fixes two related gaps on the OLMo-core path:

- `open_instruct/olmo_core_train_modules.py`: `train_batch` now records `batch["metrics"]` via `self.record_metric`, so DataPreparationActor metrics (e.g. `num_steps_off_policy`, `model_step_mean`) surface in wandb for `grpo.py` runs.
- `open_instruct/grpo.py`: `beaker_config` is merged into `json_config` so the wandb run config includes `beaker_experiment_url` / `beaker_workload_id`, matching `grpo_fast.py`.

## Empirical findings on num_steps_off_policy

Observed on `grpo_fast.py` single-GPU GRPO debug run:

### `async_steps=8` (broken — [Beaker](https://beaker.org/ex/01KPNYEE4Z8D3P25EWW0ZGP69K), [wandb](https://wandb.ai/ai2-llm/open_instruct_internal/runs/xk63aipc))

| _step | model_step_mean | num_steps_off_policy |
| ----- | --------------- | -------------------- |
| 1     | 0               | 0                    |
| 2     | 0               | 1                    |
| 3     | 0               | 2                    |
| 4     | 0               | 3                    |
| 5     | 0               | 4                    |
| 6     | 0               | 5                    |

`model_step_mean` never advances; staleness grows linearly with `training_step`.

### `async_steps=1` (works — [Beaker](https://beaker.org/ex/01KPP2W8FEZNKV7Y1GTRTE78KT), [wandb](https://wandb.ai/ai2-llm/open_instruct_internal/runs/4fpzl2te))

| _step | model_step_mean | num_steps_off_policy |
| ----- | --------------- | -------------------- |
| 1     | 0.00            | 0.00                 |
| 2     | 0.00            | 1.00                 |
| 3     | 0.00            | 2.00                 |
| 4     | 1.00            | 2.00                 |
| 5     | 2.67            | 1.33                 |
| 6     | 3.00            | 2.00                 |

`model_step_mean` advances (0 → 1 → 2.67 → 3) and `num_steps_off_policy` stabilizes around 1–2.


mnoukhov (approved):
This is a much cleaner way of doing it, thanks.

Also I checked Gemini's review and I disagree, I think model_steps will never be None in that path so no need for checking if empty

finbarrtimbers:
> This is a much cleaner way of doing it, thanks.
> 
> Also I checked Gemini's review and I disagree, I think model_steps will never be None in that path so no need for checking if empty

Thank you, I agree with you against Gemini!

PR #1617, opened 2026-04-16
Remove attention_mask from forward_for_logprobs for HF

hamishivi:
## Summary
- Remove `attention_mask` parameter from `forward_for_logprobs` — HF constructs the correct 3D intra-document mask from `position_ids` internally, so we don't need to pass it explicitly.

For HF Flash-Attention, if you pass an attention mask it just assumes its a batched (not packed) input. Since for the HF side of GRPO fast, we pack, we shouldn't pass the attention mask so the HF correctly constructs the right args from position_ids (cu_seq_lens stuff). See https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_flash_attention_utils.py#L749

## Test plan
- [x] Unit tests pass (`uv run pytest`)
- [x] GPU tests pass


open_instruct/grpo_fast.py
```diff
@@ -595,7 +595,7 @@ def step(self):
                     local_logprobs_BT, entropy_BT = grpo_utils.forward_for_logprobs(
                         self.model,
                         data_BT.query_responses[i],
-                        data_BT.attention_masks[i],
+                        None,
```
finbarrtimbers, line 600: Add a comment please!

finbarrtimbers (approved):
Is there a test that would have caught this?

PR #1618, opened 2026-04-17
Fix gpt-4o output price missing one zero (10x undercount)

Chessing234:
## Bug

`PRICE_PER_TOKEN` in `open_instruct/judge_utils.py` has the gpt-4o output rate off by one zero:

```python
"gpt-4o":          {"input": 0.0000025, "output": 0.000001},   # 0.000001 = $1 / 1M
"gpt-4o-standard": {"input": 0.0000025, "output": 0.000001},
```

OpenAI's published gpt-4o rate is $2.50 / 1M input and $10.00 / 1M output. The input side matches (`0.0000025` = $2.5/1M), so the output was just missing a zero — `0.00001` instead of `0.000001`.

## Root cause

Every other model in the table has `output >= input` (2–4x higher); only `gpt-4o` / `gpt-4o-standard` has output **lower** than input, which is the tell:

```
gpt-4:              input 3.0e-5    output 6.0e-5   (2.0x)
gpt-3.5-turbo:      input 1.5e-6    output 2.0e-6   (1.3x)
gpt-4-1106-preview: input 1.0e-5    output 3.0e-5   (3.0x)
gpt-4o-mini:        input 1.5e-7    output 6.0e-7   (4.0x)
gpt-4.1:            input 2.0e-6    output 8.0e-6   (4.0x)
gpt-4o (current):   input 2.5e-6    output 1.0e-6   (0.4x)  ← wrong
```

## Why the fix is correct

`LLMJudge.get_cost` (`open_instruct/ground_truth_utils.py:747-752`) computes `input_rate * prompt_tokens + output_rate * completion_tokens` and strips the `-standard` suffix before lookup, so both `gpt-4o` and `gpt-4o-standard` callers route through this single entry. With the typo, reported gpt-4o judge spend is 10x below actual. Restoring the missing zero (`0.00001`) realigns both entries with OpenAI's stated $10/1M output rate, matching the input side and the pattern of every other row in the table.


hamishivi (approved):
Happy to merge once you add a changelog!

Chessing234:
Just added the changelog entry! Thanks for reviewing.

PR #1620, opened 2026-04-17
Match reference SFT run: olmo_core_finetune parity with pure olmo-core

finbarrtimbers:
## Summary

Validates that `open_instruct/olmo_core_finetune.py` trains identically to a pure olmo-core SFT loop, motivated by divergence observed against reference experiment [`01KNMEJKEZNJKZH9QWQW8CS0JW`](https://beaker.org/ex/01KNMEJKEZNJKZH9QWQW8CS0JW).

## Task

The goal is step-by-step parity between `open_instruct/olmo_core_finetune.py` and an equivalent pure olmo-core SFT loop, so that future divergences from reference runs can be attributed to specific layers of the stack rather than hunted across the whole pipeline. To make debugging cheap, the rig is single-GPU Qwen3-0.6B, 3 steps, identical seeds (init 33333, data loader 34521), identical hyperparams (lr 5e-6, warmup 0.03, AC `blocks.*.feed_forward`, SkipStepAdamW, flash_2), and identical dataset/weights.

Added:
- `scripts/train/debug/olmo_core_reference_sft.py` — minimal pure-olmo-core SFT runner (DDP, no CP, `TransformerConfig.qwen3_0_6B`, HF weight load via `load_hf_model`).
- `scripts/train/debug/oc_sft_qwen3_reference.sh` / `oc_sft_qwen3_openinstruct.sh` — paired Beaker launchers.

Fix to `olmo_core_finetune.py`:
- Removed a debug `next(iter(data_loader))` peek that consumed a batch without being rewound by the subsequent `data_loader.reshuffle(epoch=1)`, causing an off-by-one shift in training data vs a plain olmo-core loop.

## Parity results (single GPU, Qwen3-0.6B, 3 steps)

| Step | Pure olmo-core | `olmo_core_finetune.py` | Δ |
|---|---|---|---|
| 0 | 8.675 | 8.675 | **0.000** |
| 1 | 6.322 | 6.320 | 0.002 |
| 2 | 4.856 | 4.842 | 0.014 |

Step 0 is byte-identical, confirming identical weights and data-loader state. Residual ~0.3% drift at later steps is consistent with a single non-deterministic op per step (bf16 reduction order, torch.compile autotune kernel choice).

## Full 7B parity results (4 nodes × 8 H100, 10 steps)

Relaunched the full reference config (seq_len=32768, grad_accum=2, lr=8e-5, 2 epochs) against reference wandb run `l6dg9xwv` / [`01KNMEJKEZNJKZH9QWQW8CS0JW`](https://beaker.org/ex/01KNMEJKEZNJKZH9QWQW8CS0JW), testing both CP strategies:

| Step | Reference (`l6dg9xwv`, llama3) | Ours — llama3 | Ours — ulysses |
|---|---|---|---|
| 1 | 0.9756 | 0.9756 | 0.9756 |
| 2 | 1.0582 | 1.0580 | 1.058 |
| 3 | 0.9726 | 0.9726 | 0.9727 |
| 4 | 0.9667 | 0.9666 | 0.9667 |
| 5 | 1.0229 | 1.0220 | 1.022 |
| 6 | 0.9809 | 0.9806 | 0.9807 |
| 7 | 0.9237 | 0.9226 | 0.9226 |
| 8 | 0.8344 | 0.8339 | 0.8339 |
| 9 | 0.9188 | 0.9184 | 0.9185 |
| 10 | 0.8585 | 0.8585 | 0.8586 |

Both CP strategies match the reference within bf16 / `torch.compile` nondeterminism (max Δ 0.0011 for llama3, 0.0009 for ulysses).

### Throughput / MFU (running avg over >60 steps)

| metric | Llama3 + ring-flash-attn | **Ulysses** | Δ |
|---|---|---|---|
| MFU (actual avg) | 38.08% | **45.88%** | **+20.5%** |
| TPS / device | ~6,350 | ~7,400 | +16.5% |
| flops/s / device | ~377 TFLOPs | ~436 TFLOPs | +15.6% |

Ulysses is mathematically equivalent to llama3 under standard softmax attention (all-to-all head repartition vs. ring accumulation — same math, different communication pattern). On Jupiter (H100, DOCA OFED interconnect), the all-to-all happens to be cheaper than the ring accumulate.

**Default switched to `cp_strategy=ulysses`.** This also eliminates the ring-flash-attn dependency entirely, so `scripts/docker/patch_ring_flash_attn.py` and the corresponding Dockerfile step have been removed.

Runs:

1. Pure olmo-core reference (Qwen3-0.6B): [Beaker](https://beaker.org/ex/01KPE49WD35E5HH9V9GY1F4P5C)
2. `olmo_core_finetune.py` open-instruct (Qwen3-0.6B): [Beaker](https://beaker.org/ex/01KPE49WDADQZHHD74W8ZE7AAG)
3. Full 7B — llama3 CP strategy (ring-flash-attn): [Beaker](https://beaker.org/ex/01KPEHZKJQRJ5DD150XBYNWT59)
4. Full 7B — ulysses CP strategy: [Beaker](https://beaker.org/ex/01KPENNBD43D1PCH1HQ0Y666QP)

## Test plan
- [x] Relaunch the full 7B `01KNMEJKEZNJKZH9QWQW8CS0JW`-equivalent run with this branch and confirm the step-0/1 losses match the reference.
- [x] Verify ulysses CP strategy produces matching loss to the reference run (and faster MFU).
- [ ] `./scripts/train/build_image_and_launch.sh scripts/train/debug/oc_sft.sh` still succeeds.

GPU_TESTS=bypass


open_instruct/olmo_core_finetune.py
```diff
@@ -0,0 +1,315 @@
...
+logger = logger_utils.setup_logger(__name__)
+
+
+_DEFAULT_EPHEMERAL_SAVE_INTERVAL = 500
+
+_TOKENIZE_BARRIER_TIMEOUT_HOURS = 24
+
+
+def _numpy_dir_is_populated(numpy_dir: str) -> bool:
+    return os.path.exists(os.path.join(numpy_dir, "token_ids_part_0000.npy"))
```
hamishivi: nit: Can we avoided the hardcoded path name, or make it a constant somewhere?

open_instruct/olmo_core_finetune.py
```diff
@@ -0,0 +1,315 @@
...
+        pre_init_rank = int(os.environ.get("RANK", 0))
+        if pre_init_rank == 0:
+            if _numpy_dir_is_populated(numpy_dir):
+                logger.info(f"Numpy SFT files already present at {numpy_dir}; nothing to do.")
+            else:
+                _tokenize_to_numpy_dir(numpy_dir, args, tc, transform_fn_args, visualize=True)
+            logger.info("Dataset cached successfully. Exiting because --cache_dataset_only was set.")
+        return
+
+    if not _numpy_dir_is_populated(numpy_dir):
```
hamishivi, line 151: Is it possible to have tokenization run as part of the job if it is not detected, or not really?

open_instruct/olmo_core_finetune.py
```diff
@@ -0,0 +1,315 @@
...
+    )
+
+    if is_main_process:
+        os.makedirs(args.checkpoint.output_dir, exist_ok=True)
+    if is_distributed():
+        dist.barrier()
+
+    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+
+    init_device = "meta" if not use_hf_ckpt else "cpu"
```
hamishivi: Does olmo-core not support meta tensor type init? Wouldn't this potentially cause RAM issues when loading large (>100B) models?

open_instruct/olmo_core_utils.py
```diff
@@ -224,16 +340,32 @@ def get_transformer_config(model_name_or_config: str, vocab_size: int, attn_back
...
-def setup_model(
-    model_name_or_path: str, config_name: str | None, attn_implementation: AttentionBackendName
-) -> tuple[Transformer, TransformerConfig]:
-    hf_config = transformers.AutoConfig.from_pretrained(model_name_or_path, trust_remote_code=True)
-    vocab_size = hf_config.vocab_size
+def setup_model(model_config_args: ModelConfig, init_device: str = "cpu") -> tuple[Transformer, TransformerConfig]:
+    model_name_or_path = model_config_args.model_name_or_path
+    if is_hf_checkpoint(model_name_or_path):
+        hf_config = transformers.AutoConfig.from_pretrained(model_name_or_path, trust_remote_code=True)
+        vocab_size = hf_config.vocab_size
```
hamishivi, line 350: Can we add some logging here that its a hf checkpoint detected? I can envision it might be useful in cases where people are loading local models and might be debugging why something weird is happening (which may be loading-path-based).

hamishivi (changes requested):
Generally looks good, esp if testing, have some nits.

open_instruct/olmo_core_finetune.py
finbarrtimbers: Yeah, but it's a bit of a footgun though, as it can take, say, 30 minutes, and uses GPUs. I don't want to have, say, 8 nodes waiting while tokenization runs.

open_instruct/olmo_core_utils.py
finbarrtimbers: yeah good point, done.

open_instruct/olmo_core_finetune.py
finbarrtimbers: The bigger issue is that, inside `load_hf_model`, Olmo-core, `load_hf_model` [materializes the full HF model on rank 0](https://github.com/allenai/OLMo-core/blob/2e570869bb3cfa93c0d637c823f4387d9cb4d804/src/olmo_core/nn/hf/checkpoint.py#L103) (`AutoModelForCausalLM.from_pretrained` at
  `checkpoint.py:103,107`). But we can use that and then try to upstream this.

open_instruct/olmo_core_finetune.py
finbarrtimbers: Done!

open_instruct/olmo_core_finetune.py
hamishivi: Sure, sounds good.

open_instruct/olmo_core_finetune.py
hamishivi: Yeah agreed there. Sometimes it can be annoying if its debug jobs or small datasets though... I guess for now its fine.

hamishivi (approved):

PR #1621, opened 2026-04-18
Drop unused ty ignore in VllmToolParser._make_request

finbarrtimbers:
## Summary

- Removes a `# ty: ignore[invalid-argument-type]` on the `ChatCompletionRequest(...)` call in `VllmToolParser._make_request` that ty no longer flags.
- Split out of #1620 (match-sft) to keep that PR focused on the SFT parity work.

## Test plan

- [x] `uv run ty check open_instruct/environments/tools/parsers.py` — all checks pass without the ignore.

### DIFF
--- a/open_instruct/environments/tools/parsers.py
+++ b/open_instruct/environments/tools/parsers.py
@@ -154,7 +154,7 @@ class VllmToolParser(ToolParser):
 
         Usually these only need the list of tools.
         """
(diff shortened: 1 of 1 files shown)


finbarrtimbers:
Closing as no-op: CI ty has vllm installed and correctly flags the ChatCompletionRequest argument type, so the ignore is still needed. My local check passed only because vllm isn't installed on macOS, making ty treat the types as Unknown.

PR #1624, opened 2026-04-20
Fix IFEvalVerifier ZeroDivisionError when constraint has no instructions

Chessing234:
## Bug

`IFEvalVerifier.__call__` in `open_instruct/ground_truth_utils.py` builds a `rewards` list by zipping the constraint's `instruction_id` / `kwargs`, then returns the mean:

```python
rewards = []
if len(prediction) == 0 or len(answer) == 0:
    logger.warning("Empty prediction received for IFEvalVerifier.")
for instruction_key, args in zip(instruction_keys, args_list):
    ...
    rewards.append(1.0 if ... else 0.0)
return VerificationResult(score=sum(rewards) / len(rewards))
```

If `constraint_dict["instruction_id"]` is an empty list — possible for ground-truth entries that arrive with no IFEval constraints attached — the `zip` yields no iterations, `rewards` stays empty, and `sum(rewards) / len(rewards)` raises `ZeroDivisionError`, crashing the whole batch's reward computation inside the training loop.

## Root cause

The final return assumes at least one constraint was evaluated. The other early-exit in this method (empty prediction/answer) already returns `score=0.0`, but the empty-constraint case has no corresponding guard.

## Why the fix is correct

Return `VerificationResult(score=0.0)` when `rewards` is empty, matching the existing empty-prediction branch four lines above:

```python
if not rewards:
return VerificationResult(score=sum(rewards) / len(rewards))
```

No constraints means no evidence that any instruction was satisfied, so 0.0 is the consistent "all-or-nothing" score. Non-empty cases are unchanged. Two added lines, no other edits.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ All notable changes to this project will be documented in this file.
 - Add deprecation warning to `finetune.py` pointing users to the OLMo-core SFT implementation (https://github.com/allenai/open-instruct/pull/1574).
 
 ### Fixed
+- Fix `IFEvalVerifier` raising `ZeroDivisionError` when a ground-truth constraint has no `instruction_id` entries, by returning `score=0.0` for the empty-rewards case (https://github.com/allenai/open-instruct/pull/1624).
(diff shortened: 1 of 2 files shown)


hamishivi:
Closing since I don't think this is an actual bug: if there are no ifeval constraints, and the user has said to use the ifeval verifier, then they should see an error, and crash, since something is wrong. we have things like the `PassThroughVerifier` for cases like where the user might want to return 0 on an empty sample.

PR #1627, opened 2026-04-20
Fixes the weight sync

finbarrtimbers:
We had two issues: 

1. the lazy init for the weight sync was hardcoded to run after step 1, when actually, we want it to run after the first step, even if that was not 1 (e.g. after resuming from a checkpoint). 
2. Checkpoint resumption wasn't working correctly, as the actor would load the initial weights. 

Ran the multi-node GRPO script: [Beaker](https://beaker.org/orgs/ai2/workspaces/open-instruct-dev/work/01KPP1TSBT3DEWP28D27HDVMY4?taskId=01KPP1TSC11YF27RPKMMJWDS2P&jobId=01KPP1TSFHNB818EHPBQ1M1ZWJ).


open_instruct/grpo_fast.py
```diff
@@ -539,6 +539,17 @@ def setup_model_update_group(self, vllm_engines):
             ray_get_with_progress(refs, desc="Initializing vLLM weight transfer engines", timeout=600)
         torch.distributed.barrier()
 
+    def warmup_for_weight_sync(self):
```
hamishivi, line 544: would just wanna test this in a job if you havent!

hamishivi (approved):

open_instruct/grpo_fast.py
finbarrtimbers: I have! Here's an example: 

https://beaker.org/orgs/ai2/workspaces/open-instruct-dev/work/01KPP1TSBT3DEWP28D27HDVMY4?taskId=01KPP1TSC11YF27RPKMMJWDS2P&jobId=01KPP1TSFHNB818EHPBQ1M1ZWJ

PR #1631, opened 2026-04-21
Now, SFT tokenization streams directly to disk, making it ~9.4x faster. 

finbarrtimbers:
Changes SFT tokenization to stream tokens directly to disk instead of accumulative arrays in memory. This means we don't need explicit checkpointing anymore, and leads to a 10x speedup. 

| Run | Collect loop | Rate | Beaker |
|---|---|---|---|
| Pre-opt | 758.21s | 158.3 samples/s | [01KPPVTVJQVEBS6FF02WTAJD52](https://beaker.org/ex/01KPPVTVJQVEBS6FF02WTAJD52) |
| Post-opt | 80.65s | 1487.8 samples/s | [01KPPVT6AA44XRJ2MV6510S8K0](https://beaker.org/ex/01KPPVT6AA44XRJ2MV6510S8K0) |
| **Speedup** | **9.40×** | — | |

## Correctness 

Followed the procedure in [`docs/verify-tokenization.md`](docs/verify-tokenization.md): built an image from this branch, ran the full production mixer (`Dolci-Think-SFT-32B` + 5 tool datasets at 3.0×, `olmo123` chat template, `max_seq_length=32768`) on weka, then byte-compared against the existing `origin/main` golden at `/weka/oe-adapt-default/finbarrt/dataset/olmo-hybrid-main-repro`.

**Accuracy — byte-identical to baseline.** The compare job ([01KPY3NNQKF0FFJ8J8ECAX0JT7](https://beaker.org/ex/01KPY3NNQKF0FFJ8J8ECAX0JT7)) reports `=== PASSED ===`: all 99 `token_ids_part_*.npy`, 99 `labels_mask_part_*.npy`, 99 `token_ids_part_*.csv.gz`, and `dataset_statistics.json` (after stripping `timestamp`/`output_directory`) have matching sha256 hashes. It runs in 30m47s, vs the baseline, which took 2 days (although we also had other optimizations in #1633 which contribute to this speedup).

### DIFF
--- a/open_instruct/numpy_dataset_conversion.py
+++ b/open_instruct/numpy_dataset_conversion.py
@@ -259,10 +259,6 @@ def convert_hf_to_numpy_sft(
 
     logger.info("Collecting tokens from dataset...")
     total_samples = len(train_dataset)
-    loop_start_time = time.perf_counter()
-    utils.maybe_update_beaker_description(
-        current_step=start_idx, total_steps=total_samples, start_time=loop_start_time
-    )
 
     if start_idx >= total_samples:
         train_dataset_iter = train_dataset.select([])
@@ -337,7 +333,7 @@ def convert_hf_to_numpy_sft(
                     num_samples_skipped += 1
                     per_dataset_filtered[dataset_source] += 1
 
-                assert all(mask == 1 for mask in sample_attention), (
+                assert (sample_attention == 1).all(), (
                     f"Expected all attention mask values to be 1, but found: {sample_attention}"
                 )
 
--- a/open_instruct/numpy_dataset_conversion.py
+++ b/open_instruct/numpy_dataset_conversion.py
@@ -9,6 +9,7 @@ The output layout for each `output_dir` is:
 """
 
 import gzip
+import itertools
 import json
 import os
 import pathlib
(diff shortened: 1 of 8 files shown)


open_instruct/numpy_dataset_conversion.py
```diff
@@ -205,62 +35,172 @@ def _select_token_dtype(vocab_size: int):
...
                 json.dump(tokenizer_config, f, indent=2)
             logger.info(f"Added chat_template from {chat_template_path} to tokenizer_config.json")
 
 
+def _recover_partial_state(
+    tokens_path: pathlib.Path, labels_path: pathlib.Path, boundaries_path: pathlib.Path, token_item_size: int
+) -> tuple[int, int]:
+    # Boundaries file is the commit log: its last complete (start,end) pair determines how many
+    # bytes of tokens/labels are valid; trailing bytes from an incomplete write get truncated.
+    for path in (tokens_path, labels_path, boundaries_path):
```
farhatkevin: I think there may be a resume edge case here if partial state is ever incomplete: if one `_*.partial.bin` file is missing but the others still exist, `_recover_partial_state()` returns `(0, 0)` so the run restarts logically from sample `0`, but the remaining partial files are later reopened in append mode. Could that preserve stale token/label bytes from the previous interrupted run and mix them into the new output unless the remaining partial files are cleared first? Is that state considered possible / supported? If so, should we clear the remaining partial files or fail loudly instead of continuing?

open_instruct/numpy_dataset_conversion.py
```diff
@@ -205,62 +35,172 @@ def _select_token_dtype(vocab_size: int):
...
+            "per_dataset_filtered": per_dataset_filtered,
+            "total_trainable_tokens": 0,
+            "num_samples_skipped": 0,
+            "max_token_id": 0,
+        }
+
+    boundaries = np.memmap(boundaries_path, mode="r", dtype=_BOUNDARIES_DTYPE, shape=(num_samples, 2))
+    sample_lengths = (boundaries[:, 1] - boundaries[:, 0]).astype(np.int64)
+
+    labels_mm = np.memmap(labels_path, mode="r", dtype=np.uint8, shape=(total_tokens,))
```
farhatkevin, line 161: Would it be safer to use `np.add.reduceat(labels_mm, starts, dtype=np.int64)` here instead of `labels_mm.astype(np.int64)`? My understanding is that the current version materializes a full `int64` copy of the labels memmap before reducing, which could recreate a large memory spike during stats aggregation.

farhatkevin (commented):
I had a concern in the new streaming path: one possible resume-corruption edge case if partial files are missing inconsistently. Also one possible memory/scalability follow-up in stats aggregation from materializing the labels memmap as `int64`.

open_instruct/numpy_dataset_conversion.py
finbarrtimbers: ah, good point. that's much better. thank you!

open_instruct/numpy_dataset_conversion.py
finbarrtimbers: ah, good point. Fixed!

farhatkevin (approved):
changes lgtm!

PR #1634, opened 2026-04-22
Fix import error when installing open-instruct as a dependency

BrownianNotion:
Currently, when we install open_instruct as a third party dependency (eg. `uv pip install "git+https://github.com/allenai/open-instruct@main"`), the package fails to import and gives a module not found error. This is because 
```
[tool.setuptools]
py-modules = ["open_instruct"]
```
assumes a single file module and looks for `open_instruct.py` which doesn't exist and won't contain the entities we want to import.

I have changed this to look for everything under `open_instruct` except for the `test_data` subfolder - this should fix the import error. Let me know if any other modules/dirs should be excluded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ All notable changes to this project will be documented in this file.
 - Add deprecation warning to `finetune.py` pointing users to the OLMo-core SFT implementation (https://github.com/allenai/open-instruct/pull/1574).
 
 ### Fixed
+- Fix import error when installing as dependency (https://github.com/allenai/open-instruct/pull/1634)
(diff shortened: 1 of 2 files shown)


finbarrtimbers (approved):
Thank you!

BrownianNotion:
@finbarrtimbers it looks like this PR was accidentally reverted by the next merge commit https://github.com/allenai/open-instruct/commit/ef4ffe4e6c2bc1714bf5bb1817885e125b319a8e#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

finbarrtimbers:
I unreverted it in https://github.com/allenai/open-instruct/pull/1637.

PR #1636, opened 2026-04-24
Add dynamic versioning based on git tags

BrownianNotion:
Project version was being hardcoded to 0.1.0 (uv init default) instead of recognising release tags. This also means `uv pip show open-instruct` will unhelpfully show 0.1.0.

I have changed this to be autogenerated based on the release tag + any additional commits (eg. `0.2.1.dev106+g6644898d1.d20260424`) using [setuptools_scm](https://pypi.org/project/setuptools-scm/).

### DIFF
--- a/open_instruct/__init__.py
+++ b/open_instruct/__init__.py
@@ -1,0 +1,1 @@
+from open_instruct._version import __version__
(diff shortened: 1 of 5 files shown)


open_instruct/__init__.py
finbarrtimbers, line 1: Why do we need this?

finbarrtimbers (approved):

open_instruct/__init__.py
BrownianNotion: Not strictly needed, but allows us to do `open_instruct.__version__` inside python upon import.

PR #1637, opened 2026-04-24
Reapply packaging fix from #1634 reverted by #1622

finbarrtimbers:
## Summary
- #1622 accidentally reverted the `pyproject.toml` packaging fix from #1634, restoring the broken `py-modules = ["open_instruct"]` declaration that prevents installing `open_instruct` as a third-party dependency.
- Reapplies #1634's `[tool.setuptools.packages.find]` + `exclude-package-data` config.

## Test plan
- [ ] `uv pip install "git+https://github.com/allenai/open-instruct@reapply-1634-packaging-fix"` imports cleanly.

GPU_TESTS=bypass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 
 ### Changed
+- Reapply packaging fix so `open_instruct` installs correctly as a third-party dependency (reverted by #1622) (https://github.com/allenai/open-instruct/pull/1634).
 - Drop unused `data_types` import and inline `batch["batch"].to(device)` in `GRPOTrainModule` (https://github.com/allenai/open-instruct/pull/1635).
 - Use incremental binary checkpoint for SFT tokenization resume, eliminating O(N²) re-serialization (https://github.com/allenai/open-instruct/pull/1633).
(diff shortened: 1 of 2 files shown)


hamishivi (approved):

finbarrtimbers:
Hi @BrownianNotion, unfortunately this breaks our workflow so I am reverting this. 

After PRs #1634 and #1637, the editable install of open-instruct produces a finder that maps no modules
  when built with the current Dockerfile. Any script invoked as torchrun open_instruct/foo.py from /stage
  (where sys.path[0] becomes /stage/open_instruct, not /stage) fails with ModuleNotFoundError: No module
  named 'open_instruct'.

  The Dockerfile's uv run --frozen step (lines 70–73) runs before COPY open_instruct open_instruct (line
  79), with only pyproject.toml/uv.lock bind-mounted. With packages.find, setuptools enumerates package
  modules at install time — finds nothing because the source dir doesn't exist yet — and PEP 660 produces a
   finder with an empty module map. The subsequent COPY puts source on disk but doesn't refresh the finder.

  The old py-modules form produced a .pth file that simply prepended /stage to sys.path, so the
  later-COPY'd source was discoverable via normal directory traversal.

  Repro:
  git checkout main
  ./scripts/train/build_image_and_launch.sh scripts/train/debug/oc_sft.sh
  Image must be a fresh build (no cached layer where open_instruct/ was present at install time). Job dies
  during the first `from open_instruct import ….`

PR #1638, opened 2026-04-26
Add scripts/submit_eval_jobs_new.py

finbarrtimbers:
Adds `scripts/submit_eval_jobs_new.py`, which submits olmo-eval-internal eval jobs directly via `beaker experiment create`.

Supports Beaker dataset/HF repos/Weka paths and writes a v2 spec YAML to `configs/beaker_configs/auto_created/`.

GPU_TESTS=bypass


scripts/submit_eval_jobs.py
hamishivi, line 1: Can we just rename this to submit_eval_jobs_old.py and add a comment saying its deprecated? And then rename the new scrpt to submit_eval_jobs

scripts/submit_eval_jobs_new.py
```diff
@@ -0,0 +1,213 @@
...
+
+from open_instruct import launch_utils
+
+
+BEAKER_ID_RE = re.compile(r"^[0-9A-Z]{26}$")
+DEFAULT_CLUSTERS = ("ai2/jupiter",)
+MAX_EXPERIMENT_NAME_LEN = 128
+EXPERIMENT_NAME_SAFE_RE = re.compile(r"[^A-Za-z0-9_.-]+")
+
+INSTALL_SCRIPT = (
```
hamishivi: Is there not an image or something we can use with olmo-eval pre-installed? This setup is slightly weird if you want to use branches of olmo-eval since it hardcodes in cloning from main.

hamishivi (commented):
Looks good, some comments. I think we should explicitly deprecate the old eval job submission script.

scripts/submit_eval_jobs_new.py
finbarrtimbers: No, the default uses gantry and installs it live. I added a flag to install a branch (defaulting to main).

scripts/submit_eval_jobs.py
finbarrtimbers: Good point, done.

scripts/submit_eval_jobs_old.py
hamishivi, line 1: Should we also update the names in auto-eval-launch in finetune/grpo_fast/etc? or will they work out of the box with the new script (i dont mind if evals are missing)?

finbarrtimbers: I have updated them to work with the old version. I'm not 100% sure about compatibility.

hamishivi (approved):

PR #1646, opened 2026-04-29
fix(if_functions): fix validate_choice operand direction

Chessing234:
## Bug

`validate_choice` has the membership test operands reversed:

```python
# Current — wrong direction
return any(text in option for option in options)
```

This checks whether `text` (the full model response) is a substring of each `option`. For any response longer than the option string, this always returns `False`, so correct answers are misreported as failing.

## Root cause

The IFEval "choose" constraint means the response **contains** one of the allowed options (e.g. the response says "A" or "I choose B"). The correct test is `option in text`:

```python
# Fix — correct direction
return any(option in text for option in options)
```

This matches the semantics used in the [original IFEval evaluation code](https://github.com/google-research/google-research/tree/master/instruction_following_eval) and the `lm-evaluation-harness` implementation (`constrained_response in value` at `instructions.py:384`).

## Example

```python
text    = "I believe the answer is B"
options = ["A", "B", "C", "D"]

# Before (wrong): any("I believe the answer is B" in opt ...) → False
# After  (fix):   any(opt in "I believe the answer is B" ...)  → True  (B matches)
```

### DIFF
--- a/open_instruct/if_functions.py
+++ b/open_instruct/if_functions.py
@@ -360,7 +360,7 @@ def validate_title(text: str) -> bool:
 
 # Choose: From Answer with one of the following options: {options}
 def validate_choice(text: str, options: list) -> bool:
-    return any(text in option for option in options)
+    return any(option in text for option in options)
 
 
 # Minimum Number Highlighted Section: Highlight at least {N} sections in your answer with markdown, i.e. *highlighted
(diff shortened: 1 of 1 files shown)


finbarrtimbers (changes requested):
Can you add a test that would have caught this?

PR #1515, opened 2026-03-06
Migrate to vLLM native weight transfer API

finbarrtimbers:
vLLM 0.16.0 introduced native weight transfer APIs (`WeightTransferConfig`, `NCCLWeightTransferEngine`, `init_weight_transfer_engine`, `update_weights`). This PR replaces our existing weight syncing system, with manual per-parameter broadcast loops, with the built-in vLLM methods. Net deletion of ~250 lines.

Notable improvement in weight sync times; the max comes down significantly ([Wandb](https://wandb.ai/ai2-llm/open_instruct_internal/workspace?nw=o7utd9ubj6)):
Screenshot 2026-04-13 at 11 09 03 AM

Investigation doc is available as a [gist](https://gist.github.com/finbarrtimbers/b1e146fee4e3ed69a9cc952f16e6e07f), but is mostly a place for Claude/Codex to dump context. 

---

> [!NOTE]
> **Medium Risk**
> Touches distributed weight synchronization between the DeepSpeed trainer and vLLM engines; failures would surface as hangs, mismatched weights, or training instability. The change removes custom NCCL process-group/broadcast code in favor of vLLM’s built-in transfer engine, reducing custom surface area but relying on newer vLLM behavior.
> 
> **Overview**
> **Migrates GRPO training’s trainer→vLLM weight sync to vLLM 0.16.0’s native weight transfer path.** The trainer now initializes `NCCLWeightTransferEngine`/`WeightTransferConfig`, precomputes parameter metadata once, and sends weights via `trainer_send_weights` over a gathered-parameter iterator instead of per-parameter Ray RPC + `torch.distributed.broadcast` loops.
> 
> Engine-side APIs are updated to `init_weight_transfer_engine`/`update_weights`, the custom `WorkerWrap` mixin and copied `init_process_group` helper are removed, and `single_gpu_mode` is explicitly rejected for weight transfer (NCCL rank/device constraints). CLI keeps `vllm_sync_backend` as a deprecated no-op for backwards compatibility.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 64dd615e68d78a1475924e29ce614f9437fc7c17. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #1649, opened 2026-04-30
Get `grpo.py` working with native weight sync

finbarrtimbers:
For PR #1515, we moved to vLLM's new native weight sync API. It worked for `grpo_fast.py`, but `grpo.py` was untested. This PR gets `grpo.py` running with the new API and brings its metrics/behavior to parity with `grpo_fast.py`.

## Summary

**Weight sync (`grpo.py` / olmo-core actor)**
- Fire an initial vLLM weight sync before `trainer.fit()` (mirrors `grpo_fast`'s `initialize_weight_sync`) so the first NCCL broadcast runs from a known-good state.
- Fix FSDP2 illegal-memory-access: also unshard the root `FSDPModule` (root-level params like `model.norm` / `lm_head` were producing local-shard buffers with global stride). Require `gather_whole_model=True` for FSDP2 — per-block iteration deadlocks because reshard/unshard collectives interleave with trainer→vLLM NCCL sends.
- Drop redundant barriers and pre-broadcast CUDA calls in the olmo-core actor.
- Plumb `inflight_updates` through `VLLMWeightSyncCallback`.

**vLLM RPC**
- Unify `LLMRayActor.update_weights` signature: takes a single `update_info` dict + `model_step`, matching the IPC calling convention. NCCL callers now pack the same dict shape.

**Metrics parity with `grpo_fast`**
- New `StepTimingCallback` records `time/total`, `time/training`, `time/weight_sync*`, `learner_tokens_per_second_*`, and MFU/MBU utilization metrics.
- `GRPOTrainModule` now records `optim/grad_norm`, `training_step`, `episode`, `epoch`, token counters, and logs prompt/response-length histograms to wandb.
- Override `optim_step`/`zero_grads` as no-ops and route through `_do_optim_step`/`_do_zero_grads` to fix a double-optimizer-step bug.
- Plumb `compile_model` and `streaming_config` through `GRPOTrainModule`.

**Refactor**
- Extract `compute_vllm_local_debug_metrics` into `grpo_utils` and reuse from both `grpo_fast.py` and `olmo_core_train_modules.py`.

**Scripts**
- Add `scripts/train/qwen/qwen3_4b_dapo_math_oc.sh` (Qwen3-4B DAPO math on the olmo-core path).


open_instruct/vllm_utils.py
```diff
@@ -1413,40 +1393,58 @@ def broadcast_weights_to_vllm(
...
-
     trainer_args = NCCLTrainerSendWeightsArgs(group=model_update_group, packed=use_packed)
 
     if isinstance(model, FSDPModule):
         if not fsdp_submodules:
             raise ValueError("FSDP2 model has no FSDP submodules.")
-        for block_name, block in fsdp_submodules:
+        if not gather_whole_model:
+            raise ValueError(
+                "FSDP2 weight sync requires gather_whole_model=True. "
```
hamishivi, line 1420: Something to note for future is that this probably is a blocker for running >= 32B models, typically you OOM with `gather_whole_model=True` for them.

open_instruct/vllm_utils.py
```diff
@@ -1413,40 +1393,58 @@ def broadcast_weights_to_vllm(
     if is_rank_0:
         ray.get([engine.sleep.remote() for engine in vllm_engines])
 
-    if model_update_group is None:
+    if model_update_group is None and is_rank_0:
         return _broadcast_weights_ipc(model, vllm_engines, name_mapper, gather_whole_model, model_step)
```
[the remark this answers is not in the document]
hamishivi: +1, I think this causes hangs? when `gather_whole_model=False`

open_instruct/vllm_utils.py
```diff
@@ -780,14 +780,13 @@ def sleep(self) -> None:
     def wake_up(self) -> None:
         return self._run_async(self.llm_engine.wake_up(tags=["scheduling"]))
 
-    def update_weights(
-        self, names: list[str], dtype_names: list[str], shapes: list[tuple[int, ...]], packed: bool, model_step: int
-    ) -> None:
+    def update_weights(self, update_info: dict, model_step: int) -> None:
```
hamishivi: why move from specific items to a less-strictly typed dict? Could we enforce the dict shape better somehow?

open_instruct/utils.py
```diff
@@ -191,12 +191,12 @@ def repeat_each(seq, k):
...
 
 def ray_get_with_progress(
-    ray_refs: list[ray.ObjectRef], desc: str = "Processing", enable: bool = True, timeout: float | None = None
+    ray_refs: Iterable[ray.ObjectRef], desc: str = "Processing", enable: bool = True, timeout: float | None = None
 ):
     """Execute ray.get() with a progress bar using futures and collect timings.
 
     Args:
-        ray_refs: List of ray object references
+        ray_refs: Iterable of ray object references (materialized internally)
```
hamishivi: why this change?

hamishivi (changes requested):

open_instruct/vllm_utils.py
finbarrtimbers: Yeah, I think you're right. I'm kicking it down the road, as right now, nothing runs.

open_instruct/utils.py
finbarrtimbers: I removed the `list(ray_refs)` which wasn't needed. 

By accepting an iterable, we broaden the types we can take (e.g. we can take the `List[ray.ObjectRef]` we already accepted, but we can also accept other types). So it's strictly better.

open_instruct/vllm_utils.py
finbarrtimbers: vllm's `IPCWeightTransferEngine.trainer_send_weights` calls `update_weights` internally with a
  `dict` (`{"update_info": {...}`). Previously, our NCCL path called the same method with positional fields, so update_weights had to support two completely different calling
  conventions.

This means there's one codepath.

finbarrtimbers: I added some `TypedDicts`; wdyt? Is that better?

open_instruct/vllm_utils.py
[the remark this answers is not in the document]
finbarrtimbers: Fixed!

hamishivi (approved):

PR #1650, opened 2026-04-30
Implements IcePop and creates a unified interface for off-policy correction

finbarrtimbers:


open_instruct/grpo_utils.py
```diff
@@ -102,7 +102,15 @@ class GRPOExperimentConfig(
...
-    """The maximum cap for truncated importance sampling ratio (0 means disabled)"""
+    """The maximum cap for truncated importance sampling ratio (0 means disabled)."""
+    use_icepop: bool = False
+    """If True, replace truncated importance sampling with the IcePop M(ρ; 1/β, β)
+    operator from https://arxiv.org/abs/2510.18855: tokens whose train/infer
+    mismatch ratio ρ = π^train_old / π^infer_old falls in [1/β, β] are reweighted
+    by ρ (an IS correction); tokens outside that band have their per-token
+    policy loss zeroed out."""
+    icepop_beta: float = 5.0
+    """IcePop bound β (paper default: 5.0); the in-range band is [1/β, β]."""
```
hamishivi: Would it be better to have icepop_upper and icepop_lower, in case you don't want the strict 1/b, b range? I think in GLM-5 paper they just do lower/higher (https://arxiv.org/pdf/2602.15763).

open_instruct/grpo_utils.py
```diff
@@ -320,21 +330,71 @@ def compute_vllm_local_debug_metrics(
...
     if cap <= 0:
-        return None, None
-    unclamped = torch.ones_like(old_logprob)
-    logprob_diff = old_logprob - vllm_logprobs
-    logprob_diff = torch.where(response_mask, logprob_diff.clamp(-10.0, 10.0), torch.zeros_like(logprob_diff))
-    unclamped = torch.where(response_mask, torch.exp(logprob_diff), unclamped)
-    clamped = torch.clamp(unclamped, max=cap)
-    return clamped, unclamped
+        return RhoCorrection(weights=None, metrics={}, histogram_metrics=rho_hist)
+    clamped = torch.clamp(rho, max=cap)
```
hamishivi: So for icepop, we apply the TIS weights AND pop/mask the ones lying too far out?

open_instruct/grpo_utils.py
```diff
@@ -320,21 +330,71 @@ def compute_vllm_local_debug_metrics(
...
+    (response tokens only); these bypass the scalar reduction and are
+    concatenated across micro-batches and logged as wandb histograms.
     """
+
+    weights: torch.Tensor | None
+    metrics: dict[str, torch.Tensor]
+    histogram_metrics: dict[str, torch.Tensor] = field(default_factory=dict)
+
+
+def compute_rho_correction(
```
hamishivi, line 391: It might be nice to also have a sequence level version or flag, potentially...?

hamishivi (changes requested):
Generally seems good. just some minor comments.

open_instruct/grpo_utils.py
finbarrtimbers: yeah good point, done.

open_instruct/grpo_utils.py
finbarrtimbers: Correct. This is what Kimi and verl do (but not what the original paper does).

open_instruct/grpo_utils.py
finbarrtimbers: so we'd compute it as the average over the sequence? Or how would you imagine that working?

hamishivi: yeah. like deepseek v3.2
image

scripts/train/qwen/qwen3_4b_dapo_math.sh
hamishivi, line 1: can we revert these changed unless you want to include this as the new baseline?

scripts/submit_eval_jobs.py
hamishivi, line 1: Can we put this in a sep pr? At least I'd like it if we had the old submit eval script still around.

open_instruct/grpo_utils.py
```diff
@@ -102,7 +102,22 @@ class GRPOExperimentConfig(
     clip_higher: float = 0.272
     """the higher clip range. Sometimes we want this to be higher, see DAPO (https://arxiv.org/abs/2503.14476)"""
     truncated_importance_sampling_ratio_cap: float = 2.0
-    """The maximum cap for truncated importance sampling ratio (0 means disabled)"""
+    """The maximum cap for truncated importance sampling ratio (0 means disabled)."""
+    use_icepop: bool = False
```
hamishivi: Can we allow stacking TIS and icepop? This allows us to e.g. recreate what periodic is doing. (source: https://x.com/agarwl_/status/2049263264249368906?s=20)

Image

hamishivi: So then we have a mask + a ratio.

hamishivi (approved):
LGTM! One thing is we could make the naming more generic for users, instead of using 'icepop'.

open_instruct/grpo_utils.py
hamishivi: discussed offline, this is already implemented here, i was just confused.

PR #1656, opened 2026-05-05
Add time/per_group_wall_time metric

finbarrtimbers:
## Summary
- Adds a new `time/per_group_wall_time` metric: the **mean** of per-group generation wall-clock times across the batch.
- Distinct from existing `time/getting_response`, which is the **max** of per-group times and gets dominated by tail-inflated groups when many requests are in flight (high `--num_async_steps`, many engines, etc.).
- Underlying timestamps are already captured (`metadata["start_time"]` in `add_request`; `current_time` in `finalize_completed_request`); per-group `generation_time` is already on `TokenStatistics`. This change just plumbs the mean through `make_batch_from_groups` and emits it as a new metric.

Applies to both `grpo_fast.py` and `grpo.py` since they share `DataPreparationActor`.

## Test plan
- [ ] `scripts/train/debug/single_gpu_grpo.sh` succeeds; `time/per_group_wall_time` appears in wandb.
- [ ] `scripts/train/qwen/qwen3_4b_dapo_math_oc.sh` (50 steps): inspect `time/per_group_wall_time` vs `time/getting_response`; mean should be <= max.

### DIFF
--- a/open_instruct/data_loader.py
+++ b/open_instruct/data_loader.py
@@ -630,6 +630,11 @@ class StreamingDataLoader(data_loader.DataLoaderBase):
             trainer_idle_wait_time = time.perf_counter() - wait_start_time
             batch_data.setdefault("metrics", {})["time/trainer_idle_waiting_for_inference"] = trainer_idle_wait_time
             self.training_step = step + 1
+            if len(batch_data["batch"]) == 0:
+                logger.warning(
+                    f"[Dataloader] Empty batch at step={step} (all groups filtered); skipping training step"
+                )
+                continue
             yield batch_data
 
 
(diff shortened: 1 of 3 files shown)


open_instruct/data_loader.py
```diff
@@ -1511,6 +1516,7 @@ def _data_preparation_loop(self):
                 total_tokens = result.token_statistics.num_prompt_tokens + result.token_statistics.num_response_tokens
                 step_metrics["val/actor_tokens_per_second"] = total_tokens / result.token_statistics.generation_time
                 step_metrics["time/getting_response"] = result.token_statistics.generation_time
+                step_metrics["time/per_group_wall_time"] = result.token_statistics.mean_per_group_wall_time
```
hamishivi, line 1519: would median or mean be better? I worry about weird outlier group times?

finbarrtimbers: I like mean for exactly that reason, it'll capture the aggregate times. I worry with the median we won't see the worst times.

PR #1657, opened 2026-05-05
Log filtered prompts at INFO + add batch/filtered_prompts_pct

finbarrtimbers:
## Summary

- Bump per-filter log in `accumulate_inference_batches` from DEBUG to INFO with `training_step`, `sampled/num_prompts`, and the zero/solved/nonzero breakdown so the active-sampling failure modes are obvious from logs alone.
- Add `batch/filtered_prompts_pct = filtered / (filtered + kept)` to `BatchStatistics`, automatically surfaced as a wandb scalar.

## Why

Hit a stuck `qwen3_4b_dapo_math` run (Beaker `01KQT9S2CR2FF6VA8XSG6CMPRN`, wandb `ai2-llm/rl-difficulty-curriculum/fo5n0g8i`) where a grad-norm spike (~11) at step 712 collapsed the policy. Post-collapse, ~99% of rollouts scored 0, so `--active_sampling` filtered hundreds of prompts per step trying to find non-zero variance. From logs the only visible signal was `Got result 1/8` repeating for hours — the filter counters were DEBUG-only, and the `filtered_prompts_zero` vs `filtered_prompts_solved` split (the thing that actually distinguishes "model failing everything" from "model converged") was buried in wandb summary.

`filtered_prompts_pct` makes the regime change obvious in wandb: ~0 when healthy, → 1 when stuck.

## Test plan

- [ ] `make style && make quality` (passes locally).
- [ ] Smoke-launch a small GRPO run and confirm `batch/filtered_prompts_pct` appears in wandb.

GPU_TESTS=bypass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 
 ### Changed
+- Log every filtered prompt in `accumulate_inference_batches` at INFO level with the zero/solved/nonzero breakdown, and add `batch/filtered_prompts_pct` to wandb so policy collapse / convergence is visible without spelunking debug logs.
 - Split `accumulate_inference_batches` into `process_single_result` and `combine_processed_results` for clarity (https://github.com/allenai/open-instruct/pull/1614).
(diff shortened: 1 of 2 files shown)


undfined (approved):
lgtm!

PR #1666, opened 2026-05-08
Makes `grpo.py` support checkpointing, and adds a fix for `mason.py`

finbarrtimbers:
## Summary

- Add `replace_or_append_flag(command, flag, value)` helper in `mason.py` so the auto-overrides for `--output_dir` and `--checkpoint_state_dir` replace any existing occurrence of the flag in the user's command instead of blindly appending. Previously, passing the flag explicitly produced a duplicated entry, and argparse picked one of the two values inconsistently.
- Collapses repeated occurrences of the same flag down to a single one with the new value, including the trailing-flag-without-value edge case.
- Make `build_command_without_args` flag-aware: a value-bearing flag now consumes the following token only if it doesn't itself start with `--`. `replace_or_append_flag` is a one-line delegation on top of it.
- Add `open_instruct/grpo.py` to `OPEN_INSTRUCT_COMMANDS` and `OPEN_INSTRUCT_RESUMABLES` so mason recognizes it the same way it does `grpo_fast.py`, **and** wire OLMo-core checkpoint save/resume into `grpo.py` (`CheckpointerCallback` + `DataPreparationActorCheckpointCallback` + `LoadStrategy.if_available`) so resumable Beaker jobs actually resume instead of silently restarting from step 0.
- Parameterized unit tests for the helper covering empty/absent/present-once/present-twice/trailing-flag/adjacent-flags/flag-followed-by-flag cases, plus two new cases for `build_command_without_args` covering the same edge cases.

Carved out of #1642 to keep the change small and reviewable.

### Example fixed by the `build_command_without_args` change

Given a command where `--checkpoint_state_dir` is immediately followed by another flag (a flag with no value, e.g. an interrupted/edited launch line):

```python
build_command_without_args(
    ["python", "grpo.py", "--checkpoint_state_dir", "--with_tracking", "--output", "out"],
    {"--checkpoint_state_dir": True},
)
```

- Before: `["python", "grpo.py", "--output", "out"]` — `--with_tracking` got silently eaten as if it were the checkpoint dir's value.
- After: `["python", "grpo.py", "--with_tracking", "--output", "out"]` — `--with_tracking` is preserved.

The same fix means `replace_or_append_flag(["--output_dir", "--output_dir", "/tmp/z"], "--output_dir", "/weka/x")` now returns `["--output_dir", "/weka/x"]` instead of leaking the orphaned `/tmp/z` token.

## Test plan

- [x] `uv run pytest test_mason.py`
- [x] End-to-end checkpoint-resume integration test on Beaker (`scripts/train/debug/grpo_checkpoint_integration_test.sh`): run 1 trains 6 steps and writes checkpoints; run 2 resumes from `step6` and trains to step 12. Run 2 logs confirm `Loading checkpoint from '.../step6'` and `Will resume training from step 6, epoch 1`.

Runs:

1. Run 1 (train 6 steps + checkpoint): [Beaker](https://beaker.org/ex/01KR4E636VH5X65QDY2P9XZJ13)
2. Run 2 (resume from step 6 → step 12): [Beaker](https://beaker.org/ex/01KR4EHBEVTS649A0N3SPK92R5)

GPU_TESTS=bypass


open_instruct/grpo_callbacks.py
```diff
@@ -235,7 +235,7 @@ def load_state_dict(self, state_dict: dict[str, Any]) -> None:
 
         try:
             data_prep_actor = ray.get_actor(data_loader_lib.DATA_PREP_ACTOR_NAME)
-            ray.get(data_prep_actor.restore_state.remote(state_dict["data_prep_state"]))
+            ray.get(data_prep_actor.set_state.remote(state_dict["data_prep_state"]))
```
hamishivi, line 238: why this change?

hamishivi (approved):

open_instruct/grpo_callbacks.py
finbarrtimbers: Embarrassingly, `restore_state` doesn't actually exist.

PR #1669, opened 2026-05-08
Move maybe_evaluate to grpo_utils; dedupe calculate_token_counts

finbarrtimbers:
Pure extraction, no behavior change. Sets up OLMo-core GRPO (`grpo.py`) to share the same eval flow as `grpo_fast.py`.


open_instruct/grpo_utils.py
```diff
@@ -4,13 +4,19 @@
...
 import numpy as np
+import pandas as pd
 import ray
+import ray.util.queue as ray_queue
 import torch
 import torch.distributed as dist
+import wandb
+from datasets import Dataset
 
+from open_instruct import data_loader as data_loader_lib
```
farhatkevin, line 19: Do we expect `grpo_utils` importers to always have vllm installed? This PR makes `grpo_utils` import `data_loader` at module load time, which pulls in vllm; I think that changes behavior for non-vllm paths that only use shared helpers like some of the tests. Could a lazy import inside `maybe_evaluate` avoid that?

farhatkevin (approved):

open_instruct/grpo_utils.py
finbarrtimbers: I don't want to do a lazy import as they can make it hard to reason about the code and I think they're a bit messy; my goal is to always have all imports at the top of the file when possible. 

I think that this is fine as it doesn't cause any of the CPU tests to fail, which they would as they don't have vllm installed. You're right to flag this, though! We should be careful about the imports.

PR #1671, opened 2026-05-08
Verify HF export at startup; rewrite save_state_dict_as_hf

finbarrtimbers:
Replace olmo-core's `save_hf_model` path with a direct `convert_state_to_hf` → `AutoModelForCausalLM.save_pretrained` flow, so HF export reuses the same converter the verifier checks against.

Add `verify_can_save_as_hf` which builds the olmo-core model and the target HF model on meta, runs the converter, and asserts the produced state-dict keys exactly match the HF model.


farhatkevin (approved):
This looks good to me once the merge conflicts are resolved and the stale PruningCheckpointerCallback mention is removed from the changelog.

PR #1672, opened 2026-05-08
GRPO OLMo-core feature parity: eval, checkpointer, schedulers

finbarrtimbers:
Brings the OLMo-core GRPO trainer (`grpo.py`) up to feature parity with `grpo_fast.py`:

- **Eval**: new `EvalCallback` that pushes eval prompts onto `prompt_Q` on cadence and drains results via `grpo_utils.maybe_evaluate`; new `setup_eval` actor RPC and `m.setup_eval.remote(...)` call from `grpo.py` main; rank-0-only eval data loader.
- **Checkpointing**: add an OLMo-core `CheckpointerCallback` to `fit()` driven by `--checkpoint_state_freq` and pruning to `--keep_last_n_checkpoints`. Warn if `--save_freq` differs (it's a no-op on the olmo-core path).
- **Scheduler**: add explicit `cosine` / `constant` / `linear` branches for `--lr_scheduler_type` (raise on anything else).
- **StepTimingCallback**: lower `priority` so its `post_step` runs after vLLM sync, and switch to `_last_step_end`-based timing so `time/total` is end-to-end.
- **Scripts**: `qwen3_4b_dapo_math.sh` / `qwen3_4b_dapo_math_oc.sh` — accept `BEAKER_IMAGE` env var, route checkpoints to `/tmp-3m/$RUN_NAME`, add `--use_rho_correction` defaults and bump `--activation_memory_budget` for the OC variant.


open_instruct/grpo_utils.py
```diff
@@ -247,8 +247,16 @@ def __post_init__(self):
             raise ValueError(
                 f"loss_denominator must be a valid float greater than 0 if not 'token', got: {self.loss_denominator}"
             )
-        if self.checkpoint_state_dir is not None and self.checkpoint_state_freq == -1:
+        if self.checkpoint_state_dir is not None and self.checkpoint_state_freq <= 0:
             raise ValueError("`checkpoint_state_freq` must be greater than 0 if `checkpoint_state_dir` is provided!")
+        if self.save_freq != self.checkpoint_state_freq:
```
farhatkevin, line 252: Should this warning live in `grpo.py` instead? `GRPOExperimentConfig` is shared with `grpo_fast.py`, and `grpo_fast.py` still uses `save_freq` for periodic model saves. Putting the warning here means non-Olmo-core runs can see an Olmo-core-specific warning. I know it says "on the olmo-core training path..." but is it better to move it?

farhatkevin (approved):

PR #1675, opened 2026-05-11
Fix: NameError `streaming_config/vllm_config` when importing from grpo_fast.py

BrownianNotion:
Running `grpo_fast.py:main` outside `grpo_fast.py` results in `NameError` due to `streaming_config/vllm_config` not being defined/passed correctly.

Specifically, these variables weren't being defined/passed properly in their local scope. This is fine when running `grpo_fast.py` directly as the variables are defined as global module constants under the `if __name__ == "__main__"` block. However, when importing from `grpo_fast.py`, the `if __name__ == "__main__"`  block doesn't run, leading to a NameError.


finbarrtimbers (approved):
Nice catch. Whoops!

PR #1677, opened 2026-05-11
Use processed vLLM logprobs for GRPO

farhatkevin:
## Summary
- Add a GRPO vLLM config knob for logprobs mode and default it to processed logprobs.
- Pass the configured mode through to vLLM engine creation so rollout logprobs are comparable to learner logprobs after sampling transforms.

### DIFF
--- a/open_instruct/grpo_fast.py
+++ b/open_instruct/grpo_fast.py
@@ -1415,6 +1415,7 @@ def create_model_and_optimizer(
         train_dataset=train_dataset,
         eval_dataset=eval_dataset,
         vllm_attention_backend=vllm_config.vllm_attention_backend,
+        vllm_logprobs_mode=vllm_config.vllm_logprobs_mode,
     )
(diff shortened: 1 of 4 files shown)


open_instruct/grpo_fast.py
```diff
@@ -1415,6 +1415,7 @@ def create_model_and_optimizer(
         train_dataset=train_dataset,
         eval_dataset=eval_dataset,
         vllm_attention_backend=vllm_config.vllm_attention_backend,
+        vllm_logprobs_mode=vllm_config.vllm_logprobs_mode,
```
finbarrtimbers, line 1418: I don't think we ever want to configure it, I think that `processed_logprobs` is the value we always want. Let's hardcode it to `processed_logprobs`.

finbarrtimbers (changes requested):

farhatkevin:
Superseded by #1678, which targets main from the clean kevinf/fix-vllm-logprobs-main branch.

PR #1682, opened 2026-05-12
Test combine_dataset with list and dict mixers against local fixtures

finbarrtimbers:
#1674 indicated a lack of coverage in our tests. This PR rectifies that, and refactors the code to have a single, `dict` path for the mixers, adding a function + tests to convert from the list format (only used by one caller) to the dict format, and adds a bunch of test coverage to the mixing code using the saved test data for the RLVR/SFT tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 
 ### Changed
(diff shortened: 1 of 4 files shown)


natolambert (approved):
LGTM, thanks!

PR #1683, opened 2026-05-12
Fix IFEval correctness bugs in if_functions and IFEvalVerifier

finbarrtimbers:
Bundles a bunch of IFEval correctness fixes:

- `open_instruct/if_functions.py`: fix `validate_choice` operand direction (was `text in option`, now `option in text`); add ±10% tolerance to `validate_frequency_capital_words` "around" quantifier (supersedes #1615, #1646).
- `open_instruct/ground_truth_utils.py`: guard `IFEvalVerifier.__call__` against `ZeroDivisionError` when the constraint's instruction list is empty (supersedes #1655).


hamishivi (approved):
ahh thanks for this

PR #1690, opened 2026-05-12
Emit per-Group generation times and rename trainer-idle metric

finbarrtimbers:
## Summary

- Rename `time/trainer_idle_waiting_for_inference` → `time/trainer_waiting_for_data` so the metric reads as "trainer waiting for data" rather than "idle waiting for inference."
- Stop max-reducing per-Group `generation_time` before reporting. `make_batch_from_groups` now keeps the full per-Group list in `BatchStatistics.per_group_generation_times`; `DataPreparationActor` emits `time/group_generation_{mean,max,min}` per step, and the histogram is auto-surfaced as `batch/per_group_generation_times`.
- Drop `time/getting_response` (it was identical to `time/group_generation_max`). Downstream consumers (`grpo_callbacks.py`, `grpo_fast.py` MFU/MBU calc) now read `time/group_generation_max` directly.

Motivation: `time/getting_response` is the slowest single Group's latency, not the batch's wall-clock. Comparing it against `time/training` makes inference look like the bottleneck even when the pipelined throughput across vllm engines matches the trainer. The per-Group min/mean/max plus the renamed wait metric together give a clean read on whether the trainer is actually starving.

## Verification

Runs:

1. Single-GPU smoke (gsm8k, Qwen3-0.6B): [Beaker](https://beaker.org/ex/01KRENN86CZMYXK8SM7E7AB9W8), wandb [7838rozw](https://wandb.ai/ai2-llm/open_instruct_internal/runs/7838rozw)
2. Large multi-node (RLVR code, Qwen2.5-7B, 16×H100): [Beaker](https://beaker.org/ex/01KRENNN35AQY9BZPF3TV7X9RH), wandb [vc61aeb9](https://wandb.ai/ai2-llm/open_instruct_internal/runs/vc61aeb9)

Both succeeded. New metrics confirmed:

| metric | single-GPU | large |
|---|---|---|
| `time/group_generation_min` | 6.05s | 2.43s |
| `time/group_generation_mean` | 12.92s | 6.29s |
| `time/group_generation_max` | 19.68s | 12.62s |
| `time/trainer_waiting_for_data` | 0.03s | 0.03s |
| `time/training` | 4.06s | 4.54s |

Large run shows the intended signal: trainer wait ~0 while inference idle is 25s, clearly trainer-bound. `time/getting_response` absent in both; `batch/per_group_generation_times` histogram present.

## Test plan

- [x] Single-GPU smoke
- [x] Large multi-node smoke

GPU_TESTS=bypass


mnoukhov (changes requested):
Looks good. Probably also worth renaming `generation_idle` to something similar then?

mnoukhov (approved):

PR #1695, opened 2026-05-18
fix: guard against empty choices and null message in LLM responses

qizwiz:
## Summary

Six locations across five files access `response.choices[0].message.content` without first verifying that `choices` is non-empty and `message` is not `None`.

Two crash vectors that bypass existing try/except blocks:

- **IndexError** — API returns an empty `choices` list (quota exhaustion, network failure, provider-side filtering)
- **AttributeError** — `choices[0].message` is `None` (documented Gemini `PROHIBITED_CONTENT` behaviour: HTTP 200 response with null message object rather than an error status)

## Files changed

| File | Locations | Guard action |
|------|-----------|-------------|
| `open_instruct/rubrics/run_utils.py` | 2 | Return `""` (matches existing except-path behaviour) |
| `open_instruct/ground_truth_utils.py` | 1 | Raise `ValueError` → caught by `except Exception` → returns zero-values |
| `open_instruct/rejection_sampling/synthetic_preference_dataset.py` | 1 | Raise `ValueError` → caught by retry-loop `except Exception` |
| `scripts/does_prompt_make_sense.py` | 1 | Raise `ValueError` → caught by retry-loop `except Exception` |
| `scripts/data/rlvr_code/the_algorithms.py` | 1 | Raise `ValueError` → caught by outer `except` → returns error dict |
| `scripts/data/rlvr_code/sft_to_rlvr_azure.py` | 1 | Guard before access; also simplifies redundant double-access |

## Pattern

```python
# Before
r = response.choices[0].message.content

# After
if not response.choices or response.choices[0].message is None:
    raise ValueError("LLM returned empty or filtered response")
r = response.choices[0].message.content
```

## References

- [OpenAI API docs — finish_reason content_filter](https://platform.openai.com/docs/guides/chat-completions)
- Detected by [pact](https://github.com/qizwiz/pact) static analysis (`llm_response_unguarded` rule)

### DIFF
--- a/open_instruct/ground_truth_utils.py
+++ b/open_instruct/ground_truth_utils.py
@@ -723,7 +723,7 @@ class LMJudgeVerifier(VerifierFunction):
             return reasoning, score
 
         try:
-            if not completion.choices or completion.choices[0].message is None:
+            if not completion.choices or completion.choices[0].message is None or completion.choices[0].message.content is None:
                 raise ValueError("LLM returned empty or filtered response")
             # remove anything between <think> and </think> including the tags using regex
             pattern = r"<think>\s*.*?\s*</think>\s*"
--- a/open_instruct/ground_truth_utils.py
+++ b/open_instruct/ground_truth_utils.py
@@ -723,6 +723,8 @@ class LMJudgeVerifier(VerifierFunction):
             return reasoning, score
 
         try:
+            if not completion.choices or completion.choices[0].message is None:
+                raise ValueError("LLM returned empty or filtered response")
             # remove anything between <think> and </think> including the tags using regex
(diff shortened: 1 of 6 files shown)


qizwiz:
Good catch @gemini-code-assist — extended all six guards to additionally check `message.content is None` (commit 9f3f166), covering providers that return a non-None message object but set content to None on max-token truncation or certain filter responses.

qizwiz:
Thanks for the review @gemini-code-assist. Pushed 9f3f166 extending all six guards to also check `message.content is None` — this covers providers that return a non-`None` message object with `content=None` (e.g. on max-token truncation or tool-call-only responses).

PR #1699, opened 2026-05-27
Fix validate_choice operand direction

Chessing234:
## Bug
`validate_choice` is backwards. When evaluating options like `['yes', 'no']` against a response like `"The answer is yes"`, the validation function returns `False`.

## Root cause
The operand check is `any(text in option for option in options)` (and a similar `text in option` loop in the scripts variant). This checks if the response `"The answer is yes"` is a substring of the option `"yes"`, which is backwards.

## Why fix is correct
Swapping the check to `option in text` checks if the option is a substring of the response, restoring the correct membership logic and matching the other `validate_*` functions in the module.


Chessing234:
Thanks for the feedback! I've updated the validation checks to be case-insensitive by using `.lower()` on both the option and the text.

PR #1700, opened 2026-05-27
Fix validate_choice false positives on substring matching using regex word boundaries

Chessing234:
**Bug:** `validate_choice` uses simple substring matching (`any(option in text for option in options)`), which incorrectly flags substring matches as correct. For example, if the option is `"A"` and the text is `"I love Apple"`, it returns `True`.

**Root cause:** Lacking word boundary enforcement during the substring search.

**Why fix is correct:** Enforcing word boundaries with a regex `(?<!\w){re.escape(str(option))}(?!\w)` ensures that the option is matched as a standalone word/choice, matching the intended behavior of IFEval.

### DIFF
--- a/open_instruct/if_functions.py
+++ b/open_instruct/if_functions.py
@@ -360,9 +360,9 @@ def validate_title(text: str) -> bool:
 
 # Choose: From Answer with one of the following options: {options}
 def validate_choice(text: str, options: list) -> bool:
-    # substring matching yields false positives (e.g. option "A" matches "Apple"); if this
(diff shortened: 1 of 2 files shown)


Chessing234:
Thanks for the review! I've removed the redundant inline `import re` and the unnecessary `bool()` conversion. Let me know if there's anything else!

Chessing234:
Thanks for the review! I've removed the redundant inline import `re` and the unnecessary `bool()` conversion. Let me know if there's anything else!

PR #1704, opened 2026-05-28
Fix sentence terminator and whitespace handling in verify_sentence_constraint

Chessing234:
Bug: The sentence constraint logic was failing on trailing spaces or multiple spaces between sentences, resulting in incorrect sentence counts. Root cause: The regex split used \s which left empty strings when splitting on multiple spaces, and it didn't filter out empty strings resulting from trailing spaces. Why fix is correct: Used \s+ in the regex and added a list comprehension to filter out empty strings, ensuring only actual sentences are counted.


Chessing234:
Thanks for the feedback! I've updated the logic to strip the leading and trailing whitespace from each resulting sentence to ensure they are fully clean.

PR #1708, opened 2026-05-28
Fix GRPO OLMo-core bookkeeping PG deadlock + Qwen3 parity tweaks

finbarrtimbers:
## Summary

- **Fix `reduce_metrics` deadlock in GRPO + OLMo-core** (`olmo_core_train_modules.py`): record a `_metrics_keepalive` metric on every rank every step, outside the `num_steps > 0` gate. OLMo-core's `_log_metrics` skips submission when `self._metrics` is empty (trainer.py:1395); under `--active_sampling --inflight_updates` a learner rank can reach a save-time flush with no recorded metrics while peers have some, desyncing the (untagged) bookkeeping process group's collective stream and cross-matching `reduce_metrics` against `_check_if_canceled`. That cross-match hangs gloo for its full 30-min timeout and kills training (verified previously in run `e6yft2fs`).
- **Qwen3 OLMo-core parity** (carried from prior work on this branch): bump OLMo-core to `finbarr/tie-lm` to enable tied word embeddings for Qwen3, use fp32 RoPE in Qwen3 OLMo-core configs, and switch GRPO OLMo-core metrics-collect-interval to 1 so logs match grpo_fast cadence.
- Switch `qwen3_4b_dapo_math_oc.sh` eval datasets to `allenai/aime_2025_openinstruct` / `allenai/brumo_2025_openinstruct`, and drop a stale `--budget` flag from `qwen3_4b_dapo_math.sh`.

## Test plan

- [x] `make style && make quality` clean.
- [x] Reproducer (`qwen3_4b_dapo_math_oc.sh`) launched on Beaker: previously hung 30 min on `reduce_metrics` at the step-200 save. With the fix, the step-200 checkpoint saves in 32 seconds and training continues past step 240+.

Runs:

1. Deadlock-fix verification (currently running, past step 240): [Beaker](https://beaker.org/ex/01KSQHT1KDEMYSSR16P5574PS1)

GPU_TESTS=bypass


open_instruct/olmo_core_train_modules.py
```diff
@@ -557,6 +555,11 @@ def train_batch(self, batch: dict[str, Any], dry_run: bool = False) -> None:
                 self._do_optim_step()
             self.optim.zero_grad(set_to_none=True)
 
+        # Keep _metrics non-empty on every rank every step so OLMo-core's
+        # _log_metrics skip can't fire asymmetrically and deadlock gloo on the
+        # untagged bookkeeping process group.
```
mnoukhov, line 562: This was a deep catch, nice. Can't decide if that is a problem with Olmo-core or not, but this is def the right decision for now

mnoukhov (approved):
lgtm! ty Finbarr 💯

PR #1714, opened 2026-06-04
Expose olmo-core's lm_head loss_implementation (e.g. Liger fused_linear) on ModelConfig

ReinforcedKnowledge:
Adds `ModelConfig.loss_implementation` to select olmo-core's LM loss implementation (default / `fused_linear` for Liger FLCE). `setup_model` applies it to the transformer config before building and only consulted when labels are passed into the model (SFT). Won't affect DPO which computes loss from logits outside the `lm_head`.

Try this with:
```python
from olmo_core.nn.lm_head import LMLossImplementation

from open_instruct.olmo_core_utils import ModelConfig, setup_model
from open_instruct.utils import ArgumentParserPlus

parser = ArgumentParserPlus((ModelConfig,))

(mc,) = parser.parse_args_into_dataclasses(["--model_name_or_path", "Qwen/Qwen3-8B", "--loss_implementation", "fused_linear"])
model, _ = setup_model(mc, init_device="meta")
assert model.lm_head.loss_implementation == LMLossImplementation.fused_linear

(mc,) = parser.parse_args_into_dataclasses(["--model_name_or_path", "Qwen/Qwen3-8B"])
model, _ = setup_model(mc, init_device="meta")
assert model.lm_head.loss_implementation == LMLossImplementation.default

print("PASS")
```

You should be able to put this script in your CI as well I think. But you'll notice this:
```
Before: ValueError: Some specified arguments are not used by the HfArgumentParser: ['--loss_implementation', 'fused_linear']
After: PASS
```

On 1×A100, Qwen3-8B head dims: vocab 151936, seq 8192, bf16, `fused_linear` saved 8.6 GB peak but ran 2.7× slower than the default, this PR only exposes the choice though and doesn't change it

GPU_TESTS=bypass


open_instruct/olmo_core_utils.py
```diff
@@ -58,6 +59,9 @@ class ModelConfig:
     """Pretrained config name or path if not the same as model_name."""
     attn_implementation: AttentionBackendName | None = None
     """Which attention implementation to use. If None, auto-detects the best available."""
+    loss_implementation: LMLossImplementation | None = None
```
finbarrtimbers: Please make the default `LMLossImplementation.default` to avoid having to handle `None` as a special case.

finbarrtimbers (dismissed):

open_instruct/olmo_core_utils.py
mnoukhov: Fixed

mnoukhov (approved):

PR #1725, opened 2026-06-15
Change default temperature to 1.0

mnoukhov:
## Summary

- Changes the default generation `temperature` from 0.7 to **1.0**.
- Consolidates to a single source of truth: `StreamingConfig.temperature` (in `data_loader.py`) is the canonical default. Both production call sites (`grpo_fast.py` and `benchmark_generators.py`) already pass `temperature=streaming_config.temperature` explicitly into `SamplingConfig`, so `SamplingConfig`'s own default was a redundant, never-relied-upon fallback.
- Makes `SamplingConfig.temperature` a required field (no default) so there is exactly one place defining the temperature default.

## Test plan

- `make style && make quality` pass.
- Verified every `SamplingConfig(...)` construction (production + tests) passes `temperature` explicitly, so the now-required field has no missing-argument call sites.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 ### Changed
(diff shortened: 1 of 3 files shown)


farhatkevin (approved):
lgtm!

PR #1730, opened 2026-06-23
Guard process_tool_tokens against None tool outputs

hamishivi:
## Summary
- Coerce tool outputs to strings (treating `None` as empty) and ensure the formatted output is a string before tokenizing in `process_tool_tokens`, guarding against `None`/non-string values that can surface from Ray serialization.

## Test plan
- Existing tool-token tests.

GPU_TESTS=bypass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,3 +16,4 @@ All notable changes to this project will be documented in this file.
 ### Removed
 
 ### Fixed
+- Guard process_tool_tokens against None tool outputs (https://github.com/allenai/open-instruct/pull/1730).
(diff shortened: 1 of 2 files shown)


hamishivi:
Closing as redundant. The base `ToolParser.format_tool_outputs` (open_instruct/environments/tools/parsers.py) already performs exactly this sanitization — it coerces `None`/non-string elements in `tool_outputs` to strings and guarantees the formatted output is a string (and additionally strips null bytes). The guard added here in `process_tool_tokens` duplicates that, so there's nothing left for this PR to do.

PR #1731, opened 2026-06-23
Make ModelDims.from_hf_config robust to explicit head_dim

hamishivi:
## Summary
- Honor an explicit `head_dim` from HF configs (e.g. composite/VLM models) instead of always deriving it from `hidden_size // num_attention_heads`, which fails when `hidden_size` is not divisible by `num_attention_heads`.
- Relax the `__post_init__` assertion to require a positive `head_dim`.

## Test plan
- Existing ModelDims tests.

GPU_TESTS=bypass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 ### Changed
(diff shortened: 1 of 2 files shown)


farhatkevin (approved):

hamishivi:
Closing in favor of replacement PR #1743, rebased onto current main and pushed from hamishivi/open-instruct.

PR #1733, opened 2026-06-23
Add DPPO loss function

hamishivi:
## Summary
- Add the DPPO policy-loss function (`--loss_fn dppo`) from the [DPPO paper](https://arxiv.org/abs/2602.04879).
- DPPO enforces a trust region via a per-token binary (Bernoulli) divergence between the rollout policy μ_θ' and the trainer policy: tokens whose update would push the policy away from μ_θ' beyond a threshold δ (and only in the "unsafe" direction) are masked out of the policy gradient. There is no symmetric clipping (Eq. 11: `L_DPPO = E[Σ_t M_t · r_t · A_t]`).
- New config knobs `--dppo_divergence_type` (`tv`/`kl`) and `--dppo_divergence_threshold` (δ).
- Requires `--use_vllm_logprobs` so the importance ratio is anchored on the rollout policy (π_θ / μ_θ'); validated in `__post_init__`.
- Wired into the `grpo_fast`/DeepSpeed loss path: the mask is computed at the call site from the rollout (old) logprobs and passed into `compute_grpo_loss`.

## Implementation notes
- `compute_binary_divergence` and `compute_dppo_mask` are pure, well-tested functions added to `grpo_utils.py`.
- `compute_grpo_loss` gains a `dppo` branch and an optional `dppo_mask` argument; existing `dapo`/`cispo` behavior is unchanged.

## Test plan
- New CPU unit tests in `open_instruct/test_grpo_utils_dppo.py` cover the TV/KL divergence math, response-mask zeroing, the asymmetric trust-region masking, gradient detachment, the `dppo` loss branch, and config validation (positive threshold + `use_vllm_logprobs` requirement).

GPU_TESTS=bypass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 
 
 ### Added
-- Add the DPPO loss function (`--loss_fn dppo`), a binary-divergence trust-region policy update (https://arxiv.org/abs/2602.04879) gated on `--use_vllm_logprobs` (https://github.com/allenai/open-instruct/pull/PR_NUMBER).
+- Add the DPPO loss function (`--loss_fn dppo`), a binary-divergence trust-region policy update (https://arxiv.org/abs/2602.04879) gated on `--use_vllm_logprobs` (https://github.com/allenai/open-instruct/pull/1733).
 
 ### Changed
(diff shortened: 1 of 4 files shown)


hamishivi:
Closing in favor of replacement PR #1745, rebased onto current main and pushed from hamishivi/open-instruct.

PR #1734, opened 2026-06-23
Add tool-schema support to SFT tokenization

hamishivi:
## Summary
- Pass the dataset `tools` column to `apply_chat_template` during SFT tokenization so tool schemas are rendered into the prompt. A new `_normalize_tools_for_chat_template` helper accepts a list/dict/JSON-string (and treats empty/`None` as no tools), rejecting tool-name-string lists.
- Derive assistant labels from token offset mappings against the rendered conversation (prefix-stable per-assistant-turn spans) instead of `mask_labels`, so labels stay correct when tools are present.
- Consume the `tools` column during SFT tokenization rather than persisting it to the cached dataset (`_SFT_TOKENIZE_FNS`).
- Add `sft_tulu_tokenize_without_truncation_v1` and register `last_turn_tulu_tokenize_and_truncate_v1` (which now also forwards `tools`).
- Bump `DATASET_CACHE_VERSION` to `v7` to invalidate stale caches.

## Notes
- For the no-tools tulu path the new tokenizer is byte-identical to the previous `mask_labels` path: the existing `GOLD_SFT` golden-hash test passes unchanged.
- This PR intentionally excludes the unrelated `dataset_config_name` plumbing and the `sft_filter_v1` rename that were bundled alongside this work on the source branch.

## Test plan
- `uv run pytest open_instruct/test_dataset_transformation.py` (20 passed locally), including:
  - existing `GOLD_SFT`/preference/rlvr golden-hash tests (unchanged),
  - new `TestToolNormalization` (JSON parsing, dict-wrapping, empty handling, rejection cases),
  - new `TestSFTTuluTokenizeLabels` (assistant-only trainable labels, tools column consumed, no-truncation variant).

GPU_TESTS=bypass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 
 ### Added
+- Add tool-schema support to SFT tokenization: the `tools` column is parsed (JSON strings accepted) and passed to `apply_chat_template`, assistant labels are derived from offset mappings, and the tools column is consumed rather than persisted (https://github.com/allenai/open-instruct/pull/PR_NUMBER).
 
 ### Changed
 - Change the default generation `temperature` to 1.0 and make `SamplingConfig.temperature` a required field so `StreamingConfig.temperature` is the single source of truth (https://github.com/allenai/open-instruct/pull/1725).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 
 
 ### Added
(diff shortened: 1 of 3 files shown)


farhatkevin (approved):
Just to make sure I understand: should the older sft_tokenize_v1 / sft_tokenize_mask_out_prompt_v1 paths also be treated as consuming tools, even though only the Tulu SFT tokenizers pass tools into apply_chat_template?

hamishivi:
Closing in favor of replacement PR #1746, rebased onto current main and pushed from hamishivi/open-instruct.

PR #1735, opened 2026-06-23
Add tiled GRPO lm-head loss

hamishivi:
## Summary
- Add an opt-in, memory-efficient GRPO loss path (`--use_liger_grpo_loss`, `--liger_grpo_loss_chunk_size`) that follows DeepSpeed's `TiledFusedLogitsLoss` pattern: the lm-head projection and scalar loss are recomputed tile-by-tile in a custom autograd `Function`, so the full-vocabulary logits are never materialized. This is a large peak-memory win for large vocabularies / long context.
- The tiled kernel reproduces the existing DAPO/CISPO objective exactly — `masked_mean(pg + beta * kl, mask, None, loss_denominator) * (world_size // sequence_parallel_size)` — so a run can toggle the flag on/off without changing the objective.
- **Default-off**: when `use_liger_grpo_loss=False` the normal loss path is completely unchanged. When on, a dedicated `_train_sample_liger_grpo` step runs the backbone (no lm-head) to get hidden states, then the tiled loss.
- New helpers in `grpo_utils`: `tiled_grpo_lm_head_loss`/`TiledGRPOLMHeadLoss`, `get_causal_lm_backbone_and_lm_head`, `forward_for_liger_hidden_states`. Config validation restricts the path to `loss_fn in {dapo, cispo}`, `kl_estimator=2`, and `record_entropy=False`.

## Scope
This is a focused port adapted to `main`'s current loss set: the source branch's kernel also embedded DPPO/TVPO, sequence-denominator, and freeze/policy-mask branches that don't exist on `main`; those are intentionally **not** included here.

## Validation
- **Correctness (CPU):** new `open_instruct/test_grpo_utils_tiled_loss.py` asserts the tiled loss and gradients (w.r.t. hidden states **and** the lm-head weight) match a dense reference across DAPO/CISPO × {with, without} ref-KL × shards {1, 3, 100} × temperature {1.0, 0.7}. I separately reproduced this with a standalone script — all cases match to 1e-5.
- **Still needs a GPU run:** the DeepSpeed ZeRO-3 integration (per-tile gradient reduction via `param.ds_grad_is_ready`) can only be exercised on GPU. Please do a GPU end-to-end smoke run before merging. I was unable to run GPU/DeepSpeed locally.

GPU_TESTS=bypass


hamishivi:
Closing in favor of replacement PR #1747, rebased onto current main and pushed from hamishivi/open-instruct.

PR #1738, opened 2026-06-25
Drop stale async rollout results

hamishivi:
## Summary
- Add `max_result_age_steps` to `accumulate_inference_batches`: in async GRPO, a rollout whose generating policy is more than that many steps behind the current trainer step (`training_step - result.model_step > max_result_age_steps`) is dropped instead of trained on.
- When a stale result is dropped, a fresh prompt is replenished into the generator (when `replenish_prompts` is on) so throughput is unaffected, and the result is not kept for requeue-on-timeout.
- Logs a `stale_results_dropped` reward metric per accumulation.
- Wired from `DataPreparationActor` using `async_steps` as the staleness threshold (results more than one async window behind are dropped).

## Notes
- `model_step` tracking and the `model_step_min/max/mean` / `num_steps_off_policy` metrics already existed on main; this only adds the drop policy on top.
- The staleness predicate is factored into a small pure `result_is_stale(...)` helper for testability.

## Test plan
- New unit tests `TestResultIsStale` cover: disabled when `max_result_age_steps`/`training_step`/`model_step` is `None`, stale when lag exceeds the threshold, not stale at the threshold boundary, and not stale when fresh.
- `ruff`, formatting, and whole-project `ty check` pass.

GPU_TESTS=bypass


farhatkevin (approved):
looks good!

PR #1732, opened 2026-06-23
Harden DockerBackend with retries, OOM detection, and concurrency limits

hamishivi:
## Summary
- Add cross-process start/exec concurrency semaphores (`SWERL_DOCKER_START_CONCURRENCY` / `SWERL_DOCKER_EXEC_CONCURRENCY`) to avoid overwhelming the Docker daemon under many concurrent rollouts.
- Retry transient Docker exec `APIError`s (e.g. "database is locked") on the same container with exponential backoff, and restart-and-retry once on `NotFound` / 409 Conflict.
- Detect OOM-killed containers and raise `SandboxOOMError` so episodes terminate cleanly instead of looping.
- Add per-command `timeout` override to `run_command`, optional `docker_host` endpoint (e.g. Podman sockets), best-effort container-state diagnostics, and opt-in timing logs.
- Close containers kill-first (then stop) for faster, more reliable teardown.

## Test plan
- New unit tests in `tests/test_docker_backend.py` cover client selection, transient-error retry/backoff, retry exhaustion, and non-retry of unknown errors (mocked Docker SDK, no daemon needed).

GPU_TESTS=bypass

Made with [Cursor](https://cursor.com)

PR #1744, opened 2026-06-28
Harden DockerBackend with retries, OOM detection, and concurrency limits

hamishivi:
Replacement for #1732, rebased onto current `main` and pushed from `hamishivi/open-instruct`.

## Summary
- Add cross-process start/exec concurrency semaphores (`SWERL_DOCKER_START_CONCURRENCY` / `SWERL_DOCKER_EXEC_CONCURRENCY`) to avoid overwhelming the Docker daemon under many concurrent rollouts.
- Retry transient Docker exec `APIError`s (e.g. "database is locked") on the same container with exponential backoff, and restart-and-retry once on `NotFound` / 409 Conflict.
- Detect OOM-killed containers and raise `SandboxOOMError` so episodes terminate cleanly instead of looping.
- Add per-command `timeout` override to `run_command`, optional `docker_host` endpoint (e.g. Podman sockets), best-effort container-state diagnostics, and opt-in timing logs.
- Close containers kill-first (then stop) for faster, more reliable teardown.

## Test plan
- New unit tests in `tests/test_docker_backend.py` cover client selection, transient-error retry/backoff, retry exhaustion, and non-retry of unknown errors (mocked Docker SDK, no daemon needed).

GPU_TESTS=bypass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 
 ### Added
+- Harden DockerBackend with transient-exec retries, OOM detection, and concurrency limits (https://github.com/allenai/open-instruct/pull/1732).
 - Drop stale async rollout results whose generating policy is more than `async_steps` behind the trainer (`max_result_age_steps`), replenishing a fresh prompt and logging a `stale_results_dropped` metric (https://github.com/allenai/open-instruct/pull/1738).
 
 ### Changed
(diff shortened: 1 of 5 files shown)


farhatkevin:
Superseded by #1784, which carries these commits rebased onto current `main` from a branch in `allenai/open-instruct`.

Moved because GitHub Actions stopped creating workflow runs for pushes to this fork branch (only GitGuardian reported), so review fixes could not be verified here. #1784 also fixes the OOM path: `SandboxOOMError` was raised but never caught, so it fell through to the generic handler in `vllm_utils` which scores 0 without setting `rollout.done` — the episode kept stepping and re-tripping the dead container instead of terminating.

PR #1701, opened 2026-05-27
Wire max_checkpoints through SFT, DPO, and GRPO paths

TimDettmers:
## Summary
- Passes `keep_last_n_checkpoints` (default: 3, from `CheckpointConfig`) to `CheckpointerCallback.max_checkpoints` across all OLMo-core training paths (SFT, DPO, GRPO)
- Centralizes the `-1 → None` (unlimited) convention inside `build_checkpointer_callback` so call sites pass the value directly
- `GRPOExperimentConfig` already inherits `keep_last_n_checkpoints` from `CheckpointConfig` — no new config fields needed

## Depends on
- allenai/OLMo-core#694 (adds `max_checkpoints` to `CheckpointerCallback`)

## Files changed
- `olmo_core_utils.py` — `build_checkpointer_callback` and `build_base_callbacks` accept and pass through `max_checkpoints`; `-1` mapped to `None` inside the builder
- `dpo.py` — passes `args.keep_last_n_checkpoints` through
- `olmo_core_finetune.py` — passes `args.checkpoint.keep_last_n_checkpoints` through
- `grpo_olmo_core_actor.py` — passes `self.grpo_config.keep_last_n_checkpoints` through

## Test plan
- [ ] Verify SFT training with default `keep_last_n_checkpoints=3` trims checkpoints
- [ ] Verify GRPO training with `--keep_last_n_checkpoints -1` keeps all checkpoints
- [ ] Verify DPO training passes the parameter correctly
- [ ] Existing tests pass unchanged (new parameter has defaults matching current behavior)

PR #1754, opened 2026-07-13
Add keep_last_n_saves to bound intermediate HF-format model saves

mnoukhov:
## Summary
- Adds `keep_last_n_saves` (default `-1`, unlimited) to `GRPOExperimentConfig`, bounding the number of intermediate HuggingFace-format model saves kept under `{output_dir}_checkpoints`.
- `grpo_fast.py`: `maybe_save_checkpoint` now trims old step saves via the existing `clean_last_n_checkpoints` helper after each periodic save.
- `grpo.py` (OLMo-core path): previously `--save_freq` was a no-op for periodic saves -- only the final model was ever saved in HF format (see the warning in `GRPOExperimentConfig.__post_init__`). Adds a new `HFCheckpointCallback` (in `grpo_callbacks.py`, wired in `grpo_olmo_core_actor.py`) that performs periodic HF-format saves on the `save_freq` cadence, gathering the FSDP-sharded state dict and writing/trimming from rank 0, mirroring the streaming path's behavior.

This is similar in spirit to #1701, which wired `keep_last_n_checkpoints`/`max_checkpoints` through the OLMo-core *native* (resumable) checkpoints. `keep_last_n_saves` is a distinct, independent knob for the separate HF-format saves used for eval/pushing during training.

## Test plan
- [x] `uv run pytest open_instruct/test_grpo_fast.py open_instruct/test_grpo_utils.py open_instruct/test_utils.py` (excluding an unrelated, pre-existing local test-data failure in `CombineDatasetTest`)
- [ ] GPU smoke test: OLMo-core GRPO run with `--save_freq` and `--keep_last_n_saves` set, verify intermediate HF saves are trimmed under `{output_dir}_checkpoints`
- [ ] GPU smoke test: DeepSpeed GRPO run (`grpo_fast.py`) with `--keep_last_n_saves`, verify trimming behavior unchanged from `keep_last_n_checkpoints`'s existing semantics

GPU_TESTS=bypass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 
 ### Added
+- Add `keep_last_n_saves` to `grpo.py`/`grpo_fast.py` to bound the number of intermediate HuggingFace-format model saves kept under `{output_dir}_checkpoints` (-1 for unlimited, matching current behavior); on the OLMo-core (`grpo.py`) training path, periodic HF-format saves following `save_freq` are now performed via a new `HFCheckpointCallback`, since previously `save_freq` was a no-op there (only the final model was saved in HF format). Builds on the `keep_last_n_checkpoints`/`max_checkpoints` wiring from https://github.com/allenai/open-instruct/pull/1701 (https://github.com/allenai/open-instruct/pull/1754).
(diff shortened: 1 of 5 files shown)


mnoukhov:
Good catch, fixed in 1254c7c2f — non-zero ranks now just call `full_tensor()` to satisfy the collective and discard immediately instead of copying the full state dict to CPU on every rank.

PR #1758, opened 2026-07-15
Upgrade to CUDA 13.0 for B300 support on ai2/holmes

mnoukhov:
## Summary

Adds B300 support on `ai2/holmes` without dropping compatibility with CUDA 12 clusters. The same Dockerfile and uv lock can now produce either stack:

- **Selectable Docker base**: `CUDA_VERSION=12` uses `nvidia/cuda:12.8.1-devel-ubuntu22.04`; `CUDA_VERSION=13` uses `nvidia/cuda:13.0.3-devel-ubuntu22.04`.
- **Build launcher**: `build_image_and_launch.sh` and its dirty-worktree variant accept `--cuda-version 12|13` (default: 12). Image names include `-cuda12` or `-cuda13`, so both variants can exist for the same commit.
- **Dependency variants**: mutually exclusive uv `cuda12` and `cuda13` dependency groups select matching torch, torchvision, torchaudio, vLLM, flash-attn, and flash-attn-3 builds from one `pyproject.toml` / `uv.lock`. CUDA 12 remains the default for normal `uv sync` and `uv run` commands.
- **GPU test routing**: CUDA 12 images can run on `ai2/jupiter`, `ai2/ceres`, or `ai2/saturn`; CUDA 13 images run only on `ai2/holmes`. The image suffix selects the cluster pool, while tests detect the image's actual CUDA build from `torch.version.cuda`.
- **Generation goldens**: retains separate deterministic no-tools outputs for CUDA 12 and CUDA 13.
- **B300 support**: adds `b300` to `GPU_SPECS`, `ai2/holmes` to cluster lists, and tests the B300 name/spec mapping.

Usage:

```bash
./scripts/train/build_image_and_launch.sh --cuda-version 12 scripts/test/run_gpu_pytest.sh
./scripts/train/build_image_and_launch.sh --cuda-version 13 scripts/test/run_gpu_pytest.sh
```

## Notes

- CUDA 13 requires NVIDIA driver >= 580, so the CUDA 13 test target is `ai2/holmes`.
- The CUDA 12 image built and uploaded successfully at commit `c5c74b240`; its Jupiter GPU test ran with the CUDA 12 image and exposed a deterministic-generation golden mismatch.
- Both dependency variants pass frozen uv resolution checks and both Docker variants pass BuildKit validation.
[...]

PR #1783, opened 2026-07-22
Publish automatic CUDA 13 Beaker image

mnoukhov:
## Summary

- keep `open_instruct_auto` on the default CUDA 12 build and also publish it as `open_instruct_auto_cuda12`
- build CUDA 13 in a dedicated job after the merge-queue integration job succeeds
- publish the CUDA 13 build as `open_instruct_auto_cuda13` with a separate BuildKit cache
- smoke-check the CUDA 13 container without adding CUDA 13 GPU tests
- fix the merge-queue workflow's CUDA 12 image lookup and increase its timeout for the larger image

## Validation

- `make style`
- `make quality`
- workflow YAML parse
- full `uv run pytest` collection was attempted but interrupted after 4m41s while importing DeepSpeed/PyTorch with no tests collected; it also reported the existing missing system `libaio` dependency

Runs:

1. CUDA 12 GRPO integration test (exit 0): [Beaker](https://beaker.org/ex/01KY5PB5660VPN9EAWNRH27ZB7)

GPU_TESTS=bypass


undfined (approved):
lgtm assuming images work as expected!

farhatkevin (approved):
Approving — the workflow lookup fix here is the root cause of the `Beaker Experiment Launch` failures that have been Slack-alerting on every merge-queue entry since 2026-07-22.

Confirmed from the launch step of run [30129995242](https://github.com/allenai/open-instruct/actions/runs/30129995242): #1758 started building `open-instruct-integration-test-<branch>-cuda12`, while the workflow still looked up the unsuffixed name, so `docker image inspect` missed and the Beaker pull 404'd *after* the experiments had already launched successfully. Two other theories are ruled out by the same log — Beaker accepted the 109-character merge-queue image name without complaint (`Pushing open-instruct-integration-test-gh-readonly-queue-... as ... (01KYB3792J1520C7HDJGPZXA86)... Done.`), and the build did run rather than taking the skip path, so no truncation or namespace fix is needed.

The `timeout-minutes` 35 → 60 bump is worth having independently: that run spent roughly 9 minutes on the image upload alone.

One non-blocking note for a follow-up rather than this PR: `publish-cuda13-image` builds and uploads a second full ~24 GiB image on every merge-queue entry, and since the Dockerfile `COPY`s `open_instruct`/`scripts`/`configs`, the upper layers change on nearly every merge, so the full push happens each time rather than hitting cache. It's also published on a `docker run --rm` smoke check while the CUDA 12 alias is gated on the GRPO integration test. If nothing consumes `open_instruct_auto_cuda13` at commit freshness yet, a nightly `schedule` + `workflow_dispatch` cadence would keep the alias current for much less per-merge cost.

---

PR #1784, opened 2026-07-24
Harden DockerBackend with retries, OOM detection, and concurrency limits

farhatkevin:
Replacement for #1744 (originally #1732), moved to a branch in `allenai/open-instruct` because GitHub Actions stopped creating workflow runs for pushes to the fork branch, and hamishivi is no longer available to iterate on it. Rebased onto current `main`.

## Summary
- Add cross-process start/exec concurrency semaphores (`SWERL_DOCKER_START_CONCURRENCY` / `SWERL_DOCKER_EXEC_CONCURRENCY`) to avoid overwhelming the Docker daemon under many concurrent rollouts.
- Retry transient Docker exec `APIError`s (e.g. "database is locked") on the same container with exponential backoff.
- Detect OOM-killed containers and end the episode instead of continuing against a dead container.
- Add per-command `timeout` override to `run_command`, optional `docker_host` endpoint (e.g. Podman sockets), best-effort container-state diagnostics, and opt-in timing logs.
- Close containers kill-first (then stop) for faster, more reliable teardown.

The original raised `SandboxOOMError` so that "episodes terminate cleanly instead of looping", but nothing caught it. It fell through to the generic `except Exception` in `vllm_utils.py`, which scores the step 0 and keeps going *without* setting `rollout.done` — so the agent re-tripped the same dead container on every remaining step, which is the looping the change set out to remove. `_execute_bash` now catches it and returns `StepResult(done=True)`, which is what `vllm_utils` checks (`if step_result.done: rollout.done = True`) to end a rollout.

Restart-and-retry across a container restart is removed for a related reason. The container filesystem *is* the episode state — the agent's edits plus the wrapper script and persisted cwd written during `reset` — so re-running a command against a freshly created container silently executed it against a pristine filesystem (`/tmp/.sandbox_bash_wrapper.sh` does not even exist there) and produced a trajectory that looked valid but was not. Container loss now raises the new `SandboxContainerLostError`, also terminal for the episode.

Retrying transient exec errors on the *same* container is unchanged; that path loses no state.

## Test plan
- `uv run pytest tests/test_docker_backend.py tests/test_generic_sandbox.py` — 33 passed.
- Unit tests cover client selection, transient-error retry/backoff, retry exhaustion, non-retry of unknown errors (mocked Docker SDK, no daemon needed), plus new cases asserting OOM and lost-container both raise terminal errors without recreating the container, and that the environment converts them into `done=True` steps.
- `make style && make quality` clean.

GPU_TESTS=bypass


open_instruct/environments/backends.py
```diff
@@ -14,6 +19,84 @@
...
+    if value is None:
+        return default
+    try:
+        return float(value)
+    except ValueError:
+        logger.warning("Invalid float for %s=%r; using default %s", name, value, default)
+        return default
+
+
+class _FileSlotSemaphore:
```
undfined: I generally don't love the llm style of prefixing _ onto all these classes.

PR #1739, opened 2026-06-25
Add SWERLVanilluxSandboxEnv

hamishivi:
## Summary
- Add `SWERLVanilluxSandboxEnv` (`config_name="swerl_vanillux_sandbox"`), a **self-contained** bash-only sandbox RL environment that mirrors the offline mini-swe-agent "vanillux" solver harness:
  - a **persistent shell** (cwd/env preserved across `bash` calls via a wrapper script),
  - mini-swe-agent **head/tail observation truncation**,
  - `(exit_code=N)` suffixes on observations,
  - prompt templates vendored from mini-swe-agent v2.2.x in `vanillux_prompts.yaml`,
  - opt-in `append_turns_remaining` and `tool_call_format_error_feedback`.
- Submit-based evaluation via `echo COMPLETE_TASK_AND_SUBMIT_FINAL_OUTPUT`.

## Standalone
- Branches off `main` and does **not** depend on `swerl_sandbox.py` or any other sandbox-env work: it inlines its own `SUBMIT_MARKER` / `LAST_STEP_WARNING` / timing constants.
- Adds the only two small shared pieces it needs: `SandboxOOMError` in `backends.py` and the optional `RLEnvironment.get_tool_call_format_error_message` hook, plus `TOOL_REGISTRY` / `__init__` registration.
- Runs on the existing Docker backend.

## Test plan
- New `open_instruct/test_swerl_vanillux_sandbox.py` (12 tests, fake backends — no Docker): registration, bash-only tool surface, opt-in format-error message, instance/prompt rendering, head/tail truncation, submit-marker → verifier, exit-code suffix, and turns-remaining/submit-warning behavior. All pass locally.
- `ruff` + formatting clean.

GPU_TESTS=bypass

PR #1788, opened 2026-07-26
Add SWERLVanilluxSandboxEnv

farhatkevin:
Replacement for #1748 (originally #1739), moved to a branch in `allenai/open-instruct` because GitHub Actions stopped creating workflow runs for pushes to the fork branch, and hamishivi is no longer available to iterate on it.

**Based on #1784, not `main`** — see below. Retarget to `main` once #1784 merges.

## Summary
- Add `SWERLVanilluxSandboxEnv` (`config_name="swerl_vanillux_sandbox"`), a self-contained bash-only sandbox RL environment that mirrors the offline mini-swe-agent "vanillux" solver harness:
  - a **persistent shell** (cwd/env preserved across `bash` calls via a wrapper script),
  - mini-swe-agent **head/tail observation truncation**,
  - `(exit_code=N)` suffixes on observations,
  - prompt templates vendored from mini-swe-agent v2.2.x in `vanillux_prompts.yaml`,
  - opt-in `append_turns_remaining` and `tool_call_format_error_feedback`.
- Submit-based evaluation via `echo COMPLETE_TASK_AND_SUBMIT_FINAL_OUTPUT`.

`_run_tests` — the success path of *every* episode — could not run at all. Two independent breaks:

1. It calls `run_command(..., timeout=...)`. That parameter does not exist on `main`; it is added by #1784. On `main` this raises `TypeError: run_command() got an unexpected keyword argument 'timeout'`. Hence this PR targets #1784's branch.
2. It calls `self._backend.put_archive(...)`, which **no backend implements** — `DockerBackend` has only `start/run_command/write_file/read_file/close`. The name belongs to the docker SDK *container* object, used correctly inside `DockerBackend.write_file`. This PR adds `put_archive` to `DockerBackend` (delegating to the container) and a non-abstract base method so other backends keep instantiating and fail loudly only if something tries to upload an archive.

Either one means a submitting agent gets a `TypeError`/`AttributeError` that the generic handler in `vllm_utils` converts into an error observation with `reward 0.0` and no `done` — so **every task would have scored zero**, looking like a model that can't solve anything rather than a bug.

Neither was caught because the only test reaching `_run_tests` patches it out, and the fake backends declare their own `run_command`/`put_archive`, so they accept calls the real backend rejects. Added tests that execute `_run_tests` for real (upload → run → parse reward, including the clamp/default-zero path), plus contract tests asserting `timeout` and `put_archive` exist on the *actual* backend classes.

`Code Quality` was also failing on the original PR (6 `ty` errors — one of which was exactly the `timeout` call). Fixed by narrowing the `_image` reassignment to `DockerBackend` and annotating `_BASH_TOOL` as `dict[str, Any]`, matching `generic_sandbox.py`.

## Test plan
- `uv run pytest open_instruct/test_swerl_vanillux_sandbox.py tests/test_docker_backend.py tests/test_generic_sandbox.py` — 49 passed.
- `make style && make quality` clean (`ty` was failing before).

GPU_TESTS=bypass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 
 
 ### Added
-- Add `SWERLVanilluxSandboxEnv` (`swerl_vanillux_sandbox`), a self-contained bash-only sandbox RL environment mirroring the offline mini-swe-agent "vanillux" solver harness: a persistent shell, head/tail observation truncation, `(exit_code=N)` suffixes, vendored prompt templates, and opt-in turns-remaining / tool-call-format-error feedback (https://github.com/allenai/open-instruct/pull/PR_NUMBER).
+- Add `SWERLVanilluxSandboxEnv` (`swerl_vanillux_sandbox`), a self-contained bash-only sandbox RL environment mirroring the offline mini-swe-agent "vanillux" solver harness: a persistent shell, head/tail observation truncation, `(exit_code=N)` suffixes, vendored prompt templates, and opt-in turns-remaining / tool-call-format-error feedback (https://github.com/allenai/open-instruct/pull/1748).
 - Drop stale async rollout results whose generating policy is more than `async_steps` behind the trainer (`max_result_age_steps`), replenishing a fresh prompt and logging a `stale_results_dropped` metric (https://github.com/allenai/open-instruct/pull/1738).
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 
 ### Added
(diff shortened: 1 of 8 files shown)


farhatkevin:
Pushed d1a67ccc5, replacing the lock-directory mutex in `resolve_task_data_dir` with an `flock`.

The original guarded extraction with `os.mkdir(lock_dir)` and removed the directory in a `finally` block. A `finally` does not run when the process is SIGKILLed — Beaker preemption or the OOM reaper — so a dead holder left the lock behind and every other actor on the node then span forever in:

```python
while not os.path.isfile(complete_file):
    try: os.mkdir(lock_dir)
    except FileExistsError: time.sleep(1); continue
```

No timeout, no logging — an unbounded hang that presents as a stall rather than a failure, on a shared per-node path.

`fcntl.flock` is released by the kernel when the holder dies, so staleness cannot occur by construction. This also matches `_FileSlotSemaphore` in `backends.py`. Added a 30-minute deadline (`VANILLUX_EXTRACT_WAIT_TIMEOUT_S`) for the case where the holder is alive but stuck, once-a-minute logging while waiting, and a re-check of the completion marker so a waiter returns as soon as the extractor finishes rather than waiting to acquire the lock itself.

Four new tests cover extraction, reuse of a completed extraction (asserting `tar` is not re-invoked), the no-tarball path, and an abandoned lock file not blocking. Worth noting the abandoned-lock test would *hang* rather than fail against the old implementation, since `os.mkdir` on an existing path loops forever.

53 tests pass across the sandbox suite; `make style && make quality` clean.

_🤖 Posted by [Claude Code](https://claude.com/claude-code)_

PR #1789, opened 2026-07-26
Send Beaker debug jobs to ai2/jupiter

farhatkevin:
## Summary
`--cluster` is declared `nargs="+"` in [mason.py:113](https://github.com/allenai/open-instruct/blob/main/mason.py#L113), so **repeating the flag overwrites rather than appends** — argparse keeps only the last occurrence. Every one of these scripts listed several clusters but ran on just the last one:

| Script | Clusters listed | Actually used |
|---|---|---|
| `dpo_integration_test.sh` | 4 | `ai2/prior` |
| `grpo_integration_test.sh` | 2 | `ai2/saturn` |
| `tool_grpo.sh` | 3 | `ai2/saturn` |
| `single_gpu_grpo.sh` | 2 | `ai2/saturn` |
| `single_gpu_on_beaker.sh` | 2 | `ai2/saturn` |

Each now names `ai2/jupiter` alone, so what the script says is what runs.

## Why it matters
This is silent — the script reads as if several clusters are eligible, and the job spec ends up with `constraints: {"cluster": ["ai2/saturn"]}`. It bites in practice: a 1-GPU debug job pinned to saturn sat unscheduled for 30+ minutes with

> Job cannot be scheduled because it would exceed the workspace slot limit.

while jupiter had capacity. The `ai2/open-instruct-dev` workspace has a per-cluster slot limit, and the Beaker API does not surface that reason — only the UI does — so it presents as a job that mysteriously never starts.

`grpo_integration_test.sh` is the script the merge-queue Beaker workflow launches, so CI was pinned to saturn as well.

## Notes
- The alternative fix is the space-separated form (`--cluster ai2/jupiter ai2/saturn`), which would keep every listed cluster eligible. Going to jupiter alone instead, since batch jobs should run there.
- **This moves the DPO integration test off `ai2/prior`.**
- All five scripts are `--num_nodes 1`, and `ai2/jupiter` is in `INTERCONNECT_CLUSTERS`, so the multi-node interconnect check in `mason.py` is unaffected either way.

## Test plan
- `pre-commit run shellcheck` on all five — passed.
- `git diff --check` clean.
- Verified against a live launch: the same script with jupiter alone produces `constraints: {"cluster": ["ai2/jupiter"]}` and schedules.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,4 +23,5 @@ All notable changes to this project will be documented in this file.
 ### Removed
 
 ### Fixed
+- Send Beaker debug/integration jobs to `ai2/jupiter`: `--cluster` is `nargs="+"` in `mason.py`, so repeating the flag overwrote instead of appending and five scripts silently ran on only their last-listed cluster (https://github.com/allenai/open-instruct/pull/1789).
 - Track the CUDA 12 image suffix in the merge-queue Beaker workflow and allow enough time for the larger image build and upload (https://github.com/allenai/open-instruct/pull/1783).
--- a/scripts/train/debug/dpo_integration_test.sh
+++ b/scripts/train/debug/dpo_integration_test.sh
@@ -2,10 +2,7 @@
 BEAKER_IMAGE="${1:-nathanl/open_instruct_auto}"
 
 uv run python mason.py \
-    --cluster ai2/neptune \
-    --cluster ai2/saturn \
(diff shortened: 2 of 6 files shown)


undfined (commented):
Is this a compat issue or something else? Dropping prior cluster makes sense but should we keep the other dev clusters saturn/neptune?

PR #1790, opened 2026-07-27
WIP: Add DPPO via the unified rho-divergence masking paradigm + refactor APIs

undfined:
Alternative to #1745 that implements DPPO through the shared GRPO ratio, correction, and masking machinery instead of adding a trainer-specific loss path. This extends the original design from #1755 while making the final configuration more explicit, composable, numerically robust, and consistent across the DeepSpeed and OLMo-core trainers.

## Summary

The policy loss is now described by three independent parts:

1. **Policy-gradient objective** — `--loss_fn`
2. **Policy-ratio denominator and rollout correction**
3. **Token-drop mask statistic and behavior**

Both `grpo_fast.py` and `grpo.py` use the same per-token score-function objective:

```text
loss_t = -M_t · stopgrad(c_t · r_t · A_t) · log π_θ(a_t | s_t)
r_t = π_θ / d
```

Here:

- `M_t` is the final structural update mask;
- `c_t` is the optional rollout-to-old-policy correction;
- `r_t` is the configured policy ratio;
- `A_t` is the token advantage; and
- `d` is either the old trainer policy `π_old` or the rollout policy `μ`.

### Policy-gradient objectives

`--loss_fn` supports:

- `dapo`: directional lower/upper PPO clipping, implemented as a structural mask;
- `cispo`: caps the detached ratio coefficient at the configured upper bound;
- `dppo`: the unclipped DPPO objective from [Eq. 11](https://arxiv.org/abs/2602.04879), whose trust region is enforced entirely by the divergence mask from Eq. 12.

DAPO and CISPO retain their PPO-compatible defaults. DPPO uses the direct, unbounded rollout-policy ratio `π_θ / μ`.

### Ratio denominator and rollout correction

The policy-ratio denominator is explicit:

- `policy_ratio_denominator=old_policy` uses `r = π_θ / π_old` and retains fixed old-policy log probabilities for the duration of the PPO update.
- `policy_ratio_denominator=rollout_policy` uses `r = π_θ / μ` directly from vLLM log probabilities and skips old-logprob computation and storage.

The rollout correction is configured separately:

- `rollout_importance_correction=clipped` applies the configured clamp to `π_old / μ`.
- `rollout_importance_correction=none` uses unit correction.

A direct rollout-policy ratio cannot also use the rollout correction because that would count the rollout denominator twice.

The default behavior remains:

```text
r = π_θ / π_old
c = clip(π_old / μ)
```

This preserves the previous PPO-compatible DAPO/CISPO update rather than silently changing the policy-ratio denominator to the rollout policy.

### Property-based rho masking

The original IcePop/VACO/DPPO masking concepts remain, but the final API expresses them through independent properties rather than named algorithm switches:

- `rho_mask_metric`: `none`, `ratio`, `tv`, or `kl`
- `rho_mask_source`: `old_policy` or `current_policy`
- `rho_mask_level`: `token` or `sequence`
- `rho_mask_direction`: `symmetric` or `increase_only`

The standard compositions are:

- **IcePop:** `ratio + old_policy + token + symmetric`
- **VACO:** `tv|kl + old_policy + sequence + increase_only`
- **DPPO:** `tv|kl + current_policy + token + increase_only`

VACO and DPPO share the same binary Bernoulli TV/KL implementation. They differ only in:

- the policy compared with the rollout policy `μ`: `π_old` for VACO and `π_θ` for DPPO; and
- the aggregation level: sequence for VACO and token for DPPO.

Both use directional filtering. A threshold violation is dropped only when:

```text
sign(log(π / μ)) · A > 0
```

This means updates that move farther from the rollout policy are filtered, while updates moving back toward `μ` remain eligible.

The binary divergence collapses the categorical distribution into a Bernoulli distribution over `{sampled token, all other tokens}` using selected-token log probabilities only. This provides a memory-efficient divergence estimate without another model forward pass.

### DPPO configuration

DPPO configurations are validated to require:

```bash
--loss_fn dppo \
--policy_ratio_denominator rollout_policy \
--rollout_importance_correction none \
--rho_mask_metric tv \
--rho_mask_source current_policy \
--rho_mask_level token \
--rho_mask_direction increase_only \
--rho_mask_upper_bound 0.1
```

`rho_mask_metric=kl` is also supported.

The retained `π_θ / μ` ratio is intentionally unbounded. The divergence mask is DPPO’s trust region; correction clamp settings do not silently clamp the DPPO coefficient.

## Shared loss implementation

`compute_grpo_loss` now owns:

- policy-ratio construction;
- rollout importance correction;
- ratio or TV/KL filtering;
- DAPO/CISPO/DPPO objective behavior;
- reference-KL masking; and
- loss-related numerical validation.

It returns a `GRPOLossOutput` containing:

- `pg_loss`
- `clipfrac`
- `kl`
- the sanitized configured `ratio`
- the full `RhoCorrection`
- the effective `kl_mask`

This replaces the duplicated ratio → rho correction → loss → metric plumbing in `grpo_fast.py` and `olmo_core_train_modules.py` with a single shared call.

Old-policy log probabilities are computed and cached only when `policy_ratio_denominator=old_policy`. Direct rollout-policy configurations avoid that extra forward pass and storage.

## Structural masking and numerical safety

All response, validity, clipping, divergence, and optional reference-KL masks are structural:

- masked selected-token log probabilities are removed from the autograd graph with `torch.where`;
- excluded tokens receive exactly zero direct policy gradient;
- masking removes gradients through all logits associated with an excluded selected token;
- padding and invalid loss metadata are excluded consistently;
- non-finite or positive current-policy log probabilities at response positions fail fast;
- invalid behavior, old-policy, advantage, or reference metadata cannot leak non-finite values into the loss;
- policy and correction ratios are computed in fp32;
- binary TV/KL uses float64 near probability boundaries to avoid cancellation;
- histogram filtering happens after transfer to CPU, avoiding dynamic-shape accelerator indexing.

Ratio overflow is checked only after the final policy mask is known:

- an overflowed ratio on a retained token fails fast;
- an overflowed token safely removed by the configured mask does not terminate the run;
- overflow incidence remains observable through `val/rho_overflow_frac`.

The loss path also reduces accelerator-to-host synchronization by combining error counts and deferring histogram filtering until tensors reach CPU.

## Reference-KL behavior

Reference-KL masking is independently configurable:

- `mask_reference_kl_with_policy=False` preserves legacy behavior: a token excluded from the policy objective may still receive reference-KL gradient when its current and reference log probabilities are valid.
- `mask_reference_kl_with_policy=True` applies the final policy mask to the KL objective as well.

KL estimator 3 always excludes tokens without a finite configured policy ratio because that estimator multiplies by the ratio directly.

Invalid reference-policy metadata removes only the affected KL gradient unless the policy path is independently masked.

## Tool-use compatibility

Tool output tokens are assigned synthetic vLLM log probabilities, so `grpo_fast.py` requires `mask_tool_use=True` whenever vLLM log probabilities affect the policy loss through:

- `policy_ratio_denominator=rollout_policy`;
- `rollout_importance_correction=clipped`; or
- an active rho mask.

Pure PPO configurations remain valid with `mask_tool_use=False` when they use:

```bash
--policy_ratio_denominator old_policy \
--rollout_importance_correction none \
--rho_mask_metric none
```

A configured rho mask whose lower and upper bounds are both disabled also does not make the policy loss depend on vLLM log probabilities.

This preserves the behavior from `main` rather than unconditionally requiring tool masking for every tool-enabled run.

## Observability

The branch adds or clarifies the following diagnostics:

- `val/ratio`: mean finite configured ratio, either `π_θ / π_old` or `π_θ / μ`
- `val/ratio_var`: variance of the configured policy ratio
- `val/rho_hist`: finite raw rho or direct rollout-policy ratio distribution
- `val/rho_weight`: effective retained correction or policy coefficient
- `val/rho_clipfrac`: fraction affected by rollout-correction clamping
- `val/rho_drop_frac`: fraction removed by the configured drop mask
- `val/rho_drop_low_frac`: fraction below the configured lower bound
- `val/rho_drop_high_frac`: fraction above the configured upper bound
- `val/rho_divergence`: TV/KL statistic used for filtering
- `val/rho_overflow_frac`: valid tokens whose raw ratio overflowed
- `policy/clipfrac_avg`: DAPO structural clipping or CISPO coefficient capping

`val/ratio` excludes raw overflowed ratios rather than averaging their sanitized zero values.

The complete policy-loss configuration is logged at startup and retained in W&B metadata, including:

- objective;
- policy-ratio denominator;
- rollout correction;
- mask metric, source, level, direction, and bounds; and
- reference-KL masking behavior.

## Migration guidance

The defaults preserve the prior DAPO/CISPO PPO behavior and do not silently change the clip baseline to `μ`.

The former configuration is replaced by explicit ratio, correction, and mask properties:

- `use_vllm_logprobs`
- `use_rho_correction`
- `rho_mask_tv_divergence`
- the intermediate named `rho_divergence_algo` / `rho_divergence_type` configuration

To replace `use_rho_correction=False` without changing its old meaning, use:

```bash
--policy_ratio_denominator old_policy \
--rollout_importance_correction none \
--rho_mask_metric none
```

This remains ordinary PPO weighting by `π_θ / π_old`; it does not expose an unclamped `π_θ / μ` coefficient.

### Sequence-level masking

`rho_mask_level=sequence` now uses the broadcast sequence statistic only for the keep/drop decision.

Retained tokens continue to use their per-token clipped `π_old / μ` correction weights. Previously, the broadcast sequence-mean ratio was also used as the correction clamp base and loss weight.

### VACO threshold scale

VACO now thresholds probability-space binary TV or KL rather than the old ratio-space sequence mean of `|ρ - 1|`.

Previous thresholds such as `2.0` are therefore not directly transferable. New thresholds should be selected in the divergence’s scale and validated using `val/rho_divergence` and `val/rho_drop_frac`.

### Loss dashboard scale

`loss/policy_avg` and `loss/total_avg` are not numerically comparable with older runs.

The score-function representation includes `log π_θ` in the reported policy-loss value, and structurally DAPO-clipped tokens now report zero instead of the clipped surrogate value. The policy gradients remain equivalent.

## Qwen3-4B DPPO recipes

Four single-node, eight-GPU Qwen3-4B math recipes are included:

- `scripts/train/qwen/qwen3_4b_dppo_math.sh`
- `scripts/train/qwen/qwen3_4b_dppo_kl_math.sh`
- `scripts/train/qwen/qwen3_4b_dppo_math_oc.sh`
- `scripts/train/qwen/qwen3_4b_dppo_kl_math_oc.sh`

They cover:

- DeepSpeed and OLMo-core trainers;
- TV and KL divergence masks;
- configurable divergence thresholds;
- default TV threshold `0.2`;
- default KL threshold `0.05`;
- descriptive run names containing the divergence, threshold, and trainer;
- Qwen3-4B-Base on the filtered DAPO Math dataset;
- active asynchronous sampling with in-flight updates;
- long 8192-token responses;
- AIME 2025 and BRUMO evaluation every 100 steps;
- resumable checkpoint-state directories;
- retention of only the latest checkpoint;
- Python-based Mason and trainer entrypoints; and
- `--no_auto_dataset_cache` for launch compatibility.

## Test plan

The expanded CPU tests cover:

- PPO-compatible default ratios and rollout correction;
- direct rollout-policy ratios;
- conditional old-logprob computation and caching;
- DAPO, CISPO, and DPPO coefficient and gradient behavior;
- binary TV/KL definitions and near-boundary numerical stability;
- IcePop-, VACO-, and DPPO-style masking;
- directional filtering;
- token- and sequence-level aggregation;
- structural gradient removal;
- padding and invalid metadata;
- safely masked versus retained ratio overflows;
- independent and shared reference-KL masks;
- metric correctness, including overflow-safe `val/ratio`;
- configuration validation;
- pure-PPO tool-use compatibility; and
- tool masking when vLLM log probabilities affect the policy loss.

Validation:

- `open_instruct/test_olmo_core_train_modules.py`: 65 passed
- `make style && make quality`: passed
- `uv run mkdocs build`: passed

GPU_TESTS=bypass


open_instruct/grpo_utils.py
```diff
@@ -392,150 +483,389 @@ def _sequence_level_mean(values: torch.Tensor, response_mask: torch.Tensor) -> t
...
+        ("advantages", advantages),
+        ("response_mask", response_mask),
+    ):
+        if tensor.shape != expected_shape:
+            raise ValueError(f"{name} shape {tensor.shape} must match new_logprobs shape {expected_shape}.")
+    if ref_logprobs is not None and ref_logprobs.shape != expected_shape:
+        raise ValueError(f"ref_logprobs shape {ref_logprobs.shape} must match new_logprobs shape {expected_shape}.")
+    if old_logprobs is not None and old_logprobs.shape != expected_shape:
+        raise ValueError(f"old_logprobs shape {old_logprobs.shape} must match new_logprobs shape {expected_shape}.")
+
```
mnoukhov, line 799: maybe add these into the loop and just do `is not None` for all of them

open_instruct/grpo_utils.py
```diff
@@ -322,19 +330,93 @@ def __post_init__(self):
...
                 )
-            if self.rho_clamp_upper_bound > 0.0 and self.rho_clamp_upper_bound <= 1.0:
-                raise ValueError(f"rho_clamp_upper_bound must be > 1 when set, got {self.rho_clamp_upper_bound}.")
+            if self.rho_mask_upper_bound < 0.0 or not math.isfinite(self.rho_mask_upper_bound):
+                raise ValueError(
+                    f"rho_mask_upper_bound must be finite and >= 0 for {self.rho_mask_metric}, "
+                    f"got {self.rho_mask_upper_bound}."
+                )
+        if self.loss_fn == GRPOLossType.dppo:
+            required = {
```
mnoukhov, line 387: This is kind of a strange paradigm. We shouldn't have to set DPPO and also correctly set things that define DPPO.

mnoukhov (changes requested):
VACO is not DPPO with pi_old. It is exactly DPPO but with a token+sequence-level consideration of TV. This is honestly, not your fault. The paper is incredibly dense and I've been trying to convince the main author to put up a clean implementation somewhere.

Overall, this feels like a really big PR and I think it is doing maybe too many things. I would maybe prefer a first PR that deprecated pi_old and does the refactor. A DPPO PR that builds on top of that would then be pretty simple. 

For the rho refactor, I'm still a little confused about rho mask direction. This seems like a DPPO specific thing. Maybe its worth going back to the having a loss_fn and just upper and lower bound then doing more monolithic-type algorithms that leverage modular functions (like getting a divergence).  

I also think we should deprecate `grpo_fast.py` at some point, to make things simpler for ourselves going forward.

PR #1792, opened 2026-07-28
Run short Beaker debug jobs preemptible

farhatkevin:
## Summary
Add `--min_runtime` support to `mason.py` and set a 30-minute minimum runtime on the four short debug/test launch scripts that request `--priority urgent`.

`--preemptible` remains available as a compatibility alias for `--min_runtime 0`. The Beaker SDK is bumped to a version that supports `min_runtime` directly.

## Why
These scripts previously asked for non-preemptible urgent slots. Beaker then attached an 8-hour `minRuntime` guarantee to jobs whose own `--timeout` is only 15–30 minutes—a much harder allocation than they need.

Measured on `ai2/jupiter`, with an otherwise identical 1-GPU job (same cluster, image, and priority):

| Config | Result |
|---|---|
| non-preemptible, `minRuntime: 28800000000000` (8h) | queued **24+ hours**, never scheduled, no node assigned |
| immediately preemptible, `minRuntime: 0` | **scheduled in 49 seconds**, ran to completion |

This PR uses a 30-minute minimum runtime as a middle ground: the scheduler only needs to guarantee a short window, while the smoke test is protected from preemption during that window.

## Scope

| Script | `--timeout` | `--min_runtime` |
|---|---:|---:|
| `single_gpu_on_beaker.sh` | 15m | 30m |
| `single_gpu_grpo.sh` | 15m | 30m |
| `envs/guess_number_beaker.sh` | 30m | 30m |
| `tools/tool_regression_beaker.sh` | 30m | 30m |

Longer multi-node scripts are unchanged.

## Test plan

- `uv lock --check`
- `make style && make quality`
- `uv run pytest test_mason.py` — 14 passed
- `pre-commit run shellcheck` on all four scripts
- `bash -n` on all four scripts
- The earlier `minRuntime: 0` scheduling comparison was exercised end-to-end: [01KYN5PPJNJ99W1GPHEP1VPWYK](https://beaker.org/ex/01KYN5PPJNJ99W1GPHEP1VPWYK) scheduled in 49 seconds and finished training with exit code 0.


undfined (approved):
Should we also set min runtime to 30 min for these?

PR #1796, opened 2026-08-04
Validate effective_steps before building the LR scheduler

abhishekraok:
Follow-up to #1795 per @farhatkevin's review: that PR fixed `oc_sft_multinode.sh`, but any undersized dataset still floored `num_training_steps` to 0 and surfaced a bare `ZeroDivisionError` from olmo-core's scheduler.

Raise a `ValueError` before the scheduler is built:

```
Computed 0 training steps from 26 packed instances // global batch 64 * 2 epochs.
Use more data, or lower --gradient_accumulation_steps / --per_device_train_batch_size.
```

Only fires when `effective_steps < 1`, which previously always crashed, so normal training is unaffected.

## Verification

Original failure, 100-example mixer at seq 4096 (26 packed instances): `ZeroDivisionError` — [01KZ4BQDCWGQ98ZFWE4P7GKRM1](https://beaker.org/ex/01KZ4BQDCWGQ98ZFWE4P7GKRM1).

Same config against this branch: [01KZ7GY7E0K863P6YJKMDRGAF3](https://beaker.org/ex/01KZ7GY7E0K863P6YJKMDRGAF3) — queued at time of writing, will update with the result.

Evidence was built at `b0282a6f4`; the only change since is shortening the error string.

GPU_TESTS=[01KZ7CWKYZZXBCY476C7DVHE0J](https://beaker.org/ex/01KZ7CWKYZZXBCY476C7DVHE0J)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,5 +23,6 @@ All notable changes to this project will be documented in this file.
 ### Removed
 
 ### Fixed
+- Raise a `ValueError` naming the packed instance count, global batch size, and epoch count when `olmo_core_finetune.py` computes fewer than one training step, instead of letting an undersized dataset surface as a bare `ZeroDivisionError` from olmo-core's LR scheduler (https://github.com/allenai/open-instruct/pull/1796).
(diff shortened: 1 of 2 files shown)


farhatkevin (approved):
LGTM!

PR #1801, opened 2026-08-05
Include all cache-key args in the tokenization remediation command

abhishekraok:
On a cache miss, `olmo_core_finetune.py` raises `FileNotFoundError` with a copy-pasteable tokenization command — but it omitted three args that feed the cache key:

- `--seed` and `--max_seq_length` → `_seed_cache_suffix`
- `--chat_template_name` and `--transform_fn` → `compute_config_hash`

So following the printed command built a cache under a different key, and the re-launched training job failed with the identical error. Reachable straight from the tracked debug scripts: `oc_sft.sh` trains with `--seed 123`, while the printed command defaulted to 42.

Rendered output now:

```
      -- uv run python open_instruct/olmo_core_finetune.py \
      --model_name_or_path allenai/OLMo-2-0425-1B \
      --tokenizer_name_or_path allenai/OLMo-2-0425-1B \
      --max_seq_length 4096 \
      --mixer_list allenai/tulu-3-sft-personas-algebra 1000 \
      --seed 123 \
      --chat_template_name tulu \
      --transform_fn sft_tulu_tokenize_and_truncate_v1 sft_tulu_filter_v1 \
      --local_cache_dir /weka/... \
      --cache_dataset_only
```

`--chat_template_name` and `--transform_fn` are emitted only when set, so the default case is unchanged apart from `--seed`.

## Verification

Message rendered from the source block with stubbed args — single-backslash continuations, all cache-key args present. Not GPU-exercised: this is a string change on an error path, covered by style/quality and unit tests.

Observed twice while onboarding: [01KZ4BQDCWGQ98ZFWE4P7GKRM1](https://beaker.org/ex/01KZ4BQDCWGQ98ZFWE4P7GKRM1) hit the miss, the suggested command produced cache `22a47a9221-681da6e5` (seed 42), and training still wanted `22a47a9221-abb4ee88` (seed 123).

## Note

`--local_cache_dir` is still printed, but `setup_tokenizer_and_cache` overwrites it unconditionally inside Beaker jobs ([olmo_core_utils.py:467](https://github.com/allenai/open-instruct/blob/main/open_instruct/olmo_core_utils.py#L467)), so the flag has no effect there. Left as-is since it's harmless for the copy-paste flow, but worth knowing that `oc_sft_olmo3_7b_full.sh` passes a value that is ignored.

GPU_TESTS=bypass


abhishekraok:
Good catch — confirmed and fixed in ee3a8b4f0.

Verified the hash does change with the split: `['train']` → `044f53994b`, `['train[:50%]']` → `3ae2407b8d`. Same defect class as the other three args, one I missed.

abhishekraok:
Pushed 2339dd2cf — replaced the hand-listed tokenizer flags with derivation from `TokenizerConfig`.

Prompted by realising `--add_bos` was also missing, and it is *mandatory* for OLMo models. Verified it changes the hash: `add_bos=True` → `8886593a78`, `False` → `efa228411a`. Seven other hashed fields (`tokenizer_revision`, `sft_messages_key`, …) were unprinted too.

Hand-enumeration had now missed three args across three rounds (`--seed`, `--mixer_list_splits` via @chatgpt-codex-connector, `--add_bos`), so the list is now derived: every non-default `TokenizerConfig` field is emitted, with an explicit skip-set for the three that aren't CLI args. A newly added hashed field updates the printed command automatically.

abhishekraok:
Narrowed back in 87bb8000f — dropped the config-introspection helper for a plain flag list.

Rationale: the SFT wrapper is expected to move back to olmo-core, so drift-resistance in a file slated for removal isn't worth the surface area. All the args are still there, including `--add_bos`.

Net diff is now 24 lines (18 insertions) in `olmo_core_finetune.py` plus the CHANGELOG line.

abhishekraok:
Related: #1815 covers the mirror case on this same cache key. This PR fixes a false *miss* — the remediation command omitted key args, so following it re-tokenized under a different key and failed identically. #1815 is a false *hit*: because the key has no source-code component, a change to label derivation (#1806) leaves `_numpy_dir_is_populated` returning True, and the run silently reuses labels built by the older code.

Hit while validating #1806: a full-mixture run finished in ~40 s having processed 0 rows, which read as "0 dropped rows" — an apparent improvement on the previous 15. Forcing a different `--seed` was the only way to get a real measurement.

Nothing to change here; the enumeration of cache-key args in this PR's description is what made the other case diagnosable.

undfined (approved):
Makes sense! I've never been a fan of these chained conditional blocks that stuff arguments together. This one is getting to be a bit complex so maybe warrants a utility method soon.

PR #1807, opened 2026-08-07
Document how to diagnose a queued Beaker job

abhishekraok:
Two lines in `AGENTS.md` (which `CLAUDE.md` symlinks to) about diagnosing Beaker jobs that don't schedule.

### Why

A 1x8 GPU job queued 9.5h, and a relaunch on a different cluster queued 2.5h more. Both had the same reason, which `beaker job events` states directly:

```
excluded by slot limit (workspace "ai2/open-instruct-dev" using 64/64
allowed unallocated slots, the job requires at least 8 slots)
```

The limit is workspace-wide, so it blocks every cluster at once and switching clusters cannot help. Inferring a cause from the cluster docs instead produced a plausible, documented, wrong answer.

The second line covers a related trap: a preempted job that Beaker retries leaves the experiment holding several jobs, so reading `jobs[0]` makes a successful retry look like a failure.

### Scope

Docs only.

GPU_TESTS=bypass


farhatkevin (approved):
Great! If this is working well, we should add more info about different clusters and expected workflow in the future.

abhishekraok:
Yeah, we should ask the agents to write down any missing info. This would have saved my agent a few hours.

PR #1808, opened 2026-08-08
Fix N-fold inflation of num_step_tokens, TPS, and MFU metrics in grpo_fast

farhatkevin:
## Summary

`DataPreparationActor.get_data` returns the **same global metrics dict** to every rank (`{"batch": prepared_data[step][rank], "metrics": self.metrics[step]}` — only `batch` is sharded), but `one_training_step` in `grpo_fast.py` concatenated `batch/prompt_lengths` and `batch/response_lengths` across all sp-leader ranks as if each rank held a distinct local shard. With `sequence_parallel_size=1` and N learner ranks, every rank passes the sp-leader filter, so every sample was counted N times.

Inflated metrics:
- `val/num_step_tokens` (×N)
- `learner_tokens_per_second_step` (×N)
- `learner_mfu` (×N — the inflation coincidentally cancels `num_training_gpus` only when `sequence_parallel_size == 1`)
- `actor_mfu` (×N — no cancelling term at all)

Verified empirically on Beaker run [01KZB01ZSXCGT3MVMGNGGC8Y7N](https://beaker.org/ex/01KZB01ZSXCGT3MVMGNGGC8Y7N) (wandb `923rvrm2`): reported `num_step_tokens` 1.143e6 vs. true global 285,750 (128 samples × 2,232 avg tokens); reported `learner_mfu` 44.3 vs. corrected ~11.1.

**Note: historical grpo_fast MFU/TPS numbers were inflated by the number of learner ranks.** The OLMo-core path (`grpo_callbacks.py` `StepTimingCallback` + `olmo_core_train_modules.py`) all-reduces token counts exactly once and was already correct.

## Fix

Read the per-sample length lists once from a single rank via a new `utils.get_global_batch_lengths` helper instead of concatenating across ranks. This matches the existing pass-through of the other array metrics, which already reads `array_metrics[0]` with the comment "these are the same across workers".

## Test

`TestGetGlobalBatchLengths` in `test_utils.py` builds per-rank array metrics for a 4-rank config (identical global dicts, as production produces) and asserts the extracted lengths and token count equal the true global batch count, not N× it.

GPU_TESTS=[01KZFDV12Q9FAGAHJW68K19RWH](https://beaker.org/ex/01KZFDV12Q9FAGAHJW68K19RWH)


abhishekraok:
I thought we were deprecating RL on open-instruct

PR #1809, opened 2026-08-10
Fix the olmo-core to HF checkpoint converter

abhishekraok:
`scripts/train/convert_olmo_core_to_hf.py` could not run at all.

1. `torch.distributed.checkpoint.state_dict.load_state_dict` no longer exists — immediate `AttributeError` on torch 2.10.
2. Switching to `torch.distributed.checkpoint.load` then fails with `'_StorageInfo' object has no attribute 'transform_descriptors'`. olmo-core writes checkpoints with its own storage layout, which torch's generic DCP reader cannot read.

Fixed by loading through `olmo_core.distributed.checkpoint.load_model_and_optim_state`, the counterpart to the save path that produced the checkpoint. It loads in place, so the state dict is read back off the model. `torch` is no longer referenced, so the import is dropped.

This path matters because olmo-eval's `olmo_core` provider also rejects open-instruct checkpoints (`config.json missing required field 'tokenizer'`), making HF conversion the only way to evaluate them.

Verified end to end: converted five Olmo-3-7B olmo-core SFT checkpoints and evaluated each on IFBench and GSM8K.

GPU_TESTS=bypass

https://claude.ai/code/session_01EfGd3gqhva8Stpgf4os7sn

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ All notable changes to this project will be documented in this file.
 ### Removed
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


abhishekraok:
Fixed in 
01f17a0f4 — switched to `from olmo_core.distributed import checkpoint as olmo_core_checkpoint` per AGENTS.md. Good catch.

saumyamalik (approved):
Looks good! Thanks — nice fix

PR #1810, opened 2026-08-10
Stop the forced ephemeral_save_interval default rejecting small checkpoint intervals

abhishekraok:
`olmo_core_finetune.py` forces `ephemeral_save_interval=500` via `parser.set_defaults`. Two consequences:

- **Any run with `--checkpointing_steps` below 500 is rejected at startup**, by olmo-core's `'ephemeral_save_interval' must be less than 'save_interval'` check. A 345-step interval cannot be used at all.
- **Ephemeral checkpoints cannot be turned off.** Omitting the flag yields 500, not `None`, so there is no way to express "off".

A non-positive value now maps to `None`, matching the existing `-1` convention for `max_checkpoints` in the same function.

## What this is not

An earlier version of this PR framed the change as a fix for `RuntimeError: Application timeout caused pair closure`, seen when a 7B SFT run died at step 401 while deleting an old checkpoint on Weka.

That framing was wrong, and @TimDettmers diagnosed the real defect: checkpoint cleanup is not fully asynchronous. `_remove_checkpoint()` submits `clear_directory()` to the bookkeeping executor, but `save_checkpoint()` / `save_checkpoint_async()` then call `_join_bookkeeping_ops()` before the next save, and that join includes the outstanding deletion future — so the async delete becomes blocking, filesystem rank 0 stalls, and the other ranks time out waiting for it. The fix belongs upstream: https://github.com/allenai/OLMo-core/pull/819.

Disabling ephemeral checkpoints avoids that path but costs rolling recovery. That trade is real — the run above resumed from its step-400 *ephemeral* checkpoint, and with these disabled it would have fallen back to step 345.

So this PR stands on the argument-parsing bug alone, and is complementary to #819 rather than a substitute for it.

## Note

#1811 passes `--ephemeral_save_interval -1` and needs this merged first.

https://claude.ai/code/session_01EfGd3gqhva8Stpgf4os7sn


TimDettmers:
> “The overall implementation looks good to me, but it is attacking the problem from the wrong angle.”

I think the root problem is that the checkpoint cleanup pipeline is not fully asynchronous. OLMo-core does begin the process asynchronously: `_remove_checkpoint()` first removes the metadata file to invalidate the old checkpoint and then submits `clear_directory()` to the bookkeeping executor. However, both `save_checkpoint()` and `save_checkpoint_async()` later call `_join_bookkeeping_ops()` before writing the next checkpoint. That join includes the outstanding deletion future, so an asynchronous delete becomes blocking at the next save.

On Weka, deleting a large checkpoint tree can be metadata-heavy and slow. While filesystem rank 0 is waiting for that deletion, the other ranks can enter the distributed checkpoint operation and wait for rank 0, eventually producing the process-group timeout described here. Disabling ephemeral checkpoints avoids that execution path, but it also gives up the rolling recovery checkpoints and works around the synchronization issue rather than fixing it.

I have here a draft of the implementation, how it should be done: https://github.com/allenai/OLMo-core/pull/819. It distinguishes checkpoint-dependent bookkeeping from independent cleanup. A checkpoint save still waits for operations needed to produce consistent state, but it does not wait for deletion of an already-invalidated, unrelated checkpoint directory. Cleanup remains tracked for error propagation and is still drained during graceful shutdown.

I do not have time to work on this further, but maybe you can take it and improve it from there.

abhishekraok:
Thanks — that's a better diagnosis than mine, and you're right that this was attacking the trigger rather than the defect. The `_join_bookkeeping_ops()` join pulling in the deletion future is the part I missed; I had it as "deleting a big tree on Weka is slow" without explaining why that blocked training.

Your point about losing rolling recovery is concrete for us too: the run this came from resumed from its step-400 *ephemeral* checkpoint. With these disabled it would have fallen back to the permanent step 345 and redone 55 steps.

I've reframed this PR so it no longer claims to fix the timeout. What's left is an argument-parsing bug that survives #819: the forced default of 500 means any run with `--checkpointing_steps` below 500 is rejected at startup, and there's no way to express "off" since omitting the flag yields 500 rather than `None`. Once [#819 lands](https://github.com/allenai/OLMo-core/pull/819), `-1` becomes a real choice instead of a workaround people are pushed into.

open_instruct/olmo_core_utils.py
```diff
@@ -249,10 +249,13 @@ def build_checkpointer_callback(
...
     ``max_checkpoints`` accepts the open-instruct convention where ``-1`` means
     unlimited.  Negative values are mapped to ``None`` (keep all).
+
+    ``ephemeral_save_interval`` follows the same convention: a non-positive value
+    disables ephemeral checkpoints.
     """
     return CheckpointerCallback(
         save_interval=checkpointing_steps,
-        ephemeral_save_interval=ephemeral_save_interval,
+        ephemeral_save_interval=ephemeral_save_interval if (ephemeral_save_interval or 0) > 0 else None,
```
undfined, line 258: Can we normalize this earlier so we don't have so much conditional logic here? Maybe the cli arg parser handles it? If it's too annoying you can disregard but it feels unsatisfying to encode the logic for what arguments mean so far downstream.

undfined (commented):
Makes sense. I added one nit about where we actually encode what the arguments from the cli mean but otherwise looks good.

abhishekraok:
build_checkpointer_callback is the one place open-instruct conventions get translated into olmo-core kwargs — max_checkpoints already does -1 → None on the next line. So I think it's better to keep both the mappings in the same place.

undfined (approved):

PR #1811, opened 2026-08-10
Add a 1-node Olmo-3-7B SFT script

abhishekraok:
Adds `scripts/train/debug/oc_sft_olmo3_7b_1node.sh`: Olmo-3-7B SFT over the full Dolci-Instruct-SFT mixture, 1 epoch, on a single 8-GPU node.

Derived from the existing `oc_sft_olmo3_7b_full.sh` (4 nodes, 2 epochs), keeping the global batch identical at 1,048,576 tokens — `grad_accum` goes 2 → 8 and `num_epochs` 2 → 1, with `cp_degree`, `cp_strategy`, `activation_memory_budget` and `attn_implementation` unchanged.

The script header records the operational findings that produced it, including why checkpoint deletion has to stay off and the expected step count (~1316) as a pre-flight check against the batch math.

### Depends on #1810

The script sets `--ephemeral_save_interval -1` to stop olmo-core deleting checkpoints mid-run. That value is only honoured with the change in #1810, so this should merge after it.

GPU_TESTS=bypass


saumyamalik (approved):
lgtm!

PR #1814, opened 2026-08-11
Match the CHANGELOG instruction to what CI enforces

abhishekraok:
Sync `CHANGELOG.md` instruction in `AGENTS.md` what CI enforces.

**Scope.** The instruction says to "always" add an entry. The check in `pr_checks.yml` only fires when the diff touches `open_instruct/`:

**The bypass is undocumented.** `CHANGELOG=<reason>` in the PR body skips the check, the same mechanism as `GPU_TESTS=bypass`, which *is* documented. Anyone hitting the check has no way to know the escape hatch exists.

**The PR URL is a chicken-and-egg.** The check requires the entry to contain the PR's own URL, but that URL doesn't exist until the PR is opened. Writing the entry while committing — the natural order, and what the instruction implies — produces a failing check on every new PR. Spelling out the order (entry → open PR → amend with URL) saves a round trip.

GPU_TESTS=bypass
CHANGELOG=documentation-only change to AGENTS.md

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -7,7 +7,9 @@
 
 
 # Workflow
-- When creating a PR, always add a summary to `CHANGELOG.md` with a link to the PR (e.g., `- Description of change (https://github.com/allenai/open-instruct/pull/123).`).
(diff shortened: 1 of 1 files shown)


undfined (approved):
lgtm

Issue #1800, opened 2026-08-05
Chat template problem in SFT open-instruct


abhishekraok:
Multi-turn SFT tokenization looks broken in olmo_core_finetune.py. Tried the
documented Olmo 3 recipe — Olmo-3-1025-7B + olmo_thinker_no_think_sft_tokenization
+ Dolci-Instruct-SFT — and it fails:

  ValueError: Cannot compute assistant label spans ... not prefix-stable

Cause: the olmo* templates append eos_token only after the final assistant turn,
so render(messages[:i+1]) isn't a prefix of the full render, and the label-span
derivation rejects it. Every assistant turn is checked by
sft_tulu_tokenize_and_truncate_v1, so any multi-turn row trips it. It only worked
for me on tulu-3-sft-personas-algebra because that data is single-turn.

--transform_fn last_turn_tulu_tokenize_and_truncate_v1 does work (only the final
turn is checked), but that masks all earlier assistant turns — 64% trainable
tokens, and presumably not what the released models did.

docs/olmo3.md says Olmo 3 SFT ran in olmo-core rather than open-instruct, and
mentions a masking bug

Failing experiment: https://beaker.org/ex/01KZ7E0BQMAX5DD0B244K19NVB
(Olmo-3-1025-7B + olmo_thinker_no_think_sft_tokenization + Dolci-Instruct-SFT, default transform)

abhishekraok:
The root design problem
Everything template-related traces to one decision: label spans are derived by differencing prefix re-renders, at dataset_transformation.py:1266-1291.

That assumes rendering is monotonic — render(messages[:i]) is a literal prefix of render(messages). Modern templates break it because they special-case the final assistant turn: olmo swaps <|im_end|> for EOS, Qwen3 injects a think block. Empirically 3 of 5 templates I tested fail. Two failure modes:

loud: prefix check fails → ValueError, which is what blocked rung 5
silent: prefix holds but the boundary lands inside the header → <think> gets masked and the model never learns to emit it
The docs call this "the underlying masking bug" and the think-tokenizer split is a workaround for it.

Principled fix: stop computing offsets. Mark assistant content in the template with {% generation %} blocks and call apply_chat_template(..., return_assistant_tokens_mask=True). Transformers returns the mask directly — no re-rendering, no offset arithmetic, no prefix-stability requirement, and it works with ChatML templates as they are. It's the standard solution to exactly this problem and it's available in the pinned version.

abhishekraok:
[...]

PR #1816, opened 2026-08-12
Add multimodal (VLM) SFT support to finetune.py

peter-sushko:
Multimodal SFT supported. Tested on qwen-3.5 and it worked. 

See docs/algorithms/multimodal_sft.md.

### DIFF
--- a/open_instruct/dataset_transformation.py
+++ b/open_instruct/dataset_transformation.py
@@ -45,6 +45,7 @@ The main things we are looking for are:
 
 import copy
 import hashlib
+import inspect
 import json
 import multiprocessing
 import os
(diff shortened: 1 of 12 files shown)


open_instruct/dataset_transformation.py
```diff
@@ -1486,6 +1576,63 @@ def sft_tulu_filter_v1(row: dict[str, Any], tokenizer: PreTrainedTokenizer):
...
+        if isinstance(image, str) and not os.path.isabs(image):
+            candidate = os.path.join(media_dir, image)
+            resolved.append(candidate if os.path.isfile(candidate) else image)
+        else:
+            resolved.append(image)
+    return resolved
+
+
+def sft_tulu_tokenize_mm_v1(
+    row: dict[str, Any], tokenizer: PreTrainedTokenizer, processor: Any, mm_plugin: Any, media_dir: str | None = None
```
abhishekraok, line 1599: This calls _tokenize_tulu_sft_with_assistant_labels directly rather than going through _tokenize_row_or_mask_out, which skips the safety net from #1806: a conversation whose assistant spans can't be derived raises out of dataset.map and kills the whole tokenization job, instead of being masked out and dropped by sft_tulu_filter_v1. On a large mixture that's a job-killer for one bad row. Routing through the wrapper should be a small change — the only difference is the max_seq_length=None and the images assignment afterwards.

Separately, on qwen3_5_nothink: the prefix instability you're working around is the same class as #1800, and the token-count fallback in [#1806](https://github.com/allenai/open-instruct/issues/1806) handles it without needing a hand-maintained template. Authoring a stable template is a legitimate choice — just flagging the alternative in case it saves you maintenance.

PR #1820, opened 2026-08-14
Choose the HF checkpoint weight naming per model type, and verify the write

abhishekraok:
save_pretrained applies transformers' conversion_mapping in reverse by
default (save_original_format=True), writing legacy on-disk weight names.
For olmo_hybrid that produced checkpoints vLLM -- and therefore olmo-eval --
cannot load: all 56 norms went out as attention_layer_norm /
feedforward_layer_norm instead of input_layernorm / post_attention_layernorm.

The correct value differs per model type, so it is chosen via
_MODERN_NAMING_MODEL_TYPES rather than set unconditionally. Everything else
keeps transformers' default; of the model types reachable through this path,
olmoe is the one genuinely released in the legacy per-expert layout, so no
currently-working export changes behavior. OLMoE is being dropped separately
in #1837.

Also verifies after writing that every converted tensor reached the file.
That failure was otherwise silent: save_pretrained returns cleanly and the
checkpoint only breaks in whatever later tries to load it.

save_state_dict_as_hf is shared with the DPO and GRPO export paths.

Co-authored-by: Claude <noreply@anthropic.com>


abhishekraok:
**Do not merge as-is — this is wrong for MoE model types.**

I set `save_original_format=False` unconditionally, but the correct value differs per model type:

| model | released on-disk layout | correct setting |
|---|---|---|
| `olmo_hybrid` | modern (`input_layernorm`, `post_attention_layernorm`) | `False` |
| `olmoe`, `mixtral`, `qwen2_moe` | legacy (per-expert `mlp.experts.N.gate_proj`) | `True` (the default) |

Verified against `allenai/OLMoE-1B-7B-0924`: 3072 expert keys in per-expert form, no fused `gate_up_proj`. `olmoe` aliases to `qwen2_moe` in transformers' `conversion_mapping`, so it has a renaming entry too — meaning this change would rewrite OLMoE exports into the fused layout and break them the same way the hybrid was broken.

`olmo_hybrid` is the odd one out: its released checkpoint is already in what transformers treats as the post-conversion naming.

The right rule is to match whatever `original_model_name_or_path` uses on disk rather than hardcoding either value. The post-write verification in this PR is still correct and worth keeping — it is what would have caught this.

abhishekraok:
Fixed. The flag is now chosen per model type rather than unconditionally:

```python
_MODERN_NAMING_MODEL_TYPES = {"olmo_hybrid"}
...
hf_model.save_pretrained(save_dir, save_original_format=model_type not in _MODERN_NAMING_MODEL_TYPES)
```

Everything except `olmo_hybrid` keeps transformers' default, so no currently-working export changes behaviour.

Verified both directions on small configs:

| model | on-disk result | matches released |
|---|---|---|
| `olmoe` | per-expert `mlp.experts.0.gate_proj`, no fused `gate_up_proj` | yes |
| `olmo_hybrid` | `input_layernorm`, no `attention_layer_norm`, all keys present | yes |

The post-write key check stays, and its error message now points at `_MODERN_NAMING_MODEL_TYPES` so the next person who hits this knows where to look.

undfined (approved):
Looks good overall but I believe we will get a false positive for an incorrectly converted OLMoE checkpoint with Transformers 5.5.3 (and probably older 5.x version as well). 

For OLMoE, I believe save_original_format=True converts each layer’s fused in-memory experts.gate_up_proj and experts.down_proj tensors into the released per-expert gate_proj/up_proj/down_proj layout. The new check compares the pre-conversion keys with the saved keys, so it falsely reports those two fused tensors as missing for every layer and raises on an otherwise valid checkpoint.

This is a pretty old arch so we could choose to no longer support without much concern tbh.

abhishekraok:
Thanks — agreed on dropping OLMoE. Filed [#1837](https://github.com/allenai/open-instruct/issues/1837) for the open-instruct side and [allenai/OLMo-core#838](https://github.com/allenai/OLMo-core/issues/838) for the OLMo-core side, each with an inventory of what's actually involved.

On the false positive though, I don't think it can trigger against the code as it stands. OLMo-core's converter ([olmo_core/nn/hf/convert.py:51-53](https://github.com/allenai/open-instruct/blob/pr/hf-export-weight-naming/olmo_core/nn/hf/convert.py#L51-L53)) maps MoE weights to the per-expert HF names and has no fused gate_up_proj mapping at all, so converted never holds the fused tensors. On our pinned 5.4.0 the check compares per-expert against per-expert and passes.

If 5.5.3 does build OLMoE fused in memory, I think the failure lands earlier than this check: hf_model.load_state_dict(converted, assign=True) is strict, and verify_can_save_as_hf runs the same key comparison at startup. That would make OLMoE broken on transformers ≥5.5 independently of this PR — which argues for dropping it rather than for something this PR has to handle.

PR #1822, opened 2026-08-14
Support SFT of Olmo-Hybrid-7B on the olmo-core path

abhishekraok:
olmo-core has no `TransformerConfig` preset for Olmo Hybrid and no HF -> olmo-core state conversion for `model_type: olmo_hybrid`, so `--config_name olmo3_hybrid_7B` (already referenced by `scripts/train/olmo-hybrid/7b_instruct_dpo_sweep_olmo_core_budget_ac.sh`) could never resolve — that preset does not exist in the pinned rev, in olmo-core `main`, or anywhere upstream. Both live in `open_instruct/olmo_core_hybrid.py` until they land there.

## Architecture notes

Two things differ from Olmo 3 and are easy to get wrong:

- It is **NoPE**. `rope_parameters` is `{"rope_theta": null}` and HF builds no rotary embedding, so the full-attention layers have no positional encoding. Do not pass `--rope_scaling_factor`.
- GDN blocks are **pre-norm** while attention blocks are **reordered-norm**, which is why the per-layer key maps are split in two.

Also note olmo-core's own `HYBRID_GDN_LAYER_KEY_MAP` disagrees with the released checkpoint (`out_proj`/`norm` vs `o_proj`/`o_norm`), so the maps here are taken from the checkpoint.

## Verification

Checked against `allenai/Olmo-Hybrid-7B` without a GPU: the config builds 523 tensors matching the checkpoint's 523 exactly (no missing keys, no extra keys, no shape mismatches), and feeding it back through olmo-core's own `get_hybrid_hf_config` reproduces the released `config.json` on every architectural field. Then on one GPU, the built model reproduces HF's logits to 0.6% with 100% argmax agreement.

That last check earned its place. Three structural checks passed green on code carrying a **38% logit error**: HF hardcodes `eps=1e-5` on the gated DeltaNet output norm rather than using `rms_norm_eps`, and nothing in the config or the weight names hints at it. Compounded over 24 GDN layers it gave 70% argmax agreement — close enough that training would have looked entirely healthy.

## Result

Full 0.1-0.5 epoch curve, 861/861 steps in 2h on one B300 node. Data, cache, seed and global batch are byte-identical to the Olmo 3 run, so the model is the only variable.

| checkpoint | epoch | IFBench (`prompt_level_loose_acc`) | GSM8K (`exact_match`) |
|---|---|---|---|
| `Olmo-Hybrid-7B` base | 0 | 0.1467 | 0.7498 |
| step 172 | 0.1 | **0.2367** | 0.7506 |
| step 344 | 0.2 | 0.2400 | 0.7710 |
| step 516 | 0.3 | 0.2433 | 0.7589 |
| step 688 | 0.4 | **0.2667** | 0.7498 |
| step 861 | 0.5 | 0.2600 | 0.7574 |
| released `Olmo-Hybrid-Instruct-SFT-7B` | 2-stage, ~15x tokens | 0.2733 | 0.8014 |

Roughly 80% of the IFBench gain arrives by 0.1 epochs (+9.0 points, ~2.6 sigma). From 0.1 to 0.5 the range is 0.2367-0.2667 against a ~3.5-point standard error on the difference, so the curve is flat within noise and the apparent 0.4 peak is not separable from sampling variation. GSM8K is flat throughout — no regression, no gain. The 0.4 checkpoint (0.2667) is inside noise of the released model (0.2733), from ~1/15 the SFT tokens on 8 GPUs rather than 64.

Throughput: **16,392 tok/s/device on B300**, 9,073 on H100, against Olmo 3's 6,580 at identical global batch.

The base row was measured with `max_position_embeddings` capped 65536 -> 32768 so vLLM would start inside olmo-eval's 300s budget. That is safe only because this model is NoPE — there is no positional embedding table for the larger value to index.

[Full report](https://wandb.ai/ai2-llm/open_instruct_internal/reports/Olmo-Hybrid-7B-SFT-through-open-instruct-it-works-on-Blackwell--VmlldzoxNzcyMjMzOQ).

Runs:

1. B300, completed 861/861 in 2h, exit 0: [Beaker](https://beaker.org/ex/01M013HTEWJYVHAKPHHJ762YJW)
2. H100, went silent at step 580/861: [Beaker](https://beaker.org/ex/01KZW9J5DQ31JAF18SPESYEWN0)
3. H100 + CUDA 13 image, failed on driver version: [Beaker](https://beaker.org/ex/01M01QDGJ98G7PVGXDGMAQZATT)
4. B300 + CUDA 12 image, failed on nvrtc arch: [Beaker](https://beaker.org/ex/01M01QX6RXVRAC1CPBVRE27539)

## Why the launch script pins ai2/holmes

Four H100 runs of this exact configuration went silent mid-training — steps 65, 205, 350 and 580, on four different nodes — stalled in the FSDP2 post-backward gradient reduce-scatter with every rank enqueued and none completing (allenai/OLMo-core#829). The same configuration completes on B300, so the script targets ai2/holmes and needs `--cuda-version 13`.

GPU architecture and CUDA toolkit are confounded here, and runs 3 and 4 above were an attempt to separate them that failed on both sides: H100 cannot run the CUDA 13 image (its driver reports 12.8, too old for that build) and B300 cannot run the CUDA 12 image (`nvrtc: invalid value for --gpu-architecture`). So Hopper-specific is the observed behaviour, not a demonstrated cause.

## Follow-ups, deliberately not in this PR

- `scripts/train/debug/check_olmo_core_matches_hf.py`, the script behind the logits check above. It is a generic "does this olmo-core config reproduce its HF reference?" tool rather than anything hybrid-specific, so it is worth its own review.
- The launch script wants `--dist_timeout_hours 2`; that flag arrives with #1821. Until then a mid-training hang holds its 8 GPUs for the full 24h default — this happened once, for 2h40m.


abhishekraok:
Two updates since this was opened.

**The script now targets Blackwell, not Hopper.** As originally written it pointed training at `ai2/jupiter`, where this exact configuration hung 4/4 times (steps 65, 205, 350, 580, across four nodes) in the FSDP2 post-backward reduce-scatter — allenai/OLMo-core#829. The same configuration completed 861/861 on a B300 node and ran 81% faster (16,392 vs 9,073 tokens/s/device). `ai2/holmes` requires the CUDA 13 image, so the documented launch command now passes `--cuda-version 13`. Tokenization stays on the CPU clusters.

Caveat recorded on the issue: the B300 run necessarily used CUDA 13 while every H100 run used CUDA 12, so toolkit version is confounded with architecture. A CUDA-13-on-H100 run is in flight to separate them; if it turns out to be the toolkit rather than the GPU, this cluster change should be revisited.

**Full results.** The completed run gives the 0.1–0.5 epoch curve:

| epoch | IFBench | GSM8K |
|---|---|---|
| 0 (base) | 0.1467 | 0.7498 |
| 0.1 | 0.2367 | 0.7506 |
| 0.2 | 0.2400 | 0.7710 |
| 0.3 | 0.2433 | 0.7589 |
| 0.4 | 0.2667 | 0.7498 |
| 0.5 | 0.2600 | 0.7574 |
| released 2-stage | 0.2733 | 0.8014 |

Roughly 80% of the IFBench gain arrives by 0.1 epochs; from 0.1 to 0.5 the spread is 3 points against a ~3.5-point standard error, i.e. flat within noise. GSM8K never moves. [Report](https://wandb.ai/ai2-llm/open_instruct_internal/reports/Olmo-Hybrid-7B-SFT-through-open-instruct:-it-works,-on-Blackwell--VmlldzoxNzcyMjMzOQ==).

farhatkevin:
@codex review

open_instruct/olmo_core_utils.py
```diff
@@ -546,7 +569,12 @@ def save_state_dict_as_hf(
     ``save_dir``.
     """
     hf_config = transformers.AutoConfig.from_pretrained(original_model_name_or_path, trust_remote_code=True)
-    converted = olmo_hf_convert.convert_state_to_hf(hf_config, state_dict)
+    if getattr(hf_config, "model_type", None) == olmo_core_hybrid.OLMO_HYBRID_MODEL_TYPE:
+        converted = olmo_core_hybrid.convert_hybrid_state_to_hf(
+            state_dict, olmo_core_hybrid.layer_types_from_hf_config(hf_config)
+        )
```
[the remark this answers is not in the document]
abhishekraok: Confirmed and fixed in 1daeab4.

Reproduced against a real hybrid state dict — the failure is a bit sharper than described. olmo-core's generic converter doesn't emit mismatched `out_proj`/`norm` names for the verifier to compare; it raises before the comparison happens:

```
RuntimeError: Some state keys were not converted:
['blocks.0.attention.A_log', 'blocks.0.attention.dt_bias',
 'blocks.0.attention.k_conv1d.weight', 'blocks.0.attention.o_norm.weight', ...]
```

So the run dies inside `convert_state_to_hf` rather than reaching the "HF export is not implemented" message. Same outcome for DPO and GRPO either way. The hybrid converter on the same input gives 523 keys, 0 missing, 0 extra.

Rather than add a third copy of the `model_type == olmo_hybrid` branch, both the preflight check and the save now go through one `convert_olmo_core_state_to_hf()`, mirroring `load_hf_weights_into_olmo_core()` for the load direction — one dispatching function per direction, so the two can't disagree about which converter a model needs. That also replaced the inline branch in `save_state_dict_as_hf`, so it's roughly size-neutral.

Covered by `test_export_dispatches_on_model_type`, which needs no GPU. Verified it's load-bearing: deleting the dispatch fails it.

open_instruct/olmo_core_utils.py
```diff
@@ -288,6 +288,21 @@ def build_scheduler(lr_scheduler_type: str, warmup_ratio: float, num_training_st
     raise ValueError(f"Unknown lr_scheduler_type: {lr_scheduler_type!r}")
 
 
+def load_hf_weights_into_olmo_core(
```
farhatkevin, line 291: Non-blocking: OLMo-core GRPO still calls `load_hf_model` directly at `grpo_olmo_core_actor.py:136` and `:206`, bypassing this Hybrid dispatcher. Consider routing those calls through `load_hf_weights_into_olmo_core` in a follow-up.

farhatkevin (approved):

open_instruct/olmo_core_utils.py
abhishekraok: Confirmed — filed #1842. It's reachable once this lands (the actor resolves the
config from `model_name_or_path`, and this PR maps `Olmo-Hybrid-7B` there), but it
fails loudly: nothing maps the GDN `linear_attn.*` keys, so `StateConverter` raises
before writing anything. No risk of a half-initialized model, so I'd rather fix the
dispatch alongside a GRPO run that gets past startup than claim support here.

PR #1823, opened 2026-08-14
Make --resume_from_checkpoint actually do something

abhishekraok:
`CheckpointConfig.resume_from_checkpoint` was declared but never read in the olmo-core SFT path, so passing it did nothing. Worse, on the olmo-core path (`use_hf_ckpt == False`) the trainer was built with `load_strategy=never` and the base weights were loaded before `fit()`, which sets `checkpoint_loaded` — so `fit()` never loaded anything: neither `--resume_from_checkpoint` nor the `output_dir` auto-resume worked at all on that path.

The fix builds the trainer with `load_strategy=never` unconditionally, drops `load_path`, and performs the loads explicitly in `fit()`'s own precedence order:

1. an existing checkpoint in `--output_dir` (with trainer + optimizer state) — a preempted or crashed run resumes;
2. otherwise `--resume_from_checkpoint`, for starting from a *different* run's directory — raising if that path holds no checkpoint, so a typo cannot silently restart from the base weights;
3. otherwise the base weights from `--model_name_or_path` (skipped on the HF path, where they were already loaded before the trainer was built).

**Behavior change:** an olmo-core SFT run whose `--output_dir` already contains a checkpoint now resumes instead of silently restarting from the base model. This is what the HF path already did (`load_strategy=if_available`); the two paths now behave the same.

Also documents the flag: resuming a preempted run needs only `--output_dir`; `--resume_from_checkpoint` is for warm-starting from a different directory.

Runs:

1. HF base (Qwen3-0.6B), fresh `output_dir` — trains from step 1, checkpoints saved: [Beaker](https://beaker.org/ex/01M0E5E1TZQP79PRCVAZVKVDCS)
2. Relaunch into the same `output_dir` — auto-resumes at `step=46/92, epoch=3` (trainer + optimizer state restored, not a restart): [Beaker](https://beaker.org/ex/01M0E697NFPR0JFW5G1QB64Z1M)
3. `--resume_from_checkpoint` into an empty `output_dir` — logs "Resuming from", starts at the loaded step, not 0: [Beaker](https://beaker.org/ex/01M0E69SH0326Y3530PS8BTXM6)
4. OLMo-2-1B HF base, fresh dir (mints the olmo-core checkpoint for runs 5–7): [Beaker](https://beaker.org/ex/01M0EGVSRWPM78M5BY174EMPBK)
5. olmo-core `model_name_or_path`, fresh `output_dir` — explicit base-weight load, starts at step 1 (trainer state deliberately not loaded): [Beaker](https://beaker.org/ex/01M0EJCEYANDPPSNC60NYFX2YE)
6. olmo-core path, relaunch into the same `output_dir` — auto-resumes at `step=48/96`, base load skipped. **Dead on main** (restarted from base weights): [Beaker](https://beaker.org/ex/01M0EMS4G4KKWGDGWD6SH0JQF2)
7. olmo-core path, `--resume_from_checkpoint` into an empty `output_dir` — resumes at `step=48/96`, base load skipped. **Dead on main**: [Beaker](https://beaker.org/ex/01M0EMV2HTSZ6X5VRZQXWTPCRE)

Note from validation: an HF-path checkpoint can only be reloaded on the olmo-core path when the HF config's `vocab_size` equals the tokenizer's `padded_vocab_size()` (true for OLMo-2's dolma2 preset, false for Qwen3: 151936 vs 151680). Pre-existing behavior, unrelated to this PR's change.
8. Typo'd `--resume_from_checkpoint` (dir with no checkpoint) — crashes with `FileNotFoundError: No checkpoints found` before any training, instead of silently restarting from base weights: [Beaker](https://beaker.org/ex/01M0G361XHDVCE9VRMR947M4GE)
9. Happy-path explicit resume after switching to `load_checkpoint()` — loads the requested dir's latest checkpoint and resumes at its step: [Beaker](https://beaker.org/ex/01M0G6RTATFTY3PSA051DFMWGS)


farhatkevin:
@codex review

open_instruct/olmo_core_finetune.py
```diff
@@ -311,10 +311,15 @@ def main(args: SFTArguments, tc: dataset_transformation.TokenizerConfig) -> None
     trainer_callbacks["config_saver"] = callbacks.ConfigSaverCallback(_config=config_dict)
     trainer_callbacks["garbage_collector"] = callbacks.GarbageCollectorCallback()
 
+    # load_path is only consulted when load_strategy is not `never`, so setting
+    # resume_from_checkpoint has to turn it on.
     load_strategy = LoadStrategy.never if not use_hf_ckpt else LoadStrategy.if_available
+    if args.checkpoint.resume_from_checkpoint is not None:
+        load_strategy = LoadStrategy.if_available
```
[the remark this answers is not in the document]
abhishekraok: There's a real bug here, but not this one.

`TrainerConfig.build()` doesn't load anything — loading happens in `fit()`, which gates the entire block on `not self.checkpoint_loaded` (trainer.py:679-682), and `trainer.load_checkpoint()` sets `_checkpoint_loaded = True` (trainer.py:922). So for `use_hf_ckpt == False` the base load runs first, sets the flag, and `fit()` skips loading altogether.

Weights and trainer state don't end up mismatched — nothing is restored at all. `--resume_from_checkpoint` was still silently ignored for olmo-core checkpoints, and the `output_dir` auto-resume this PR documents didn't happen on that path either, since `load_strategy` was `never` there.

Fixed in 1b6ee47 by doing the loads explicitly in `fit()`'s own precedence order: a checkpoint already in `output_dir`, then `resume_from_checkpoint`, then the base weights. Note this means an olmo-core SFT run whose `output_dir` already holds a checkpoint now resumes rather than restarting, which is the documented behavior but is a change on that path.

---

open_instruct/olmo_core_finetune.py
```diff
@@ -311,19 +311,25 @@ def main(args: SFTArguments, tc: dataset_transformation.TokenizerConfig) -> None
...
 
-    if not use_hf_ckpt:
+    # Loaded here rather than by fit(), which skips its own load once anything has
+    # been loaded: putting the base weights in first would suppress it. Precedence
+    # is fit()'s own -- an interrupted run in output_dir, then an explicit resume,
+    # then the base weights.
+    resumed = trainer.maybe_load_checkpoint(args.checkpoint.output_dir, load_trainer_state=True, load_optim_state=True)
+    if not resumed and args.checkpoint.resume_from_checkpoint is not None:
+        logger.info(f"Resuming from {args.checkpoint.resume_from_checkpoint}...")
+        resumed = trainer.maybe_load_checkpoint(args.checkpoint.resume_from_checkpoint)
```
farhatkevin: An explicit `--resume_from_checkpoint` is still silently ignored when the path contains no checkpoint: `maybe_load_checkpoint()` returns False, then training falls back to the base weights. A typo could therefore launch an expensive restart despite logging “Resuming from…”. Please fail when the explicitly requested checkpoint cannot be loaded, e.g. use `load_checkpoint()` here.

abhishekraok: Fixed in 8f930563e — the explicit-resume branch now calls `load_checkpoint()`, which raises `FileNotFoundError` when the path holds no checkpoint. Also added a warning when the flag is ignored because `output_dir` already resumed (mirrors `fit()`'s own load_path warning). Validating both paths on Beaker; will add the runs to the body.

abhishekraok: Nice catch.

abhishekraok: Validated both ways (runs 8–9 in the body): a `--resume_from_checkpoint` pointing at an empty dir now dies with `FileNotFoundError: No checkpoints found` before any training step, and the happy path still loads and resumes at the checkpoint's step.

farhatkevin (approved):
looks good!

PR #1829, opened 2026-08-17
Record the Beaker experiment in wandb for olmo-core runs

### LINKED ISSUE #1812: olmo-core runs don't record their Beaker experiment in wandb
`olmo_core_finetune.py` / `dpo.py` runs log to wandb with no reference to the Beaker job that produced them, so there is no way to get from a wandb run to its experiment (or back) without grepping job logs and matching timestamps.

## Already solved elsewhere

`utils.maybe_get_beaker_config()` returns a `BeakerRuntimeConfig` with `beaker_workload_id`, `beaker_experiment_url`, `beaker_node_hostname` and dataset ids. Both non-olmo-core paths merge it into the wandb config:

- `grpo_fast.py:1062` — `all_configs.update(vars(beaker_config))`
- `dpo_tune_cache.py:198` — `experiment_config.update(vars(beaker_config))`

The olmo-core path never calls it. `olmo_core_utils.build_common_callbacks` passes `config=config_dict` straight to `WandBCallback`, and that dict has no Beaker fields.

## Suggested fix

Merge `maybe_get_beaker_config()` into the config passed to `WandBCallback` in `build_common_callbacks`, matching what the other two paths do. Guarded by `is_beaker_job()`, so local runs are unaffected. This also covers the olmo-core DPO path, which shares those callbacks.

## Why it matters

Tracing a finished 7B SFT run back to its Beaker job currently requires `beaker experiment logs <id> | grep 'run-<timestamp>-<runid>'`, or matching the wandb `host` and `startedAt` against job metadata. The wandb run's `root` path does contain the Beaker-minted `$CHECKPOINT_OUTPUT_DIR` suffix, but that lives under `deletable_checkpoint_states` and is not durable — and it is ambiguous when a run resumes into a previous run's folder, where one directory maps to two experiments.

Concrete example: wandb run `ed453531` in `ai2-llm/open_instruct_internal` is Beaker experiment `01KZHT32T30M2VHCWKRJS1G9P7`, and confirming that took log grepping plus a timestamp match.

VishnuR23:
Closes #1812.

## Problem

olmo-core SFT (`olmo_core_finetune.py`) and DPO (`dpo.py`) runs log to wandb with no reference to the Beaker job that produced them. Getting from a wandb run back to its experiment means grepping job logs and matching timestamps.

Both non-olmo-core paths already do this:

- `grpo_fast.py:1061` — `all_configs.update(vars(beaker_config))`
- `dpo_tune_cache.py:197` — `experiment_config.update(vars(beaker_config))`

The olmo-core path never calls `maybe_get_beaker_config()`. `build_base_callbacks` passes `config_dict` straight to `WandBCallback`, and that dict has no Beaker fields.

Worth noting for review: `build_base_callbacks` *does* already install a `BeakerCallbackV2`, so this looks covered at a glance. It isn't — that callback writes the **wandb URL into the Beaker experiment description**, which is the opposite direction. Nothing writes Beaker identifiers into wandb.

(The issue names the helper `build_common_callbacks`; the function is `build_base_callbacks`. Same function, and it is shared by both SFT and DPO as the issue describes.)

## Fix

Adds `build_wandb_config()` and uses it for the `WandBCallback` config. Both callers of `build_base_callbacks` are covered. `maybe_get_beaker_config()` returns `None` off Beaker, so local runs are unaffected — no extra `is_beaker_job()` guard is needed.

Two deliberate details:

**The lookup is rank 0 only.** `maybe_get_beaker_config` shells out to the Beaker CLI once for the experiment and again per job result (`get_beaker_dataset_ids` → `beaker dataset get`). `build_base_callbacks` runs on every rank, so unguarded this would issue hundreds of redundant subprocess calls on a 224-GPU Olmo 3 run. Only rank 0 creates the wandb run, so nothing is lost. `get_rank()` returns 0 when not distributed, so single-process runs still get the fields.

**The merge returns a new dict rather than mutating.** The same `config_dict` is also passed to `BeakerCallbackV2` and `ConfigSaverCallback`, both of which serialize it to disk, so in-place mutation would leak the Beaker fields into those artifacts. There is a test pinning this.

## Testing

Five unit tests in `open_instruct/test_olmo_core_finetune.py`, alongside the existing `olmo_core_utils` coverage there. I confirmed all five fail against unmodified `olmo_core_utils.py` and pass with the change, so they genuinely pin the behaviour:

- Beaker fields are merged into the config
- config is returned unchanged off Beaker
- the caller's dict is not mutated
- `build_base_callbacks` wires the fields through to the wandb callback (the actual issue)
- non-zero ranks do not call the Beaker CLI

Local results on macOS:

- `uv run pytest open_instruct/test_olmo_core_finetune.py` — 19 passed
- `uv run ruff format --check`, `ruff check`, `ty check`, `compileall` — all clean

I do not have Beaker access, so the end-to-end path (a real job populating `BEAKER_WORKLOAD_ID` and the fields landing in a wandb run config) is covered by mocks rather than an actual run. That is the one claim here I could not verify directly, and it is the thing worth checking before merge.

GPU_TESTS=bypass


VishnuR23:
Rebased onto `main` (098424c4). Conflict was the `unittest.mock` import in `test_olmo_core_finetune.py`; I adopted `from unittest import mock`, which is both what main now uses and what `AGENTS.md` prescribes ("Use `from package import module`"). Re-verified after the rebase: 27 passed in that file, and the 5 tests added here still fail against main's `olmo_core_utils.py`.

One note on overlap, since it is easy to assume this is now redundant. #1872 writes a provenance `README.md` into the SFT `output_dir` containing Beaker and W&B pointers, which solves traceability *from a checkpoint directory*. This PR is the other direction — from a W&B run back to its Beaker experiment — which #1872 does not cover, because nothing it adds reaches the wandb config. `build_base_callbacks` on current main still passes `config=config_dict` unchanged.

I also kept `maybe_get_beaker_config()` rather than switching to the newer `get_beaker_experiment_url()` helper #1872 added. The former returns the full `BeakerRuntimeConfig` (workload id, experiment URL, node hostname, dataset ids) that `grpo_fast.py` and `dpo_tune_cache.py` already merge, and matching those two paths is the point of this change. Happy to switch if you would rather the olmo-core path record only the URL.

PR #1833, opened 2026-08-19
Fix factuality and refusal judge verdicts collapsing to 0.0

arthi-arumugam-git:
### The wrong number

A judge that follows the factuality template's own instruction scores every verdict 0.0. The template says *"If the Provided Answer is correct say exactly \"True\", otherwise say \"False\""* and its JSON example quotes the score slot, so a compliant judge returns `"SCORE": "True"`. `extract_json_score_with_fallback` calls `float()` on that, which raises `ValueError`, the regex fallback matches digits only, and the outer `except` returns the 0.0 error sentinel with a warning that calls the compliant output "invalid json". The result: correct and incorrect answers are indistinguishable to the factuality judge, and the refusal template is a copy of the same text, so it has the same dead signal.

Reproduction on `main` (6256267):

```
>>> extract_json_score_with_fallback('{"REASONING": "Dates match.", "SCORE": "True"}')
WARNING - Could not parse score from due to invalid json: ... defaulting to 0.0
('{"REASONING": "Dates match.", "SCORE": "True"}', 0.0)
>>> extract_json_score_with_fallback('{"REASONING": "Wrong person.", "SCORE": "False"}')
('{"REASONING": "Wrong person.", "SCORE": "False"}', 0.0)
```

### The fix

`SCORE` now passes through a small `coerce_score` helper: `"True"`/`"False"` strings in any case, and JSON boolean literals, map to 1.0/0.0; numeric strings and numbers keep exact float semantics; anything else still raises and lands on the existing 0.0 sentinel path. The regex fallback accepts `True`/`False` alongside digits. Nothing else in the extractor changes, and the other judges routed through this extractor (quality_rubric, safety, creative_writing) emit numeric scores, which behave exactly as before, covered by tests.

### Tests

`open_instruct/test_judge_utils.py` is new. On `main` the behavioural tests fail with both verdicts equal to 0.0 (they import only symbols that exist on `main`; the `coerce_score` unit tests skip there). With this branch: 19 passed.

### Not changed

The refusal template's TODO ("just a copy") and the empty safety_template stub are untouched; this PR only makes the extractor honour the contract the templates already declare. Happy to file the template TODOs as a separate issue if useful.


arthi-arumugam-git:
Merged main in, the CHANGELOG clash with #1820 is sorted (kept both entries in the existing order). Tests still pass, 19 of them. Ready when you are.

PR #1838, opened 2026-08-19
Detect olmo-core checkpoints by config contents; let mason pass a GCS credential

abhishekraok:
Two small, independent fixes that came out of running SFT from olmo-core checkpoints.

### `is_hf_checkpoint` misidentifies olmo-core checkpoints

It treated any directory containing a `config.json` as HuggingFace. olmo-core checkpoints carry one too — the full experiment config — so passing an olmo-core checkpoint directory as `--model_name_or_path` took the HF branch and failed downstream, despite `setup_model` explicitly supporting olmo-core checkpoints via `--config_name`.

Decide on contents instead: HF configs carry a top-level `model_type`, olmo-core's do not.

**This deliberately flips the behaviour asserted by `test_local_dir_with_config_json_is_hf`.** That test encoded presence-only detection, which is the bug. It's replaced by three tests: an HF config, an olmo-core config, and an unreadable one (which stays non-HF rather than raising).

### mason: pass a GCS credential to jobs

Adds `GOOGLE_APPLICATION_CREDENTIALS` to `useful_secrets`, so jobs can read `gs://` paths through olmo-core's io layer, which speaks GCS natively. 

GPU_TESTS=bypass


open_instruct/olmo_core_utils.py
```diff
@@ -335,11 +335,21 @@ def is_hf_checkpoint(path: str) -> bool:
     """Detect whether a model path is a HuggingFace checkpoint (vs olmo-core format).
 
     Returns True for HF hub IDs (e.g. 'allenai/Olmo-3-1025-7B'), local/weka paths
-    containing config.json, and paths with a '-hf' component. Returns False for
+    holding an HF config.json, and paths with a '-hf' component. Returns False for
     olmo-core distributed checkpoints.
     """
     if os.path.isdir(path):
-        return os.path.isfile(os.path.join(path, "config.json"))
+        config_path = os.path.join(path, "config.json")
```
farhatkevin, line 365: **[P2] Handle GCS checkpoint URLs before the HF fallback**

`gs://` OLMo-core checkpoints are still classified as HF: they are neither local directories nor absolute filesystem paths, so this function returns `True`. `setup_model()` then calls `AutoConfig.from_pretrained()` instead of the OLMo-core GCS loader and fails before the new credential is useful. Please handle cloud checkpoint URLs explicitly (preserving the `-hf` convention or inspecting the remote config) and add a `gs://` regression test.

abhishekraok: Good catch — implemented in cb7019b36. Remote URLs without an -hf marker are now classified as olmo-core (the -hf convention still marks remote HF exports), with gs:// regression tests for both cases..

farhatkevin (approved):
lgtm!

PR #1839, opened 2026-08-20
Add a script that checks an olmo-core config reproduces its HF logits

abhishekraok:
Loads a model twice, once through `transformers` and once by building an olmo-core `TransformerConfig` and converting the HF weights into it, then runs both on the same input and compares logits. Per-block divergence is reported too, so a mismatch names the first block that drifts.

Key-and-shape checks cannot see a right-shaped tensor behind the wrong name. On the hybrid conversion three of them passed clean on code carrying a 38% logit error.

```
python scripts/train/debug/check_olmo_core_matches_hf.py \
    --model-name allenai/Olmo-Hybrid-7B --config-name olmo3_hybrid_7B
```

GPU_TESTS=bypass — one manually-run script under `scripts/`, no `open_instruct/` code.

### DIFF
--- a/scripts/train/debug/check_olmo_core_matches_hf.py
+++ b/scripts/train/debug/check_olmo_core_matches_hf.py
@@ -1,0 +1,124 @@
+"""Check that an olmo-core config + HF conversion reproduce the HF model's logits.
+
+Key-level checks (does every parameter name and shape line up?) cannot catch a
+conversion that puts the right-shaped tensor behind the wrong name, and neither
+can a state-dict round trip. Only running both models on the same input does.
+
+Needs a GPU: olmo-core's transformer forward does not run on CPU.
+
+Usage:
+    python scripts/train/debug/check_olmo_core_matches_hf.py \
+        --model-name allenai/Olmo-Hybrid-7B --config-name olmo3_hybrid_7B
+"""
+
+import argparse
+
+import torch
+import transformers
+
+from open_instruct import logger_utils, olmo_core_utils
+
+logger = logger_utils.setup_logger(__name__)
+
+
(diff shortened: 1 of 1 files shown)


scripts/train/debug/check_olmo_core_matches_hf.py
```diff
@@ -0,0 +1,141 @@
...
+    hf_attn = HF_ATTN_IMPLEMENTATIONS.get(args.attn_implementation)
+    logger.info(f"Loading HF model {args.model_name} with attn_implementation={hf_attn}")
+    try:
+        hf_model = transformers.AutoModelForCausalLM.from_pretrained(
+            args.model_name, dtype=torch.bfloat16, trust_remote_code=True, attn_implementation=hf_attn
+        ).to(device)
+    except (ValueError, ImportError) as exc:
+        # Not every architecture implements every backend, and running under HF's own
+        # choice is far more informative than refusing to run at all.
+        logger.warning(f"HF rejected attn_implementation={hf_attn!r} ({exc}); using its default instead.")
```
farhatkevin: When HF rejects the requested backend, this reloads it using HF’s default, but the olmo-core model below still uses args.attn_implementation. That recreates the different-kernel comparison and can trigger a false mismatch through the tolerance or strict argmax check. Please either fail with an actionable backend error or select a fallback backend and use it for both models.

abhishekraok: Fixed in da2faf44e — you're right, the fallback only moved HF. The backend is now
one variable both models read, so a fallback moves them together, and it re-raises
instead of retrying when the fallback is what already failed. Backends with no HF
counterpart (flash_4, te) are now rejected by argparse rather than silently
compared across kernels.

farhatkevin (approved):
looks good

PR #1840, opened 2026-08-20
Detect olmo-core checkpoints by config contents; let mason pass a GCS credential

abhishekraok:
Fixes #1838. Two small, independent fixes from running SFT off olmo-core checkpoints.

### `is_hf_checkpoint` misidentified olmo-core checkpoints

It treated any directory containing a `config.json` as HuggingFace, but olmo-core checkpoints carry one too (the full experiment config), so an olmo-core checkpoint directory passed as `--model_name_or_path` took the HF branch and failed downstream. Now decided on contents: HF configs have a top-level `model_type`, olmo-core's do not. An unreadable or malformed `config.json` stays non-HF rather than raising.

This deliberately flips `test_local_dir_with_config_json_is_hf`, which encoded the presence-only bug. It's replaced by three tests: an HF config, an olmo-core config, and an unreadable one.

### mason: pass a GCS credential to jobs

Adds `GOOGLE_APPLICATION_CREDENTIALS` to `useful_secrets` so jobs can read `gs://` paths through olmo-core's io layer. On-prem clusters (jupiter/ceres/holmes) have no GCE metadata server, so without a credential a job falls back to anonymous access and fails. Uses the existing user-scoped-over-shared precedence; no credential is injected unless the secret exists.

Both changes are behaviour-preserving for existing callers: HF hub IDs, `-hf` paths, and directories without a `config.json` are unaffected.

GPU_TESTS=bypass


abhishekraok:
Duplicate of #1838 — that PR already carried both fixes; the one extra improvement here (guarding a non-dict config.json) has been folded into #1838.

Issue #1857, opened 2026-08-28
MoE think DPO on the LC KDA 1.2B: Dolci-Think-DPO-7B via the olmo-core DPO path


abhishekraok:
Take the #1854 SFT checkpoint through DPO, so the proxy has a validated SFT -> DPO handoff before the hero base lands. Umbrella: #1853. RL is tracked separately in #1858.

The Olmo 3 paper's ordering is SFT -> DPO -> RL, and its ablation is explicit that DPO is a better RL starting point than SFT on the same data. So DPO is a prerequisite for the RL arm, not a parallel track.

## Design

| | |
|---|---|
| Start | `deletable_checkpoint_states/7u2rne45/step11768` — olmo-core DCP, 0.5 ep Dolci-Think, seq 32768 (#1854) |
| Data | `allenai/Dolci-Think-DPO-7B` — 150,000 pairs, 3.3 GB, the released think preference mix |
| Loss | `dpo_norm`, beta 5 — the `DPOExperimentConfig` defaults, matching `scripts/train/debug/dpo/7b_instruct_dpo_olmo_core.sh` |
| LR | sweep 8e-8 / 3e-7 / 1e-6 on a 20k-pair subset, then run the winner full. Released 7B think DPO used 8e-8, 7B instruct 1e-6; neither is validated at 1.2B active |
| Seq | 16384 — Olmo 3's think DPO length. The SFT ran at 32768, but DPO packs one pair per sequence, so length here is set by the data, not by the base |
| Hardware | 1x8 or 2x8 B300 (holmes / olmo-instruct). No `--preemptible`; pass an explicit `--timeout` |
| Tokenizer | `allenai/olmo-3-tokenizer-instruct-dev`, as SFT used |

**No second live model.** Both DPO entrypoints share `dpo_utils.build_reference_logprobs_cache` — reference logprobs are precomputed and cached rather than held in a frozen copy. DPO memory is roughly SFT memory; the extra cost is one forward-only pass over the pairs.

## Prior art and a known risk

**DPO has not reproduced strong gains on the hybrid models trained so far**, and that is still open. Treat a flat DPO result on this proxy as a plausible outcome with precedent rather than as evidence of broken plumbing — and run the dense control in step 3 so the two are distinguishable.

[...]

PR #1860, opened 2026-08-29
Support DPO on olmo-core MoE checkpoints (OLMoDDPModel)

abhishekraok:
Adds DPO support for olmo-core MoE checkpoints that train under `OLMoDDPModel` — the OLMoE3 latent-KDA models, which refuse FSDP2 by design and therefore cannot use `dpo.py`'s existing `DPOTrainModule`. First stage of the post-training pipeline work in #1857.

## What this does

**`MoEDDPDPOTrainModule`** on olmo-core's `OLMoDDPTrainModule`. It supplies only the stepping the DDP base needs and the transformer base does not: `model_parts` rather than `model`, explicit `finalize_grad_reduce()` (the microbatch context accumulates every non-final micro-batch under `no_sync()`), and dry-run handling that resets the MoE's router metrics so they do not leak into step 1.

**The DPO objective is shared, not copied.** It is factored into `DPOObjectiveMixin`, so the loss, the token-weighted metrics and their all-reduce exist exactly once and both train modules inherit them. Each train module is now just `__init__` and `train_batch`. A copy would let the dense and MoE paths drift, and a drift there would surface as an apparent architecture effect rather than as a bug — which is precisely what #1857 exists to measure. `test_olmo_core_moe_dpo.py` asserts the objective methods resolve to the same function objects on both classes.

**`dpo.py` no longer assumes an HF checkpoint** in four places:
- `setup_model` was not being passed `tc`, which olmo-core checkpoints need to derive the padded vocab size. They are also now built on meta — materializing an 18.5B-parameter MoE per rank would otherwise want ~37 GB of host RAM each.
- Weights are loaded from the olmo-core checkpoint before the reference log-probability cache is built, which is before the `Trainer` that would otherwise load them exists.
- The startup export check and the post-training HF export are skipped; those checkpoints are exported with `convert_moe_checkpoint_to_hf.py`.
- MFU reporting is now optional, since there is no HF config to derive model dimensions from. Throughput is still reported.

**Model and optimizer come from the checkpoint's own config json** for these models. An `OLMoDDPModel`'s optimizer shards its state across the device mesh, so substituting a plain AdamW would silently change both the memory profile and the update.

## Incidental fix

`build_ddp_train_module_config` moves out of `olmo_core_utils` into a new `olmo_core_ddp` module. Its top-level `olmo_core.optim.moe_optimizer` import made the shared utilities — and therefore every entrypoint and most of the test suite — fail to import on any olmo-core revision without the MoE work.

## Testing

`ruff format` and `ruff check` clean. 66 passed, 4 skipped across `test_olmo_core_moe_dpo.py`, `test_olmo_core_finetune.py`, `test_dpo_utils.py`, `test_padding_free_collator.py`.

Note for anyone running the tests locally on this branch: `uv sync` installs torch 2.11+cu130 but reverts torchvision to cu128, and the mismatch makes the whole suite fail to collect. Re-run `uv pip install --index-url https://download.pytorch.org/whl/cu130 --no-deps "torchvision==0.26.0+cu130"` after every sync.

## Hardware status

**Dense DPO works end to end** ([01M17R211C6GWABJW8KXQM1F4A](https://beaker.org/ex/01M17R211C6GWABJW8KXQM1F4A), W&B `fmdr3u34`), which validates the shared objective, the metrics and the reference-logprob cache:

| decile | loss | reward accuracy | margin |
|---|---|---|---|
| 1 | 0.7034 | 0.434 | -0.015 |
| 3 | 0.6850 | 0.558 | +0.043 |
| 10 | 0.6744 | 0.595 | +0.094 |

Accuracy crosses 0.5 by decile 3 and holds, margin grows monotonically, loss falls from ln(2), overall margin t=+11.3. Step 0 sits at exactly ln(2) with margin exactly 0 -- the signature of policy == reference at init, which is what confirms the reference cache is right.

**Read DPO metrics as binned trends, never as the W&B summary.** That run's final logged step reads accuracy 0.399, margin 0.005, loss 0.694: on the summary alone this looks like a total failure. A step is 16 pairs, so single steps are close to pure noise.

**The MoE path is blocked by a GPU-level fault and this PR should not be read as validating it.** Three separate B300 nodes were driven into unrecoverable Xid 31/43 faults and cordoned. The fault is on the first forward pass through the KDA MoE, ~95 s after the reference pass starts (consistent with torch.compile finishing), at batch 0 every time. Dropping `--reference_cache_batch_multiplier` from 4 to 1 changed the batch count 63 -> 250 and reproduced the fault identically, so per-call token count is not the trigger. Full evidence table in #1857.

What the MoE path *did* validate before faulting: model built, the checkpoint's own DDP optimizer compiled, `load_olmo_core_base_weights` loaded 8 ranks in ~57 s, `load_optim_state=False` honored. No OOM, no import error.

## If you are picking this up

1. **Do not relaunch the MoE smoke without a diagnosis** -- each attempt has cost a B300 node.
2. Build a **single-GPU repro of the forward alone**. It isolates the fault without risking 8-GPU nodes and is the artifact the olmo-core/KDA owners would need. The suspects are `DPOLMHead` materializing logits instead of the checkpoint's `fused_linear` path, and `concatenated_forward_olmo` making the batch dimension >1 where SFT's was 1.
3. A #beaker-users heads-up about the three cordoned nodes was drafted but never posted.
4. The dense control mode (`kda_moe_dpo.sh dense_control`) is the regression test for the objective; it is cheap and it passes.

GPU_TESTS=bypass -- the standard GPU suite does not cover this path (B300s plus the MoE olmo-core pin), and the MoE hardware validation is blocked as described above rather than skipped silently.


abhishekraok:
The four fixes independent of the MoE path are now split into standalone PRs: #1865 (Dockerfile google-cloud-cli), #1866 (data_loader vllm import), #1867 (datasets pin), #1868 (--max_train_samples in dpo.py). If you pick this branch up, rebase onto main once they merge — the same changes exist here and will collapse to no-ops.

abhishekraok:
**Candidate diagnosis for the reference-pass crash, found via Codex review of #1868.**

The `--max_train_samples` implementation on this branch selects after shuffling but keeps the pre-shuffle `index` column. `build_reference_logprobs_cache` allocates tensors at subset size (2,000) and scatters by that column, whose values range up to ~149,985 — an out-of-bounds CUDA scatter, i.e. an illegal memory access, on the first cache batch.

This fits every observed signature of the MoE crash:
- **Deterministic at batch 0**: any first batch of 2,000 indices drawn from [0, 149,986) almost surely contains one ≥ 2,000.
- **Invariant to `--reference_cache_batch_multiplier`**: both the 63-batch and 250-batch configs crash at batch 0, which is what ruled out batch shape — consistent with OOB, which doesn't care about batch size.
- **~95 s in, attributed to the first forward**: the scatter runs immediately after batch 0's forward, and CUDA reports async errors at the next sync, so an OOB scatter can masquerade as a forward-pass fault.
- **Dense control unaffected**: it never passed `--max_train_samples` (it uses a 20k-row mixer with no select), so it never hit this path. SFT has no reference cache at all.

Not proven — the Xid 31/43 plus node cordoning is heavier fallout than an app-level illegal access usually causes, so a KDA-kernel fault isn't excluded. But it revises the 'do not relaunch without a diagnosis' guidance: there is now a diagnosis worth one test. If you pick this up, rebase onto main after #1868 merges (or cherry-pick its two commits) and re-run the MoE smoke once; a cheap pre-check is asserting `dataset['index'].max() < len(dataset)` before the reference pass.

abhishekraok:
Claude Code · agent-written on Abhishek's instruction (2026-09-15)

Closing: we are not pursuing DPO on the KDA MoE checkpoints right now, and everything this branch adds beyond DPO is superseded. For SFT on OLMoE3 latent-KDA / Olmo 3.5 hero checkpoints use kda-think-sft (#1886), which carries the shared OLMoDDP runtime, the config generator and the Dolci-Think launch script.

The MoE DPO trainer (`olmo_core_moe_dpo.py`, its DDP train module and tests) and the validated dense DPO run live only here, so the branch `dpo/kda-moe` stays in the repo as the rebase point if DPO comes back. Last state: merged with main and pinned to OLMo-core `89e7dcb7`, the same pin as #1886 (untested on this branch).

Issue #1853, opened 2026-08-25
Auto-research: SFT data mixture search on a fast proxy for Olmo 3.5


abhishekraok:
Systematically search SFT data mixtures on a fast proxy model, so the Olmo 3.5 hero SFT run starts from measured settings rather than guesses. Umbrella issue for the mixture-search programme; specific dataset additions are tracked separately (#1851, #1852).

**Scope (updated 2026-08-28): think-only.** The programme trains on Dolci-Think; instruct data is out. This retires the two axes that were previously ranked 1 and 2 in the search space (think:instruct ratio, and mixed-vs-staged think→instruct) — see *Search space* below. Instruct results are kept where they are *evidence* for the think-only choice, not as things to sweep.

**Timeline.** ~1 month of proxy work while pretraining finishes, then ~2 weeks for hero SFT, then ~2 weeks DPO + RL. The proxy work has to be done by the time the hero base lands.

**Deliverable.** Not "the best mixture" — a *mixture-vs-metric response surface with error bars*, so that when the hero base arrives with a different midtrain profile we can reason about how to adjust rather than reopening the search with two weeks left.

## Proxy model

OLMoE3 latent-KDA 1.2B. The full loop (tokenize → SFT → HF convert → eval) is validated end to end: #1843.

**Switch the base from midtrain to long-context.** Pipeline validation used the midtrain checkpoint only because it was the only stage with an HF export at the time. Production SFT happens *after* long-context extension — both dense `Olmo-3-1025-7B` and `Olmo-Hybrid-7B` were SFT'd from post-LC 65k bases, so midtrain made the MoE a stage behind everything it was compared against. Use:

```
oe-training-default/ai2-llm/checkpoints/jacobm/olmoe3/eval-comparisons/legacy-cx8-v2/latent-kda-l2/1.2B/long-context/
```

Its HF config declared `max_position_embeddings: 8192` against 65536 trained; fixed in place 2026-08-26 with Jacob's authorization (all 9 LC exports had it; backups + revert at `~/handoff/lc-config-fix-backup/`). Switching bases invalidates the existing baseline set, so run it as a deliberate arm, not a silent substitution.

The LC think baseline (0.5 epoch, seq 32768) is running as its own tracked arm: #1854.

[...]

PR #1864, opened 2026-08-31
Increase Beaker CLI download timeout

### LINKED ISSUE #1845: Increase cURL timeout in Dockerfile
Hi,

The `--max-time` value in the Dockerfile seems unnecessarily low and can cause the build to fail (even on a gigabit connection it seems).

```
RUN curl --silent \
    --connect-timeout 5 \
    --max-time 10 \
    --retry 5 \
    --retry-delay 0 \
    --retry-max-time 40 \
    --output beaker.tar.gz \
    "https://beaker.org/api/v3/release/cli?os=linux&arch=amd64&version=${BEAKER_VERSION}" \
    && tar -zxf beaker.tar.gz -C /usr/local/bin/ ./beaker \
    && rm beaker.tar.gz
```

It builds successfully after increasing this to 100.

JimmyWang0417:
## Summary

- Increase the Beaker CLI download timeout from 10 seconds to 180 seconds.
- Extend curl's retry window from 40 seconds to 600 seconds so slow initial downloads can still be retried.

## Testing

- Reproduced the failure with a 2 MiB/s transfer limit: the 10-second timeout stopped the 24,440,394-byte download after 15,943,118 bytes.
- Repeated the same download with the 100-second timeout: it completed in 13.919 seconds, and the archive contained `./beaker`.
- Extracted the downloaded binary and ran `beaker --version` (`v1.5.235`).
- Verified the updated curl options are supported by curl 8.21.0.
- `python -m compileall -qq open_instruct mason.py test_mason.py`
- `git diff --check`

I could not rerun the Beaker download in this environment because DNS could not resolve `beaker.org`. A full CUDA image build is unavailable because the Docker daemon is inaccessible. Ruff is not installed in this checkout; the Dockerfile-only change does not touch Python sources.

Closes #1845


abhishekraok:
Claude Code · bridge-cse session (Abhishek's #1853/#1859 anchor work) — agent-written

Data point from hitting this same failure twice on 2026-08-26: beaker.org served the ~24 MB CLI at ~0.5 MB/s, so a clean download takes ~48 s. `--max-time 100` covers that, but note `--retry-max-time 40` is still in place — curl only starts a retry while total elapsed time is under that limit, so once the first slow attempt runs past 40 s, no retry ever fires. Suggest raising it alongside (I ran with `--max-time 180` / `--retry-max-time 600` on a branch and it rode out the slow window).

JimmyWang0417:
Thanks for catching the interaction with `--retry-max-time` and for testing the slower download. I’ve updated the Dockerfile to use `--max-time 180` and `--retry-max-time 600`, so retries remain possible during a slow download.

abhishekraok:
Thanks for the fix.

abhishekraok (approved):

PR #1865, opened 2026-08-31
Install google-cloud-cli, not the dropped google-cloud-sdk package

abhishekraok:
Google renamed the package years ago and has now removed the `google-cloud-sdk` transitional package from its apt repo (the cloud-sdk suite's amd64 index lists `google-cloud-cli` and no `google-cloud-sdk`). Cold image builds therefore fail with:

```
Package 'google-cloud-sdk' has no installation candidate
```

A cached layer keeps working, so this breaks with **no change to the Dockerfile** — which is how a build passed three hours before an identical one failed. Anyone doing a cold build (fresh machine, cache eviction, `--no-cache`) is hit today.

Ruled out along the way, so nobody re-treads: the apt repo is reachable and its Release file current; the cuda13 base image has intact Ubuntu sources; doca-host does not clobber sources.list; the deprecated `apt-key` form is not at fault.

Verified on the `dpo/kda-moe` branch: cold builds fail before this change and succeed after (Beaker images built and ran).

GPU_TESTS=bypass


gregorybchris (approved):

gregorybchris:
I had to do this as well in #1873 so I'll rebase if you want to merge this first

PR #1873, opened 2026-09-02
Upgrade torch to 2.13

gregorybchris:
# Upgrade to PyTorch 2.13 / vLLM 0.28 on a single CUDA 13.0 build

Moves the training stack to **PyTorch 2.13.0** and **vLLM 0.28.0**, and replaces the CUDA 12.8 / CUDA 13.0 build split with a single CUDA 13.0 path for the B300 clusters.

The split cannot survive the upgrade: PyTorch 2.12 dropped its cu128 build, and no prebuilt flash-attn exists for cu129 + torch 2.13, so CUDA 13.0 is the only version every dependency in this graph publishes for. It covers x86_64 and aarch64 alike, so the `cuda12`/`cuda13` dependency groups, the second Docker build arg, and the separate CUDA 13 image job all go away with it.

## Dependency changes

- torch 2.10 → **2.13.0**, torchvision 0.25 → 0.28.0, torchaudio 2.10 → 2.11.0 (its final release), triton 3.6 → 3.7.1
- vLLM 0.19.1 → **0.28.0**
- transformers → 5.16.1, and the `override-dependencies` floor raised `>=5.4.0` → `>=5.15.0`. That override supersedes vLLM's own floor, so the old value let the lock keep a transformers older than vLLM supports.
- deepspeed, liger-kernel and accelerate floated up — all tightly torch-coupled
- flash-attn and flash-attn-3 wheels repointed to their cu130 + torch 2.13 builds
- **flash-attn-4 pinned to `4.0.0b19`** — the newest release whose declared floors vLLM 0.28's own pins already satisfy (`nvidia-cutlass-dsl>=4.5.2` against vLLM's 4.6.2, `apache-tvm-ffi>=0.1.5` against vLLM's 0.1.11), so **no dependency override is needed**. Neighbouring versions don't fit: b5 and earlier call `cute.core.ThrMma`, which moved to `cute.atom` in cutlass 4.6.x; b20–b26 pin `cutlass==4.6.0.dev0`; b27+ require `apache-tvm-ffi>=0.1.12`, above vLLM's pin.
- Two new floors, both because a vLLM dependency declares less than it needs: `openai>=2.25.0,<3` (vLLM 0.28's tool parsers import `NamespaceTool`, added in 2.25.0, while vLLM declares only `>=2.0.0`) and `datasets>=4.8.4,<5` (torchvision 0.28 removed `torchvision.io.VideoReader`; datasets guarded that import in 4.8.4).

## Code changes

Nearly all of the diff is dependency and build configuration. One substantive source change is forced: **vLLM 0.28 removed `trainer_send_weights` and the `*TrainerSendWeightsArgs` dataclasses**, so trainer-side weight sync moves onto its `TrainerWeightTransferEngine` API — a `WeightSource` over the training model, a `VLLMWeightSyncClient`, and `start_weight_update` / `finish_weight_update` RPCs on `LLMRayActor`. `RayEngineWeightSyncClient` re-stamps a per-engine `rank_offset`, since we run N independent engines where upstream assumes a single deployment.

Smaller forced changes: vLLM renamed the Qwen3 XML tool parser; transformers 5.16 types `config.num_labels` as optional, which `ty` rejects; and the deterministic-generation fixture was regenerated for the new sampling stack.

Build: a single CUDA 13.0 Dockerfile stage; `UV_PROJECT_ENVIRONMENT` pinned, because uv ≥ 0.12 otherwise places the project environment in the build cache mount and ships an image with no Python in it; the image's uv bumped to 0.12.8 to read lockfile revision 4; and `google-cloud-sdk` → `google-cloud-cli` after upstream renamed the package.

## Testing

- **GPU suite on B300: 53 passed, 1 skipped** — run by CI against this branch
- **Multi-GPU FSDP2 GRPO on B300: exit 0** ([`01M1PZ2ETY7WJCBR82MH90VWV7`](https://beaker.org/ex/01M1PZ2ETY7WJCBR82MH90VWV7)) — 2 FSDP2 learners plus 2 vLLM engines at TP2, covering the weight-transfer rendezvous, per-engine rank offsets, initial sync, and four training steps with checkpoints
- `make style && make quality` clean; the local suite matches main's baseline exactly

## Before merge

- **Depends on allenai/OLMo-core#857.** With flash-attn 4 available, Blackwell selects the `flash_4` backend, and OLMo-core's `dispatch_flash_attn_4` passes four arguments positionally that newer flash-attn 4 reordered. The pinned OLMo-core rev should be bumped once that merges.
- The single skipped GPU test is `TestFSDP2BroadcastWithVLLM`, skipped in source as needing multi-GPU; the GRPO run above exercises that path instead.
- **`open_instruct_auto_cuda12` is retired** — there is no cu128 torch 2.13 to build it from. Consumers should repoint to `open_instruct_auto`.


abhishekraok:
Claude Code · session 09974a79 (claude.ai/code), for @abhishekraok

We just ran this same road one exit short of you: vllm-026-waypoint (#1874) is a GPU-validated 0.26/torch 2.11 bump (single-GPU + 2-node GRPO smokes green on H100s). Most of its vllm-side work is superseded by your 0.28 weight-sync rewrite, but a few findings from the GPU runs apply to this branch verbatim and are invisible until you run on hardware — sharing them so you don't burn the same debug cycles.

**1. This branch will fail at the first Ray actor export: `cannot pickle 'ConfigModuleInstance'`.** torch ≥2.11 config modules (e.g. `torch.distributed.config`) are `torch.utils._config_module.ConfigModule` instances subclassing `ModuleType`, and Ray's vendored cloudpickle serializes modules by reference only on an *exact* `ModuleType` check — so any actor/function whose globals transitively reference `torch` dies at export. We verified this still holds for your lock: torch 2.13 keeps `ConfigModuleInstance` with no pickle support ([`_config_module.py:179`](https://github.com/pytorch/pytorch/blob/v2.13.0/torch/utils/_config_module.py#L179)) and your uv.lock has the identical ray 2.53.0. The fix we validated (in `vllm_utils.py`, next to your `MambaSpec` patch — commit `eeed94c3f` on #1874):

```python
from torch.utils import _config_module

def _reduce_config_module(self: _config_module.ConfigModule) -> tuple[Any, tuple[str]]:
    return (importlib.import_module, (self.__name__,))

_config_module.ConfigModule.__reduce__ = _reduce_config_module
```

**2. Budget `--timeout 3h`, not 1h, for the multi-node GRPO smoke.** transformers 5.16 invalidates the transformed-dataset cache key, so the first run cold-tokenizes (~45 min for the 25k-sample `large_test_script.sh` set). Our first 2-node attempt "exited 0" but was actually SIGTERM'd mid-tokenization at the 1h default — it looks like a pass in `beaker experiment get` unless you read the logs.

**3. Scheduling under the new allocation-based scheduler** (jupiter/titan/ceres since Aug 28): don't use `--preemptible` (it maps to backfill-only); set a small `minRuntime` instead. `ai2/olmo-instruct` has no ceres allocation (422 at submit); `ai2/open-instruct-dev` has both jupiter and ceres and scheduled our 2-node job in ~3 min. Also: a canceled Beaker job never reaches `exited` — poll for `canceled` too or a watcher waits on a corpse.

**4. Validation recipe on offer.** The two smokes that caught all five of our GPU-stage failures (each at a strictly later pipeline stage — worth re-running after every fix rather than batching): `scripts/train/debug/single_gpu_on_beaker.sh` (IPC weight path, ~8 min) then `scripts/train/debug/large_test_script.sh` (2-node NCCL path, 17 weight syncs). Happy to point them at this branch on holmes — say the word.

**5. Nit:** the PR body says flash-attn-4 is "left at beta5 deliberately" but the branch correctly removes it entirely (the pyproject comment has the right analysis — beta5 dies at runtime on `cute.core.ThrMma` under cutlass ≥4.6, which only surfaces on GPU import). The branch is right; the body is stale.

abhishekraok (approved):
Looks good to me. Approving on the understanding that this merges after the target clusters support CUDA 13 and OLMo-core #857 is merged, with this PR's OLMo-core pin updated to include that fix.

Issue #1804, opened 2026-08-07
Thinker chat templates cannot be used for SFT: every row's labels are undecidable


abhishekraok:
## Summary

The `*_thinker*` chat templates rewrite assistant content while rendering, so no span-based
derivation can locate the assistant turn. After the detection added for #1800, this no longer
corrupts labels silently — but it means **every conversation using these templates is dropped**,
so the templates are effectively unusable for SFT.

Affected (verified against the bundled test tokenizer, both `eos_token` variants, both
alternating and consecutive-assistant conversation shapes — 16 of 44 swept combinations):

- `olmo_thinker`
- `olmo_thinker_remove_intermediate_thinking`
- `tulu_thinker`
- `tulu_thinker_r1_style`

## Why derivation fails

Both available derivations locate an assistant turn by re-rendering conversation prefixes —
either as strings (char offsets into the full render) or as token counts. Both assume the turn
appears in the full render the same way it appears in a prefix render. The thinker templates
break that assumption directly: they split on `</think>`, strip intermediate thinking, or inject
a `<think>` opener into the generation prompt, so the rendered turn is not the message content.

Before the #1800 fix this was **silent**. The token-count path produced spans that started
mid-content:

```
olmo_thinker, alternating:
  trained='<|im_end|>WO<|endoftext|>'
  # ASSISTONE dropped entirely; only the fragment "WO" of ASSISTTWO trained

olmo_thinker, thinking content:
  trained='THINKONE</think>ANSWERONE<|im_end|>THINKTWO</think>ANSWERTWO<|endoftext|>'
  # the opening <think> tag masked out of the loss
```

That second case is the one worth keeping in view: a thinking model that never sees `<think>` in
its loss. It is now detected rather than trained on, but it is not *fixed*.

## Related: consecutive assistant turns

The same root cause makes consecutive assistant turns undecidable for **every** template
including `tulu` and `olmo`, because deriving the second turn's boundary requires rendering a
prefix that ends in an assistant turn — which templates special-casing the final turn render
differently from the full conversation. Rare in practice (15 rows of 939k in
`tulu-3-sft-olmo-2-mixture`) and currently dropped, but a `{% generation %}` fix would resolve
it too.

## Proposed fix

Mark assistant content in each template with `{% generation %}` blocks and read the mask from
`apply_chat_template(..., return_assistant_tokens_mask=True)`. This removes the offset
arithmetic entirely: no prefix re-rendering, no prefix-stability requirement, and content
[...]

PR #1885, opened 2026-09-06
Derive SFT label spans correctly under think chat templates

pdasigi:
## Problem

Think chat templates (`olmo_thinker`, `tulu_thinker`, the released think tokenizers) end their generation prompt with an opening `<think>` tag, so the model is started *inside* the reasoning block. SFT span derivation treated the entire `add_generation_prompt=True` render as the assistant header, which put the label boundary after a tag the turn does not necessarily reproduce. Two consequences:

- **Plain replies under a think template lost their first token(s)**, and on multi-turn conversations, which route through the token-count fallback, the verifier then rejected the whole row. Every non-reasoning reply and every multi-turn think conversation was silently dropped by `sft_tulu_filter_v1`.
- **Reasoning in a `reasoning_content` field never survived**, because the fallback verifier compared the decoded span against the raw `content` string, and a rendered turn begins with `<think>reasoning</think>` instead.

Single-turn conversations were unaffected only because they take the char-offset path, which runs no verifier. Measured on a multi-turn, tool-calling, think-formatted SFT set: 40 of 300 rows survived tokenization before this change; 297 of 300 after (the remaining three are consecutive assistant turns, see below).

## Fix

1. **Header boundary** (`_assistant_turn_boundaries`). The header is the generation-prompt render when the turn reproduces it, so a `<think>` the template forces at inference stays masked. When the turn does not reproduce its tail, the header ends where the message's own text (`reasoning_content` or `content`) first appears inside the region both renders agree on. The boundary is anchored to message text rather than snapped to a token edge, so it can never land inside a word.
2. **Verifier** (`_verify_assistant_spans_cover_content`). Spans are checked against the text the turn rendered (`through` minus header, EOS stripped) instead of raw `content`, so reasoning fields, structured tool calls and think tags are compared as the template emits them. The three checks (too narrow, starts inside header, runs past the turn) are unchanged in spirit.
3. **Tool-call arguments** (`_normalize_tool_call_arguments`). `tool_calls[].function.arguments` given as a JSON string is parsed to a mapping before rendering. Templates iterate arguments as a mapping, and datasets have to store them as strings whenever argument names or value types vary across rows, since no Arrow struct can hold that. Non-object strings pass through untouched.
4. **`DATASET_CACHE_VERSION` v7 → v8.** Labels change for every think-template dataset and the cache hash does not see code, so existing caches would otherwise stay hash-valid with stale labels.

## What stays rejected, deliberately

- Templates that ignore `add_generation_prompt` (`simple_chat`): the role marker is emitted only inside the turn, so no render can separate header from content. These are handed to the old content-anchored path and rejected exactly as before (`test_span_starting_inside_the_header_is_rejected` still passes).
- Consecutive assistant turns under templates that special-case the final turn: the prefix renders diverge inside the prompt, so the common prefix says nothing about the header. Unchanged.

## Tests

- Eight sweep cases move from "must be masked out" to "must train exactly the assistant text": `tulu_thinker`, `tulu_thinker_r1_style`, `olmo_thinker`, `olmo_thinker_remove_intermediate_thinking`, alternating shape, both EOS variants.
- New `TestThinkTemplateLabelSpans`: plain reply, inline `<think>`, `reasoning_content` field, no-think generation prompt (tag becomes trainable), partial-prefix content (`<thinking aloud>`), and single-turn/multi-turn agreement.
- New `TestToolCallArgumentNormalization`: string vs mapping arguments tokenize identically; non-object strings and flat tool calls pass through; input rows are not mutated.
- `open_instruct/test_dataset_transformation.py`: 109 passed. `test_dpo_utils.py`, `test_numpy_dataset_conversion.py`: 33 passed. `make style-check quality-check` clean.

GPU_TESTS=bypass — CPU-only tokenization change, fully covered by the unit tests above.

https://claude.ai/code/session_01F5T3P87ELvCy55JVEKxzSA

## Update: two more shapes from a second dataset

A terminal-agent SFT set (10-turn trajectories, one `bash` tool, reasoning in a `reasoning_content` field) kept only 115 of 200 sampled rows with the first version of this PR. Two independent causes, both now handled and tested:

- **Empty reasoning fields.** Arrow cannot omit a struct field, so turns without reasoning carry `reasoning_content == ""` and the template renders an empty `<think></think>` before the content. The header anchor required the body to *start* with the message's text, so these fell back and were rejected. When the turn reproduces the whole generation prompt, that prompt is the header by construction and anything after it is the turn's own output; the text anchor is now only used to place the boundary when the turn does not reproduce the prompt's tail. Templates that emit no header at all are detected directly, by the generation prompt adding nothing to the render of the preceding messages.
- **Tokens straddling the header edge.** `<think>` is not a special token in every tokenizer, so `>` and a following `<` can merge into one token across the boundary. The fallback trains that token; the verifier is told how many of its characters are header and discounts them.

After: 196 of 200 kept on that set, 297 of 300 on the first; every remaining drop is a consecutive-assistant-turn row. `test_dataset_transformation.py`: 110 passed.


abhishekraok:
See #1804

PR #1886, opened 2026-09-08
OLMoE3 latent-KDA MoE SFT on the olmo-core path, with the 65536 Dolci-Think launch script

abhishekraok:
Claude Code · bridge-cse session (Abhishek's #1853/#1859 anchor work) — agent-written

The olmo-core SFT path for the OLMoE3 latent-KDA 1.2B-active MoE, and the launch script that produced the Dolci-Think 65536 baseline (lc-think-sft-baseline (#1859), anchor-0.5ep-annealed (#1880)). Until now this lived only on spike branches; kda-moe-dpo (#1860) carries the same runtime commits as a side effect, so this is the standalone route for the SFT half.

## What it adds

- **`OLMoDDPModel` support in `olmo_core_finetune.py`.** MoE v2 models refuse FSDP2 by design; they now train through olmo-core's DDP train module, configured from the checkpoint's own `train_module` section with the SFT knobs (LR, weight decay, schedule, batch geometry, grad clipping) overridden. `--config_name` may be a path to an olmo-core `config.json`, so the architecture is read from the checkpoint rather than transcribed into a preset. `is_hf_checkpoint` now requires a top-level `model_type`, because olmo-core checkpoints also carry a `config.json`. Weight init from an olmo-core checkpoint skips the optimizer state.
- **`scripts/train/debug/oc_sft_olmoe3_kda_think.sh`** — Dolci-Think SFT on the long-context (65536-native, NoPE) KDA base. Modes for tokenize / gate / lr_probe / train / convert; `RESUME_FROM` + `EPHEMERAL_STEPS` for holmes preemption recovery; global batch fixed at 1,048,576 tokens regardless of `SEQ` and world size. The header explains every setting.
- `oc_sft_olmoe3_kda_1node.sh` (Instruct-SFT sibling on the midtrain base), `make_kda_sft_config.py` + the two generated configs, `convert_moe_checkpoint_to_hf.py`, `eval_olmoe3_kda.sh`, `probe_kda_forward.py`.
- Image: CUDA 13.3.1 + cuDNN base, transformer-engine 2.16.1 (MoE v2's permutation kernels), torchvision realigned to the torch 2.11 override, bundled cuBLAS pointed at the system copy. Only the `--cuda-version 13` image is affected.

[...]

PR #1905, opened 2026-09-20
Add template-only overrides for RL checkpoint exports

abhishekraok:
Codex · codex://threads/01a0bc67-0eac-7681-babe-a1a745942f39

The KDA SFT converter exports an instruct-dev chat template, so RL prompts do not force `<think>`. Add `--export-chat-template` to replace only the saved template and any embedded template metadata, plus a CPU-only command for existing HF exports. The override leaves tokenizer segmentation, special tokens, and weights untouched.

The new `convert_rl` launcher mode requires the saved training tokenizer directory and writes a separate `-think` export using the checked-in think-dev template pinned to `cf6d9298ed2f5d53f098f1785b58a23a24dff8ed`. Ordinary conversion defaults are unchanged. This is a temporary template choice until Kevin Farhat's Olmo 3.5 template is ready; it does not change the converter's existing tokenizer serialization behavior.

Stacked on kda-think-sft (#1886), which contains the converter and launcher. Addresses the RL export handoff in dolci-think-sft-anchor (#1895).

Validation: 15 CPU tests passed, covering template rendering, tokenizer/weight preservation, linked metadata, CLI integration, and launcher arguments. `make style quality`, script Ruff checks, shell syntax, and `git diff --check` passed using the existing local environment without dependency sync. Weight conversion was mocked in the integration tests; no compute was launched and no existing checkpoint was modified.

GPU_TESTS=bypass — template metadata and launcher changes; no model computation changes.


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 
 ### Added
+- Add an optional template-only override for MoE HF exports, a CPU-only command for existing exports, and a pinned think-dev template selected by the KDA launcher's `convert_rl` mode. Tokenizer serialization and ordinary SFT conversion defaults are unchanged.
```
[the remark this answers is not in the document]
abhishekraok: Codex · codex://threads/01a0bc67-0eac-7681-babe-a1a745942f39

Fixed in 48a6c2d79: the entry now includes https://github.com/allenai/open-instruct/pull/1905. The current Check diff CI check passes.

abhishekraok:
Codex · codex://threads/01a0bc67-0eac-7681-babe-a1a745942f39

Ran an independent Opus review at high effort on 48a6c2d79. It found no blocking bugs and raised six hardening/inherited-behavior findings. Addressed them in b22a35658:

1. Tokenizer mismatch/ignored source: snapshot the saved training tokenizer.json before conversion and require JSON equality afterward, before installing the template or emitting CONVERSION_OK. The check catches a different checkpoint-local tokenizer or Transformers backend reconstruction without silently repairing either.
2. Source-directory requirement: retained deliberately, with a concrete source documented. numpy_dataset_conversion.py saves the actual tokenizer under <numpy-cache-directory>/tokenizer/. A Hub ID is not an equivalent record of the training pipeline.
3. Inherited EXPORT_CHAT_TEMPLATE: preserve the documented explicit override for ordinary convert, but print destination, tokenizer, template, and timeout before submission; regression coverage includes an inherited override.
4. Embedded named templates: reject both dict and list forms before changing files.
5. Test import isolation: preload the torch distributed state-dict module before patching sys.modules.
6. Conversion timeout: separate CONVERT_TIMEOUT from the gate's 45m default while respecting an explicitly supplied JOB_TIMEOUT. Defaults to 2h; CPU runs may need longer.

Validation: 26 CPU tests passed; make style quality, script Ruff checks, shell syntax, and git diff --check passed. No compute launched. Conversion integration tests mock weight conversion.

The Codex bot's only finding (missing PR URL in the changelog) was already fixed in 48a6c2d79; replied to and resolved that thread.

scripts/train/debug/convert_moe_checkpoint_to_hf.py
```diff
@@ -142,6 +160,17 @@ def main() -> None:
         device=args.device,
         validation_device=args.device,
     )
+    if reference_tokenizer is not None:
+        exported_tokenizer = json.loads(
+            (pathlib.Path(args.huggingface_output_dir) / "tokenizer.json").read_text(encoding="utf-8")
+        )
+        if exported_tokenizer != reference_tokenizer:
```
farhatkevin, line 169: Could we also verify the exported tokenizer’s special tokens and IDs against the saved training tokenizer? tokenizer.json can be identical while tokenizer_config.json changes the loaded EOS token. I reproduced that case, and it would pass this check despite potentially changing RL stopping behavior.

open_instruct/export_chat_template.py
```diff
@@ -0,0 +1,79 @@
...
+
+logger = logger_utils.setup_logger(__name__)
+
+
+def read_export_chat_template(path: str | pathlib.Path | None) -> str | None:
+    """Read before converting weights so a missing or empty template fails early."""
+    if path is None:
+        return None
+    template = pathlib.Path(path).read_text(encoding="utf-8")
+    if not template.strip():
```
farhatkevin, line 25: Could we compile or render a minimal prompt with the template before reporting `CONVERSION_OK`? A nonempty but invalid template such as `{% if %} `installs successfully and fails only when `apply_chat_template` is used later.

farhatkevin (changes requested):
After these two changes looks good.

scripts/train/debug/convert_moe_checkpoint_to_hf.py
abhishekraok: Codex · codex://threads/01a0bc67-0eac-7681-babe-a1a745942f39

Fixed in 4a0fa52ad. The converter snapshots the loaded special-token role mappings and resolved IDs before conversion, then compares them with the exported tokenizer before installing the template or reporting CONVERSION_OK. This supplements the backend JSON check. Regression cases change EOS, PAD, and BOS only in tokenizer_config.json, verify tokenizer.json remains byte-identical, and require export qualification to fail.

open_instruct/export_chat_template.py
abhishekraok: Codex · codex://threads/01a0bc67-0eac-7681-babe-a1a745942f39

Fixed in 4a0fa52ad. Templates are compiled before weight conversion and again at the installation boundary, before any metadata writes. This uses the same Transformers Jinja compiler as apply_chat_template, preserving generation-block support. Regression tests reject {% if %} without changing export files or calling the converter, and verify a generation-block template still renders correctly. All 33 CPU tests and make style quality pass.

farhatkevin (approved):

